PR #1, opened 2022-02-07
Add first version of the task

IvanRamyk:


MasterpieceElbow (approved):

app/point.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, x, y):
+        self._x = x
+        self._y = y
+
+    def __eq__(self, other):
+        if not isinstance(other, Point):
+            return False
+        return self.x == other.x and self.y == other.y
+
+    def __hash__(self):
```
danylott, line 11: return self.x + self.y for debugging

PR #4, opened 2022-02-08
'Solution'

cloudprogramm:


app/main.py
```diff
@@ -1,2 +1,67 @@
...
+            index += 1
+
+        if capacity <= self._length:
+            self.resize()
+
+        self._container[index % self._current_length] =\
+            self._Node(key, value, hash_)
+        self._length += 1
+
+    def __getitem__(self, key):
```
MasterpieceElbow, line 42: It is not right, because first you need to address to index of `hash(key)`, you loop over whole container instead.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #6, opened 2022-02-08
Solution 

Abnormaltype:


liuda-chuzdiuk (approved):

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+    def __init__(self):
+        self.initial_capacity = 8
+        self.storage = [[] for _ in range(self.DEFAULT_CAPACITY)]
+        self.length = 0
+
+    def resize(self):
+        saved_storage = self.storage[:]
+        self.length = 0
+        self.initial_capacity *= self.RESIZE
```
IvanRamyk, line 14: If you change the value of the initial capacity maybe change its name to the just `capacity`?

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+        elif hashed_value == self.storage[index][2] and \
+                key == self.storage[index][0]:
+            self.storage[index][1] = value
+
+        else:
+            index = (index + 1) % self.initial_capacity
+
+            while True:
+                if not self.storage[index]:
```
IvanRamyk, line 40: Lines from 40 to 48 is almost a copy of lines from 28 to 37

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+                        key == self.storage[index][0]:
+                    self.storage[index][1] = value
+                    return
+                index = (index + 1) % self.initial_capacity
+
+    def __getitem__(self, key):
+        hashed_value = hash(key)
+        index = hashed_value % self.initial_capacity
+
+        for _ in range(self.initial_capacity):
```
IvanRamyk, line 54: This loop takes O(n) time for every search of the missing key. You should optimize it :(

IvanRamyk: I understand that you can't raise KeyError when you meet the first empty value because of deletion.
Probably, you should change deletion to fix it

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+    def __delitem__(self, key):
+        hashed_value = hash(key)
+        index = hashed_value % self.initial_capacity
+
+        for _ in self.storage[index]:
+            if hashed_value == self.storage[index][2] and \
+                    key == self.storage[index][0]:
+                return_index = self.storage[index][1]
+                self.storage[index] = []
```
IvanRamyk, line 74: Updating this element with an empty list leads to increasing time complexity in the getter.

IvanRamyk (changes requested):

app/main.py
Abnormaltype: Yes, we have O(n) only when we have many collisions

app/main.py
Abnormaltype: I've fixed __delitem__

IvanRamyk (changes requested):
```
d = Dictionary()

d[1] = True
print(d[1])
```

app/main.py
```diff
@@ -1,2 +1,113 @@
...
+                    self.storage[index][1] = value
+                return
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key):
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+
+        for _ in range(self.capacity):
+            if self.storage[index][0] == \
```
IvanRamyk, line 47:
```
d = Dictionary()
print(d[0])
```
IndexError: list index out of range instead of KeyError

IvanRamyk:
```
@pytest.mark.timeout(5)
def test_deletion():
    dictionary = Dictionary()
    for key, value in items:
        dictionary[key] = value
    for key, value in items:
    for key, value in items:
        del dictionary[key]
    print(len(dictionary))
```

IvanRamyk: The test above produces a KeyError

IvanRamyk (changes requested):

app/main.py
Abnormaltype: I've fixed

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+                    self.storage[index][1] = value
+                return
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key):
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+
+        if self.length:
+            for _ in range(self.capacity):
```
IvanRamyk, line 49: Get with line complexity :(

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+        index = hashed_value % self.capacity
+
+        while True:
+            if not self.storage[index]:
+                self.storage[index] = [key, value, hashed_value]
+                self.length += 1
+                return
+            elif hashed_value == self.storage[index][2] and \
+                    key == self.storage[index][0]:
+                if self.storage[index][2] is True:
```
IvanRamyk, line 35: This condition is always False when the condition in line 33 is True.

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+
+        if self.length:
+            while self.storage[index]:
+                if self.storage[index][0] == key \
+                        and self.storage[index][2] is True:
+                    raise KeyError
+
+                if len(self.storage[index]) != 0:
```
IvanRamyk, line 54: The same condition as in line 49

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+        try:
+            if _m is not None:
+                for k, v in _m.items():
+                    self.__setitem__(k, v)
+
+            for k, v in kwargs.items():
+                self.__setitem__(k, v)
+        except KeyError:
+            raise
+        return
```
IvanRamyk, line 126:
```suggestion

```

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+        self.storage = [[] for _ in range(self.DEFAULT_CAPACITY)]
+
+    def __contains__(self, key):
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+
+        if index > self.capacity:
+            return False
+
+        for _ in range(self.capacity):
```
IvanRamyk, line 103: 1. Isn't it a duplication of the get function?
2. One more O(n) method :)

IvanRamyk (changes requested):

app/main.py
Abnormaltype: I've fixed

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+        return self.length
+
+    def __delitem__(self, key):
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+
+        while self.storage[index]:
+            if hashed_value == self.storage[index][2] and \
+                    key == self.storage[index][0]:
+                return_index = self.storage[index][1]
```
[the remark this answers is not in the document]
Abnormaltype: I've fixed

app/main.py
Abnormaltype: I've fixed

app/main.py
Abnormaltype: I've fixed

app/main.py
Abnormaltype: I've fixed

app/main.py
```diff
@@ -1,2 +1,122 @@
...
+
+    def clear(self):
+        self.length = 0
+        self.storage = [[] for _ in range(self.DEFAULT_CAPACITY)]
+
+    def __contains__(self, key):
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+
+        if index > self.capacity:
```
IvanRamyk, line 98: After the operation in line 96 index is always less than capacity

IvanRamyk (changes requested):
Last one!

app/main.py
Abnormaltype: I've fixed

IvanRamyk (approved):
Well done!

PR #7, opened 2022-02-08
'Solution'

denyslipin:


Oleksandr563487953 (approved):

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        else:
+            index = (index + 1) % self.my_dict_length
+            while True:
+                if not self.my_dict[index]:
+                    self.my_dict[index] = [hash_, key, value]
+                    self.capacity += 1
+                    break
+                if self.my_dict[index][1] == key:
+                    self.my_dict[index][2] = value
+                    break
```
MasterpieceElbow, line 24: You did the same check (10-14 and 18-24) twice. You can write it concisely within one loop.

MasterpieceElbow (changes requested):

IvanRamyk (approved):

PR #13, opened 2022-02-09
'Solution'

liuda-chuzdiuk:


app/main.py
```diff
@@ -1,2 +1,60 @@
...
+        while self._hashtable[index] is not None:
+            current_node = self._hashtable[index]
+
+            if hash_ == current_node.hash and key == current_node.key:
+                current_node.value = value
+                return
+            index = (index + 1) % self._capacity
+
+        if (self._size + 1) > Dictionary._LOAD_FACTOR * self._capacity:
+            self._resize()
```
MasterpieceElbow, line 57: Your solution fails tests sometimes, it is bad idea to check for resize in the middle of method, because you calculate index for one capacity and then you can write it to hashtable with another capacity.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #15, opened 2022-02-09
Solution

IgorOlifirenko:


Oleksandr563487953 (approved):

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        if (self.size + 1) > self.capacity * self.LOAD_FACTOR:
+            self.resize()
+
+        hash_ = hash(key)
+        index = hash_ % len(self.elements)
+        node = [key, value, hash_]
+
+        while self.elements[index] is not None:
+            if self.elements[index][0] == key \
+                    and self.elements[index][2] == hash_:
```
MasterpieceElbow, line 31: It is bad to check in that order because, there is no case when `keys` are **equal** but `hashes` **not**

MasterpieceElbow (changes requested):

PR #18, opened 2022-02-09
Develop

Oleksandr563487953:


app/main.py
```diff
@@ -1,2 +1,111 @@
...
+    hash: int
+    key: Any
+    value: Any
+
+
 class Dictionary:
-    pass
+    table_length = 8
+    resize_coefficient = 2 / 3
+    multiplying_coefficient = 2
```
Abnormaltype, line 22: constant must be defined CAPS `multiplying_coefficient = 2 `-> `MULTIPLYING_COEFFICIENT = 2`

Abnormaltype (changes requested):

Abnormaltype (approved):

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+            current_instance = self._table[index_in_table]
+            if hash_ == current_instance.hash and key == \
+                    current_instance.key:
+                current_instance.value = value
+                return
+            index_in_table = (index_in_table + 1) % self._capacity
+
+        if (self._size + 1) > Dictionary.RESIZE_COEFFICIENT * \
+                self._capacity:
+            self._resize()
```
MasterpieceElbow, line 56: You have tests fails sometimes because you calculate index, then you may **RESIZE**, and then you store in hashtable. It is incorrect order. You may have index for OLD hashtable and try to store it in NEW hashtable.

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+            (instance.key, instance.value)
+            for instance in self._table if instance is not None
+        ]
+        return f"{message}"
+
+
+if __name__ == "__main__":
+    d = Dictionary()
+    d[25] = 100
+    d[9] = 200
```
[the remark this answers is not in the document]
Oleksandr563487953: Fixed.

app/main.py
Oleksandr563487953: That block of code was replaced, so size checking is performing before index assigned. Tests passed successfully, i hope i understood the problem properly.

app/main.py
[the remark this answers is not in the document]
IvanRamyk: Are you sure?))

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #25, opened 2022-06-02
solution

MarkoKhodan:


app/main.py
```diff
@@ -1,2 +1,61 @@
...
+                    self.container[index] = self.Node(key, value, key_hash)
+                    self.cells += 1
+                    break
+                elif self.container[index].key == key:
+                    self.container[index].value = value
+                    break
+                index = (index + 1) % self.length
+
+    def __getitem__(self, key):
+        for cell in self.container:
```
Abnormaltype, line 54: This loop takes O(n) time for every search for the missing key. It would be best if you optimized it :(

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+        index = key_hash % self.length
+
+        if self.cells == round(self.length * self.LOAD_FACTOR):
+            self.resize()
+
+        if not self.container[index]:
+            self.container[index] = self.Node(key, value, key_hash)
+            self.cells += 1
+
+        elif self.container[index].key == key:
```
IvanRamyk, line 38: Maybe compare hashes first? It will be faster

IvanRamyk (changes requested):

app/main.py
MarkoKhodan: but I can have two equal hashes, with different keys

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+
+    class Node:
+        def __init__(self, key, value, key_hash):
+            self.key = key
+            self.value = value
+            self.key_hash = key_hash
+
+    def __init__(self):
+        self.container = [None] * self.INITIAL_CAPACITY
+        self.cells = 0
```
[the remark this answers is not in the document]
IvanRamyk: cells - it's actually a number of the added elements

app/main.py
IvanRamyk: Yes, but the comparison of the hashes works much faster (imagine the comparing time for two tuples of lengths 10^6). So you can compare hashes first which are two numbers and only when they are equal compare actual keys.
If the hashes are not equal you will save a lot of time :)

IvanRamyk (changes requested):
Waiting for fixes

IvanRamyk (approved):

PR #31, opened 2022-06-08
Solution

Magus6784:


Magus6784:
Oleksii, i see you're here. Where is my comment?:D

app/main.py
```diff
@@ -1,2 +1,58 @@
...
+    def __getitem__(self, item):
+        hash_ = hash(item)
+        index_ = hash_ % self.__capacity
+
+        while self.__memory[index_] is not None:
+            if self.__memory[index_][2] == hash_ and (
+                    self.__memory[index_][0] == item
+            ):
+                return self.__memory[index_][1]
+            index_ = (index_ + 1) % self.__capacity
```
MasterpieceElbow, line 55: What if there is no such key? Make behavior similar to the original `dict`.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #33, opened 2022-06-08
Solution

DmytroSvirsa:


app/main.py
```diff
@@ -1,2 +1,39 @@
...
+
+    def __len__(self):
+        return self._flow
+
+    def _refactor(self):
+        self._capacity *= 2
+        self.old_storage = self._storage
+        self._storage = [None for _ in range(self._capacity)]
+        self._flow = 0
+        for i in self.old_storage:
```
KatiukhaO, line 37: good solution, but I think, variable name "i" is not readable in this situation. Change name "i" on more informative, and then code stand more readable

KatiukhaO (changes requested):

app/main.py
```diff
@@ -1,2 +1,39 @@
...
+        if self._flow > self._capacity * 2 / 3:
+            self._refactor()
+
+    def __getitem__(self, key_):
+        key_hash = hash(key_)
+        index = key_hash % self._capacity
+        while self._storage[index] is not None:
+            if self._storage[index][0] == key_hash:
+                return self._storage[index][1]
+            index = (index + 1) % self._capacity
```
MasterpieceElbow, line 27: What if there is no such key? Make behavior similar to the original `dict`.

app/main.py
```diff
@@ -1,2 +1,39 @@
...
+
+    def __len__(self):
+        return self._flow
+
+    def _refactor(self):
```
MasterpieceElbow, line 32: It is not a good function name:
1. Refactor is actually altering the structure to improve operations without altering the functionality.
2. Refactor has very broad meanings, rename it according to the function functionality.

app/main.py
```diff
@@ -1,2 +1,39 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self._capacity = 8
+        self._storage = [None for _ in range(self._capacity)]
+        self._flow = 0
```
MasterpieceElbow, line 5: `flow` is not obvious name here.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #36, opened 2022-06-08
Solution

darten360:


app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        self._hash_table = [None for _ in range(self._capacity)]
+
+    def __getitem__(self, key):
+        hash_ = hash(key)
+        index = hash_ % self._capacity
+        while self._hash_table[index]:
+            if self._hash_table[index][0] == hash_ and (
+                    self._hash_table[index][1] == key):
+                return self._hash_table[index][2]
+            index = (index + 1) % self._capacity
```
rotsen18, line 16: What if key isn't in your dict?

rotsen18 (changes requested):

PR #40, opened 2022-06-08
'Solution'

daria-niklaevna:


app/main.py
```diff
@@ -1,2 +1,45 @@
 class Dictionary:
-    pass
+
+    def __init__(self):
+        self.capacity = 8
+        self.size = 0
+        self.new_dict = [None for i in range(self.capacity)]
+
+    def __setitem__(self, key, value, hash_=None):
+        hash_ = hash(key)
```
rotsen18, line 9: If you every time calculate hash_ you don't need hash_ as argument in function. If you change it, you should remove from 44-45 hash_
```suggestion
    def __setitem__(self, key, value):
        hash_ = hash(key)
```

rotsen18 (approved):

app/main.py
TPolina: agree here

TPolina (approved):

PR #42, opened 2022-06-08
Solution

Hectorovich:


app/main.py
```diff
@@ -1,2 +1,107 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.ls = [None] * 8
+        self.capacity = 8
+        self.size = 0
+
+    def __setitem__(self, key, value):
+        if self.size == int(self.capacity * 2 / 3):
```
rotsen18, line 8: It would be better write 2/3 in variable

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+            raise KeyError("No value with such key in this dict.")
+
+        while self.ls[index] is not None:
+            if index == self.capacity:
+                index = 0
+            if self.ls[index][2] == key_hash and self.ls[index][0] == key:
+                return self.ls[index][1]
+            index = (index + 1) % self.capacity
+
+    def pop(self, key):
```
rotsen18, line 89: Read how pop() method work and if you want you may improve your implementation.
https://www.w3schools.com/python/ref_dictionary_pop.asp

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+
+        while self.ls[index] is not None:
+            if index == self.capacity:
+                index = 0
+            if self.ls[index][2] == key_hash and self.ls[index][0] == key:
+                self.ls[index] = None
+                self.size -= 1
+            index = (index + 1) % self.capacity
+
+    def get(self, key):
```
rotsen18, line 76: Read how get() method work and if you want you may improve your implementation.
https://www.w3schools.com/python/ref_dictionary_get.asp

rotsen18 (approved):

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+            if self.ls[index][2] == key_hash and self.ls[index][0] == key:
+                pop = self.ls[index][1]
+                self.ls[index] = None
+                self.size -= 1
+                return pop
+            index = (index + 1) % self.capacity
+
+    def update(self, dictionary):
+        for key, value in dictionary.items():
+            self.add_element(key, value)
```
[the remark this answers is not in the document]
Hectorovich: You are right. Thank you!

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+        self.size = 0
+        for cell in copied_ls:
+            if cell is not None:
+                self.add_element(cell[0], cell[1])
+
+    def __getitem__(self, key):
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+        if self.ls[index] is None:
+            raise KeyError("No value with such key in this dict.")
```
MasterpieceElbow, line 44: What if there are a few different keys with the same `key_hash % self.capacity`? `self.ls[index]` will not be `None` but the required key is missing. Rework.

app/main.py
MasterpieceElbow, line 76: `get` is not a full copy of `__getitem__`. Rework.

app/main.py
MasterpieceElbow, line 107: I agree with Taras. In this implementation `self.size` can achieve `self.capacity`. Rework.

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+
+        while self.ls[index] is not None:
+            if index == self.capacity:
+                index = 0
+            if self.ls[index][2] == key_hash and self.ls[index][0] == key:
+                pop = self.ls[index][1]
+                self.ls[index] = None
+                self.size -= 1
+                return pop
+            index = (index + 1) % self.capacity
```
MasterpieceElbow, line 103: Again, if there are a few keys with the same `key_hash % self.capacity`? `pop()` doesn't work correctly. Rework.

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+                self.ls[index] = None
+        self.size = 0
+
+    def __delitem__(self, key):
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+        if self.ls[index] is None:
+            raise KeyError("No value with such key in this dict.")
+
+        while self.ls[index] is not None:
```
MasterpieceElbow, line 68: The same problem as with `__getitem__`.

MasterpieceElbow (changes requested):

PR #46, opened 2022-06-08
All done. Finally.

Flashmobber:


NOSSUDAK (approved):

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+    def __delitem__(self, key):
+        if self[key]:
+            index = hash(key) % self.__capacity
+            self.__hash_table[index] = None
+            self.__cells -= 1
+        else:
+            raise KeyError(f"Key: {key} not found.")
+
+    def get(self, key, default=None):
+        if self[key] is not None:
```
MasterpieceElbow, line 52: Why can't `self[key]` be `None`?! You can assign a `None` value to some key. Rework.

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+            return self[key]
+        return default
+
+    def pop(self, key, default=None):
+        if self[key] is not None:
```
MasterpieceElbow, line 57: The same as with `get`.

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+            self.__delitem__(key)
+        else:
+            if default is not None:
+                return default
+            else:
+                raise KeyError(f"Key: {key} not found.")
+        return self.__getitem__(key)
```
MasterpieceElbow, line 64: You cannot return an item if you have already deleted it in the first condition.

MasterpieceElbow (changes requested):

PR #48, opened 2022-06-08
Solution

Kiawami:


app/main.py
```diff
@@ -1,2 +1,59 @@
...
+        self._elements_num = 0
+        self._hash_table = [[] for _ in range(self._size)]
+
+    def __setitem__(self, key, value):
+        hash_f = hash(key)
+        index = hash_f % self._size
+
+        if not self._hash_table[index]:
+            self._hash_table[index] = [key, hash_f, value]
+            self._elements_num += 1
```
MasterpieceElbow, line 15: You already have this condition inside the loop.

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+        elif self._hash_table[index][0] == key:
```
MasterpieceElbow, line 16: And this too.

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+
+    def __getitem__(self, key):
+        hash_f = hash(key)
+        index = hash_f % self._size
+
+        while self._hash_table[index] is not None:
+            try:
+                if self._hash_table[index][0] == key:
+                    return self._hash_table[index][2]
+                index = (index + 1) % self._size
```
MasterpieceElbow, line 54: Where should you get `IndexError` if you have (0, 1, 2) indexes inside not None `self._hash_table[index]`?

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #55, opened 2022-06-09
'Solution'

dashaprotsko:


app/main.py
```diff
@@ -1,2 +1,41 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.length = 8
+        self.size = 0
+        self.hash_table = [[] for _ in range(self.length)]
```
Korenely, line 5: use None instead of []

app/main.py
```diff
@@ -1,2 +1,41 @@
...
+        raise KeyError(f"Key: {key} not in dictionary!")
+
+    def resize(self):
+        self.length *= 2
+        hash_table_before = self.hash_table
+        self.hash_table = [[] for _ in range(self.length)]
+        self.size = 0
+        for item in hash_table_before:
+            if item:
+                self.__setitem__(item[1], item[2])
```
Korenely, line 38: use self[] instead of __setitem__

Korenely (approved):

app/main.py
```diff
@@ -1,2 +1,41 @@
...
+        self.size = 0
+        self.hash_table = [[] for _ in range(self.length)]
+
+    def __setitem__(self, key, value):
+        hash_ = hash(key)
+        index = hash_ % self.length
+        while self.hash_table[index]:
+            if self.hash_table[index][1] == key:
+                self.hash_table[index][2] = value
+                self.size -= 1
```
MasterpieceElbow, line 13: It is bad code because if you already have the key you assign it twice.

MasterpieceElbow (changes requested):

PR #58, opened 2022-06-10
Solution

Vasyl-Smutok:


IvanRamyk (changes requested):
Well done :) 
Actually, in this task, you should write your own implementation of the dictionary without using a built-in solution :)

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+        self.capacity = 8
+        self.hash_table = [None for _ in range(self.capacity)]
+        self.size = 0
+
+    def __getitem__(self, item):
+        index = hash(item) % self.capacity
+
+        for _ in range(2):
+            for i in range(self.capacity):
+                if index + i == self.capacity:
```
IvanRamyk, line 12: What about `index + i > self.capacity` case?

app/main.py
```diff
@@ -1,2 +1,66 @@
...
-    pass
+    def __init__(self):
+        self.capacity = 8
+        self.hash_table = [None for _ in range(self.capacity)]
+        self.size = 0
+
+    def __getitem__(self, item):
+        index = hash(item) % self.capacity
+
+        for _ in range(2):
```
IvanRamyk, line 10: What does this loop?

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+            for i in range(self.capacity):
+                if index + i == self.capacity:
+                    index = 0
+                if self.hash_table[index + i] is not None:
```
IvanRamyk, line 14: Could you just call break after the first None value? 
Now you search key  in the whole list which is not optimal

IvanRamyk (changes requested):

app/main.py
Vasyl-Smutok: does not work without it (

app/main.py
Vasyl-Smutok: when I do so it does not work (

IvanRamyk (approved):
Good job!

PR #60, opened 2022-06-19
Solution

MykolaIF:


app/main.py
```diff
@@ -1,2 +1,62 @@
...
+                j = self.element[i][0] % new_dict_len
+
+                while True:
+                    if new_dict[j] is None:
+                        new_dict[j] = self.element[i]
+                        break
+                    else:
+                        j = (j + 1) % new_dict_len
+
+        Dictionary.dict_len = new_dict_len
```
IvanRamyk, line 61: If you change attribute value it should be an object attribute (not class)

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+                i = (i + 1) % Dictionary.dict_len
+
+    def __getitem__(self, key):
+        hash_key = hash(key)
+        i = hash_key % Dictionary.dict_len
+
+        while True:
+            if self.element[i] is None:
+                raise KeyError
+            elif self.element[i][0] == hash_key:
```
IvanRamyk, line 32: Here you also can first check if hashes are equal

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+        while True:
+            if self.element[i][0] == hash_key:
+                return self.element[i][2]
+            else:
+                i = (i + 1) % self.dict_len
+
+    def __len__(self):
+        length = 0
+
+        for i in range(self.dict_len):
```
IvanRamyk, line 40: Could you store length in an attribute? 
Then you don't need to use a loop every time to find the length

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+            if self.element[i] is not None:
+                length += 1
+
+        return length
+
+    def resize(self):
+        new_dict_len = self.dict_len * 2
+        new_dict = [None] * new_dict_len
+
+        for i in range(self.dict_len):
```
IvanRamyk, line 50: It's better to use the __setitem__ method to add items from the old list
It would help to avoid code duplication

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+                i = (i + 1) % self.dict_len
+
+    def __getitem__(self, key):
+        hash_key = hash(key)
+        i = hash_key % self.dict_len
+
+        if self.element[i] is None:
+            raise KeyError
+
+        while True:
```
IvanRamyk, line 33: What if the key isn't found? 
It will lead to the infinite loop :) 
If you visit an empty cell you can raise a KeyError exception

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #61, opened 2022-06-20
Solution

LaskoA:
Hello, I have created class 'Dictionary' and implemented some methods - could you please check if eveything is OK so far? Hovewer I have a question regarding next task: "Also, not forgot to store (key, hash, value) as node in hash table" - should I create class 'Hashtable' for it? Or it should be done is some other way? Thanks in advance.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,26 +1,46 @@
 class Dictionary:
 
-    def __init__(self, init=None):
-        if init is not None:
-            self.__dict__.update(init)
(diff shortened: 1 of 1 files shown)


IvanRamyk (changes requested):
The task is to implement your own solution of the dictionary

LaskoA:
Finally :)

IvanRamyk (approved):
Good job!

PR #67, opened 2022-07-06
Solution

Splend1ed:


app/main.py
```diff
@@ -1,2 +1,58 @@
...
-    pass
+    def __init__(self):
+        self.length = 0
+        self.hash_table = [[] for _ in range(8)]
+        self.capacity = len(self.hash_table)
+
+    def __setitem__(self, key, value):
+        self._resize()
+        index = hash(key) % self.capacity
+        for _ in self.hash_table:
```
IvanRamyk, line 10: Why use `for` loop here
When you don't know the number of iterations `while` is a more common solution

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+        while iteration != len(self) + 1:
+            iteration += 1
+            if self.hash_table[index]:
+
+                if hash(key) == self.hash_table[index][0] and\
+                        key == self.hash_table[index][1]:
+                    self.hash_table[index] = [hash(key), key, value]
+                    break
+
+                index += 1
```
Abnormaltype, line 20: It can be much more concise. It is relevant to use such incrementation:` index = (index + 1) % self.capacity`.

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+                break
+
+    def __getitem__(self, key):
+        index = hash(key) % self.capacity
+        for _ in self.hash_table:
+            if self.hash_table[index]:
+                if self.hash_table[index][1] == key and\
+                        self.hash_table[index][0] == hash(key):
+                    return self.hash_table[index][2]
+                index += 1
```
Abnormaltype, line 36: the same here

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+                if index == len(self.hash_table):
+                    index = 0
+
+        else:
```
Abnormaltype, line 40: remove unnecessary `else`

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+            raise KeyError
+
+    def __len__(self):
+        temp_ls = []
+        for i in self.hash_table:
```
Abnormaltype, line 45: what is `i`?

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+            if i:
+                temp_ls.append(i)
+        return len(temp_ls)
```
Abnormaltype, line 48: why not `return self.length`?)

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+
+    def __setitem__(self, key, value):
+        self._resize()
+        index = hash(key) % self.capacity
+        iteration = 0
+        while iteration != len(self) + 1:
+            iteration += 1
+            if self.hash_table[index]:
+
+                if hash(key) == self.hash_table[index][0] and\
```
Abnormaltype, line 15: create variable `hash_` and store `hash(key)`

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,51 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.hash_table = [[] for _ in range(8)]
+        self.capacity = 8
```
Abnormaltype, line 4: what about `length`?

app/main.py
Abnormaltype: fix this please

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        return len(temp_ls)
+
+    def _resize(self):
+        size_to_resize = round((self.capacity / 3) * 2)
+        if len(self) >= size_to_resize:
+            temp_list = []
+
+            for item in self.hash_table:
+                if item:
+                    temp_list.append(item)
```
Abnormaltype, line 45: how to create a copy of `list`?)

Abnormaltype (changes requested):

Abnormaltype (approved):
well done!

PR #70, opened 2022-07-22
Solutin

DHushchyk:


Abnormaltype (changes requested):
Well done :)
Actually, in this task, you should write your own implementation of the dictionary without using a built-in solution :)

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+            self.threshold = int(self.capacity * (2 / 3))
+            all_items = [item for item in self.dict if item]
+            self.dict = [None for _ in range(self.capacity)]
+            for item in all_items:
+                self.key = item[0]
+                self.value = item[1]
+                self.add_to_dict()
+
+    def __hash__(self):
+        return hash((self.key, self.value))
```
Abnormaltype, line 65: you should use `hash()` function

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+
+    def __len__(self):
+        return sum(1 for item in self.dict if type(item) is list)
```
Abnormaltype, line 68: in this case, would be better to return `self.length`

app/main.py
```diff
@@ -1,2 +1,68 @@
 class Dictionary:
-    pass
+
+    def __init__(self):
+        self.key = None
+        self.value = None
```
Abnormaltype, line 5: remove this attribute

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+        self.capacity = 8
+        self.threshold = 5
```
Abnormaltype, line 7: What is it?

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+            self.dict[index] = [self.key, self.value, self.hash]
+            self.non_empty += 1
+            return
+        for index in range(len(self.dict)):
+            if type(self.dict[index]) is list:
+                if self.dict[index][0] == self.key:
+                    self.dict[index][1] = self.value
+                    return
+        if self.dict[index]:
+            for index_after in range(index + 1, self.capacity):
```
Abnormaltype, line 38: This loop takes O(n) time for every search of the missing key. You should optimize it :(

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+                if not self.dict[index_after]:
+                    index = index_after
+                    self.dict[index] = [self.key, self.value, self.hash]
+                    self.non_empty += 1
+                    return
+            for index_before in range(index):
```
Abnormaltype, line 44: the same here

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+    def __repr__(self):
+        return f"{{{self.key}: {self.value}}}"
+
+    def choose_index(self):
+        index = self.hash % self.capacity
+        if not any(self.dict):
+            self.dict[index] = [self.key, self.value, self.hash]
+            self.non_empty += 1
+            return
+        for index in range(len(self.dict)):
```
Abnormaltype, line 32: the same here

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+        self.dict = [None for _ in range(self.capacity)]
+
+    def __setitem__(self, key, value):
+        self.key = key
+        self.value = value
+        self.hash = self.__hash__()
+        self.add_to_dict()
+
+    def __getitem__(self, key):
+        for item in self.dict:
```
Abnormaltype, line 18: the same here

Abnormaltype (changes requested):

app/main.py
Abnormaltype: the same question

app/main.py
```diff
@@ -1,2 +1,43 @@
...
+            index_ = (index_ + 1) % self.capacity
+        self.dict[index_] = [self.key, self.value, hash_]
+        self.length += 1
+        if self.length == self.threshold:
+            self.resize()
+
+    def __getitem__(self, key):
+        index_ = hash(key) % self.capacity
+        while self.dict[index_]:
+            if self.dict[index_][0] == key:
```
Abnormaltype, line 28: you should check hash and key

app/main.py
```diff
@@ -1,2 +1,43 @@
...
+        self.length = 0
+        self.dict = [[] for _ in range(self.capacity)]
+
+    def __setitem__(self, key, value):
+        self.key = key
+        self.value = value
+        hash_ = hash(self.key)
+        index_ = hash_ % self.capacity
+        while self.dict[index_]:
+            if self.dict[index_][0] == key:
```
Abnormaltype, line 15: you should check key and hash

app/main.py
```diff
@@ -1,2 +1,43 @@
...
+                return self.dict[index_][1]
+            index_ = (index_ + 1) % self.capacity
+        raise KeyError
+
+    def resize(self):
+        self.capacity *= 2
+        self.threshold = int(self.capacity * (2 / 3))
+        all_items = [item for item in self.dict if item]
```
Abnormaltype, line 36: how to create a copy of the list?

Abnormaltype (changes requested):

app/main.py
DHushchyk: I need self.capacity, self.threshold, self.length values to use them. Can I create them as a class atributes?

Abnormaltype: yeap

Abnormaltype: self.threshold - what about it? Why do you use her?

Abnormaltype (changes requested):
comments

app/main.py
DHushchyk: I agree. I don`t need it. But what about self.length and self.capacity = 8? Why creating this attributes for instance is a bad idea?

Abnormaltype (changes requested):
comments

app/main.py
```diff
@@ -1,2 +1,44 @@
...
+        self.dict = [[] for _ in range(self.capacity)]
+
+    def __setitem__(self, key, value):
+        self.key = key
+        self.value = value
+        hash_ = hash(key)
+        index_ = hash_ % self.capacity
+        while self.dict[index_]:
+            if self.dict[index_][0] == key \
+                    and self.dict[index_][2] == hash(key):
```
Abnormaltype, line 15:
```suggestion
                    and self.dict[index_][2] == hash_:
```

app/main.py
```diff
@@ -1,2 +1,44 @@
...
+        self.dict[index_] = [self.key, self.value, hash_]
+        self.length += 1
+        if self.length == int(self.capacity * (2 / 3)):
+            self.resize()
+
+    def __getitem__(self, key):
+        index_ = hash(key) % self.capacity
+        while self.dict[index_]:
+            if self.dict[index_][0] == key\
+                    and self.dict[index_][2] == hash(key):
```
Abnormaltype, line 29: the same here

app/main.py
```diff
@@ -1,2 +1,44 @@
...
-    pass
+
+    def __init__(self):
+        self.capacity = 8
+        self.length = 0
+        self.dict = [[] for _ in range(self.capacity)]
+
+    def __setitem__(self, key, value):
+        self.key = key
+        self.value = value
```
Abnormaltype, line 10: it's not correct, remove this

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #72, opened 2022-07-26
Dev

DanRoman-code:


app/main.py
```diff
@@ -1,2 +1,51 @@
...
+    def __getitem__(self, item):
+        for element in self.hash_table:
+            if len(element) > 0 and element[0] == item:
+                return element[-1]
+            if element == item:
+                return item
+        raise KeyError
+
+    def __len__(self):
+        return len(self.items)
```
Abnormaltype, line 32: it's not correct, use attribute length

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+                self.hashing(hash(element[0]) % self.capacity,
+                             element[1],
+                             element[0])
+            return
+
+        hash_ = hash(key) % self.capacity
+        self.hashing(hash_, value, key)
+
+    def __getitem__(self, item):
+        for element in self.hash_table:
```
Abnormaltype, line 24: Why use for loop here
When you don't know the number of iterations while is a more common solution

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+                return element[-1]
+            if element == item:
+                return item
+        raise KeyError
+
+    def __len__(self):
+        return len(self.items)
+
+    def hashing(self, hash_, value, key_):
+        for index in range(hash_, self.capacity):
```
Abnormaltype, line 35: the same here, also you should contain the hash in the list
ex
[key, hash_, value]

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        self.items.append([key, value])
+        threshold = int((self.capacity * 2 / 3))
+        size = len(self.items)
+        if size == threshold:
+            self.capacity *= 2
+            self.hash_table = [[] for _ in range(self.capacity)]
+            for element in self.items:
+                self.hashing(hash(element[0]) % self.capacity,
+                             element[1],
+                             element[0])
```
Abnormaltype, line 17: move this logic to another method.

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        hash_ = hash(key) % self.capacity
+        self.hashing(hash_, value, key)
+
+    def __getitem__(self, item):
+        n = 0
+        while True:
+            if len(self.hash_table[n]) > 0:
+                if self.hash_table[n][0] == item:
+                    return self.hash_table[n][1]
+            n += 1
```
Abnormaltype, line 29: It can be much more concise. It is relevant to use such incrementation: `index = (index + 1) % self.capacity`.
It is possible to write this block within one while loop in 5 lines.

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+            for element in self.items:
+                self.hashing(hash(element[0]) % self.capacity,
+                             element[1],
+                             element[0])
+            return
+
+        hash_ = hash(key) % self.capacity
+        self.hashing(hash_, value, key)
+
+    def __getitem__(self, item):
```
Abnormaltype, line 23: maybe `key` instead of `item`?)

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        n = 0
+        while True:
+            if len(self.hash_table[n]) > 0:
+                if self.hash_table[n][0] == item:
```
Abnormaltype, line 27: also, you should check hash

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+                    return self.hash_table[n][1]
+            n += 1
+            if n >= len(self.hash_table):
+                raise KeyError
+
+    def __len__(self):
+        return self.items.__len__()
+
+    def hashing(self, hash_, value, key_):
+        n = hash_
```
Abnormaltype, line 37: uninformative name for a variable

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+                self.hash_table[n] = [key_, value, hash_]
+                break
+            if n == self.capacity - 1:
+                for res_index in range(self.capacity):
+                    if not self.hash_table[res_index]:
+                        self.hash_table[res_index] = [key_, value, hash_]
+                        break
+            n += 1
+            if n >= len(self.hash_table):
+                break
```
Abnormaltype, line 57: the same here

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,65 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.items = []
```
IvanRamyk, line 3: It seems that it's a redundant list. It just requires extra memory and time to add/remove elements

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        if size == threshold:
+            self.resize()
+
+        hash_ = hash(key) % self.capacity
+        self.hashing(hash_, value, key)
+
+    def resize(self):
+        self.capacity *= 2
+        self.hash_table = [[] for _ in range(self.capacity)]
+        for element in self.items:
```
IvanRamyk, line 27: here you can use an old table instead of the `self.items`

IvanRamyk (changes requested):

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #73, opened 2022-07-27
solution

tarkaiev:


app/main.py
```diff
@@ -1,2 +1,72 @@
...
+            if self.elements[input_index][0] == key \
+                    and self.elements[input_index][2] == hash_code:
+                self.elements[input_index][1] = value
+                break
+            input_index = (input_index + 1) % self.current_capacity
+
+    def check_and_resize(self):
+        if self.length == int(self.current_capacity * 2 / 3):
+            self.current_capacity *= 2
+            temp_elements = [element for element in self.elements
```
Abnormaltype, line 28: how to create a copy of the list?)

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+
+    def get(self, key, *args):
+        hash_code = hash(key)
+        output_index = hash_code % self.current_capacity
+        while self.elements[output_index] is not None:
+            if self.elements[output_index][0] == key \
+                    and self.elements[output_index][2] == hash_code:
+                return self.elements[output_index][1]
+            output_index = (output_index + 1) % self.current_capacity
+        return None if len(args) == 0 else args[0]
```
Abnormaltype, line 72: it's not correct implementation of `get()`

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+
+    def __len__(self):
+        return self.length
+
+    def clear(self):
+        self.length = 0
+        self.current_capacity = 8
+        self.elements = [None for _ in range(self.current_capacity)]
+
+    def __delitem__(self, key):
```
Abnormaltype, line 53: add this test please)
```
@pytest.mark.timeout(5)
def test_deletion():
    dictionary = Dictionary()
    for key, value in items:
        dictionary[key] = value
    for key, value in items:
    for key, value in items:
        del dictionary[key]
    print(len(dictionary))
```

Abnormaltype (changes requested):

app/main.py
tarkaiev: added

app/main.py
tarkaiev: my bad

app/main.py
tarkaiev: updated

Abnormaltype (approved):
Well done!

PR #75, opened 2022-08-02
Solution

andriy-demeshko:


app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+    def __setitem__(self, key, value):
+        if len(self) + 1 > self.capacity * 2 / 3:
+            self._resize()
+        index = hash(key) % self.capacity
+        while True:
+            if len(self.hash_table[index]) > 0:
+                if hash(key) == self.hash_table[index][0] \
+                        and key == self.hash_table[index][1]:
+                    self.hash_table[index] = [hash(key), key, value]
```
IvanRamyk, line 18: you could save the hash to the variable not to call the `hash` function twice

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+    def _resize(self):
+        previous_list = []
+        for item in self.hash_table:
+            if len(item) > 0:
+                previous_list.append(item)
+        self.capacity *= 2
+        self.hash_table = [[] for _ in range(self.capacity)]
+        for item in previous_list:
+            index = hash(item[1]) % self.capacity
```
IvanRamyk, line 46: could you use item[0] instead of calling the hash function?

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #80, opened 2022-08-12
Solution

Vlad-Peresta:


app/main.py
```diff
@@ -1,2 +1,100 @@
...
+        while self._cells[hashed_index] is not None:
+            cell_key, cell_hash, cell_value = self._cells[hashed_index]
+            if cell_key == item and cell_hash == hash(item):
+                return cell_value
+            hashed_index = (hashed_index + 1) % self._hash_table_size
+        else:
+            raise KeyError
+
+    def __len__(self):
+        return sum((cell is not None for cell in self._cells))
```
Abnormaltype, line 34: you should create attribute length and use it. Your solution it's additional time for working dictionary

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+            if cell_key == key:
+                cell[hashed_index] = cell_key, cell_hash, value
+                break
+            else:
+                hashed_index = (hashed_index + 1) % self._hash_table_size
+        cell[hashed_index] = (key, hash(key), value)
+
+    def clear(self):
+        del self._cells
+        self.__init__()
```
Abnormaltype, line 73: Incorrect implementation of this method

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+
+        while self._cells[hashed_index] is not None:
+            cell_key, cell_hash, cell_value = self._cells[hashed_index]
+            if cell_key == key:
+                self._cells[hashed_index] = None
+                break
+            else:
+                hashed_index = (hashed_index + 1) % self._hash_table_size
+        else:
+            raise KeyError
```
Abnormaltype, line 86:
```
@pytest.mark.timeout(5)
def test_deletion():
    dictionary = Dictionary()
    for key, value in items:
        dictionary[key] = value
    for key, value in items:
    for key, value in items:
        del dictionary[key]
    print(len(dictionary))
```
add this test

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+        hashed_index = self._get_hashed_index(key)
+
+        while self._cells[hashed_index] is not None:
+            cell_key, cell_hash, cell_value = self._cells[hashed_index]
+            if cell_key == key:
+                self._cells[hashed_index] = None
+                break
+            else:
+                hashed_index = (hashed_index + 1) % self._hash_table_size
+        else:
```
Abnormaltype, line 85: unnecessary else

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+            raise KeyError
+
+    def __iter__(self):
+        return iter(self.table)
+
+    def get(self, key):
+        return self.__getitem__(key)
```
Abnormaltype, line 92: incorrect implementation

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+
+    def pop(self, key):
+        value = self.__getitem__(key)
+        self.__delitem__(key)
+        return value
+
+    def update(self, key, value):
+        self.__setitem__(key, value)
```
Abnormaltype, line 100: also, incorrect implementation

Abnormaltype (changes requested):

app/main.py
Vlad-Peresta: I couldn't imagine that changing this method could make the program 5x faster

app/main.py
```diff
@@ -1,2 +1,134 @@
...
+
+    def _get_hashed_index(self, key):
+        hashed_key = hash(key)
+        index = hashed_key % self._hash_table_size
+        return index
+
+    def _is_necessary_hash_table_resizing(self):
+        threshold = int(2 / 3 * self._hash_table_size)
+        if self._length >= threshold:
+            self._hash_table_size *= 2
```
IvanRamyk, line 49: Shouldn't this action be placed in the `_perform_hash_table_resize ` method? 
I believe it's a part of that logic

IvanRamyk (approved):
Overall, good job!

PR #83, opened 2022-08-22
need help, where error?

hikehikehike:


app/main.py
```diff
@@ -1,2 +1,50 @@
...
+            prev_list = [item for item in self.hash_table if item]
+            self.hash_table = [[] for _ in range(self.capacity)]
+            for item in prev_list:
+                index = item[0] % self.capacity
+                while True:
+                    if len(self.hash_table[index]) > 0:
+                        index = (index + 1) % self.capacity
+                    else:
+                        self.hash_table[index] = item
+                        break
```
Abnormaltype, line 22: move this logic to an individual method

app/main.py
```diff
@@ -1,2 +1,50 @@
...
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.threshold = self.capacity * self.load_factor
+        self.hash_table = [[]for _ in range(self.capacity)]
+
+    def __setitem__(self, key, value):
+        if self.length > self.threshold:
+            self.capacity *= 2
+            self.threshold = self.capacity * self.load_factor
+            prev_list = [item for item in self.hash_table if item]
```
Abnormaltype, line 13: how to create a copy of the list?

app/main.py
```diff
@@ -1,2 +1,50 @@
...
+    def add_new_element(self, key, value):
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+        while True:
+            if len(self.hash_table[index]) == 0:
+                self.hash_table[index] = [hashed_value, key, value]
+                self.length += 1
+                break
+            else:
+                if key == self.hash_table[index][1]:
```
Abnormaltype, line 44: also, you should check the hash

Abnormaltype (changes requested):

app/main.py
hikehikehike: its not copy? prev_list = [item for item in self.hash_table if item]

Abnormaltype: yes, but we have a short and more best practice way to create a copy

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+            if self.hash_table[index][1] == item \
+                    and self.hash_table[index][0] == hashed_value:
+                return self.hash_table[index][2]
+            index = (index + 1) % self.capacity
+        raise KeyError
+
+    def rebuilding_hash_table(self):
+        self.capacity *= 2
+        self.threshold = self.capacity * self.load_factor
+        prev_list = [item for item in self.hash_table if item]
```
Abnormaltype, line 27: what about this?

Abnormaltype (changes requested):

app/main.py
hikehikehike: you mean i need to use deepcopy?

hikehikehike: but i need not copy, i don`t need empty cell

Abnormaltype: okay, one more comment

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.hash_table = [[] for _ in range(self.capacity)]
+        for item in prev_list:
+            index = item[0] % self.capacity
+            while True:
+                if len(self.hash_table[index]) > 0:
+                    index = (index + 1) % self.capacity
+                else:
+                    self.hash_table[index] = item
+                    break
```
Abnormaltype, line 36: use setitem here

Abnormaltype (changes requested):

app/main.py
hikehikehike: don't understand

Abnormaltype: create a loop, and use the magic method` __setitem__`

hikehikehike: for item in prev_list:
            self.__setitem__(item[1], item[2])

like this?

Abnormaltype: yes

hikehikehike: if i do like this, program don`t work

hikehikehike: in __setitem__ i call method rebuilding_hash_table

hikehikehike: Maybe you mean i need call method "add_new_element"?

Abnormaltype: image
it should work correctly if you add self.length = 0.

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+
+    def add_new_element(self, key, value):
+        hashed_value = hash(key)
+        index = hashed_value % self.capacity
+        while True:
+            if len(self.hash_table[index]) == 0:
+                self.hash_table[index] = [hashed_value, key, value]
+                self.length += 1
+                break
+            else:
```
Abnormaltype, line 41: remove unnecessary else

Abnormaltype (changes requested):
last one)

Abnormaltype (approved):
Well done!)))))))))

PR #85, opened 2022-08-24
Pythons

InnaKushnir:


app/main.py
```diff
@@ -1,2 +1,117 @@
...
+            while self.base_data[index_] is not None:
+
+                if self.base_data[index_][0] == key:
+                    self.base_data[index_] = [key, hash(key), value]
+                    self.length -= 1
+                    a = 1
+                if index_ == self.capacity - 1:
+                    index_ = -1
+
+                index_ += 1
```
Abnormaltype, line 32: you don't have to use an index. Use index = (index + 1) % self.capacity.

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+            self.base_data[index_] = [key, hash(key), value]
+
+        elif self.base_data[index_][0] == key:
+            self.base_data[index_] = [key, hash(key), value]
+            self.length -= 1
+        else:
+
+            while self.base_data[index_] is not None:
+
+                if self.base_data[index_][0] == key:
```
Abnormaltype, line 25: you should also check the hash

app/main.py
```diff
@@ -1,2 +1,117 @@
...
-    pass
+
+    def __init__(self):
+        self.capacity = 8
+        self.threshold = int(self.capacity * (2 / 3))
+        self.length = 0
+        self.base_data = [None for i in range(self.capacity)]
+
+    def __setitem__(self, key, value):
+        a = 0
```
Abnormaltype, line 10: unnecessary logic

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+
+                index_ += 1
+            if a != 1:
+                self.base_data[index_] = [key, hash(key), value]
+        self.length += 1
+
+    def __getitem__(self, key):
+        if self.length != 0:
+            index_ = hash(key) % self.capacity
+            while self.base_data[index_] is not None\
```
Abnormaltype, line 40: also, you should check the hash

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+                    and (index_ < self.capacity):
+                if self.base_data[index_][0] == key:
+                    return self.base_data[index_][2]
+
+                index_ += 1
+
+            raise KeyError(f"Not Key {key} ")
+        else:
+            raise KeyError(f"Not Key {key} ")
```
Abnormaltype, line 49: you should raise an error after while loop

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+
+        self.base_data = [
+            None for i in range(self.capacity)]
+        self.length = 0
+        for number in data_:
+            self.__setitem__(number[0], number[2])
+
+    def clear(self):
+        self.base_data = [
+            None for i in range(self.capacity)]
```
Abnormaltype, line 70: incorrect implementation

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+            raise KeyError(f"Not Key {key} ")
+
+    def __len__(self):
+        return int(self.length)
+
+    def resize(self):
+        self.capacity *= 2
+        self.threshold = int(self.capacity * (2 / 3))
+        data_ = []
+        for num in self.base_data:
```
Abnormaltype, line 58: it's a num? dict stored nums?

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+
+    def __delitem__(self, item):
+        for i in range(self.length):
+            if self.base_data[i][2] == item:
+                self.base_data[i] = None
+
+    def pop(self, key):
+        for i in range(self.length):
+            if self.base_data[i][0] == key:
+                self.base_data[i] = None
```
Abnormaltype, line 80: incorrect implementation

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+
+    def get(self, key):
+        self.__getitem__(key)
```
Abnormaltype, line 83: incorrect implementation

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+
+    def update(self, key, value):
+        self.__setitem__(key, value)
```
Abnormaltype, line 86: incorrect implementation

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+    print(d.__getitem__("b"))
+    print(d.__getitem__("1234"))
+    print(d.__getitem__("[1, 2, 3]"))
+    print(d.__len__())
+    d.__iter__()
+    d.update("1", "one")
+    d.get("1")
+    d.pop("1")
+    d.__delitem__("y")
+    d.clear()
```
Abnormaltype, line 117: remove these tests

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+            self.__setitem__(number[0], number[2])
+
+    def clear(self):
+        self.base_data = [
+            None for i in range(self.capacity)]
+
+    def __delitem__(self, item):
+        for i in range(self.length):
+            if self.base_data[i][2] == item:
+                self.base_data[i] = None
```
Abnormaltype, line 75: add these tests

```
@pytest.mark.timeout(5)
def test_deletion():
    dictionary = Dictionary()
    for key, value in items:
        dictionary[key] = value
    for key, value in items:
    for key, value in items:
        del dictionary[key]
    print(len(dictionary))
```

Abnormaltype (changes requested):

app/main.py
InnaKushnir: But the code without this logic does not work

Abnormaltype: you should create a logic without this variable

app/main.py
InnaKushnir: I don`t know where exactly to add this test, a separate file? In the condition it is also recommended to check the correctness of the code using the class, how to implement it?

Abnormaltype: add these code to the tests.py file

app/main.py
Abnormaltype: you should fix this case

app/main.py
Abnormaltype: ?

app/main.py
Abnormaltype: ?

app/main.py
Abnormaltype: ?

app/main.py
Abnormaltype: ?

app/main.py
```diff
@@ -1,2 +1,93 @@
...
+
+    def __iter__(self):
+        if self.length != 0:
+            counter = 0
+            while counter <= self.length:
+                yield self.base_data[counter]
+
+
+if __name__ == "__main__":
     pass
```
Abnormaltype, line 93: do you need this?

Abnormaltype (changes requested):

app/main.py
InnaKushnir: I need this pass

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+
+                if self.base_data[index_][0] == key and\
+                        self.base_data[index_][1] == hash(key):
+                    self.base_data[index_] = [key, hash(key), value]
+                    self.length -= 1
+                    break
+                index_ = (index_ + 1) % self.capacity
+
+            self.base_data[index_] = [key, hash(key), value]
+        self.length += 1
```
Abnormaltype, line 33: write all logic through while

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+
+    def resize(self):
+        self.capacity *= 2
+        self.threshold = int(self.capacity * (2 / 3))
+        data_ = []
+        for member in self.base_data:
+            if member is not None:
+                data_.append(member)
+        self.base_data = [
+            None for i in range(self.capacity)]
```
Abnormaltype, line 59:
```suggestion
            None for _ in range(self.capacity)]
```

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+        for member in data_:
+            self.__setitem__(member[0], member[2])
+
+    def clear(self):
+        self.init()
+
+    def __delitem__(self, item):
+        i = 0
+        while self.base_data[i] is not None:
+            if self.base_data[i][2] == item:
```
Abnormaltype, line 70: also, you should check the hash

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+
+    def pop(self, key):
+        index_ = hash(key) % self.capacity
+        while self.base_data[index_] is not None:
+            if self.base_data[index_][0] == key and\
+                    self.base_data[index_][1] == hash(key):
+                self.base_data[index_] = None
+                return
+            index_ = (index_ + 1) % self.capacity
+        return None
```
Abnormaltype, line 82: incorrect implementation, use delitem here

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+            index_ = hash(key) % self.capacity
+            while self.base_data[index_] is not None\
+                    and (index_ < self.capacity):
+                if self.base_data[index_][0] == key and \
+                        self.base_data[index_][1] == hash(key):
+                    return self.base_data[index_][2]
+
+                index_ = (index_ + 1) % self.capacity
+
+        return None
```
Abnormaltype, line 95: check the original method `get`

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+
+    @staticmethod
+    def update(**kwargs):
```
Abnormaltype, line 98: also, incorrect

Abnormaltype (changes requested):

app/main.py
InnaKushnir: I do not know how to check here hash

InnaKushnir: Hash is bound to the key

InnaKushnir: i = hash(key) % self.capacity  Check through this expression?

app/main.py
InnaKushnir: It should not be a staticmethod? Are the variables accepted by the function written incorrectly?

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+            if self.base_data[index_][0] == key and \
+                    self.base_data[index_][1] == hash(key):
+                return self.base_data[index_][2]
+
+            index_ = (index_ + 1) % self.capacity
+
+        raise KeyError(f"Not Key {key} ")
+
+    def __len__(self):
+        return int(self.length)
```
IvanRamyk, line 41:
```suggestion
        return self.length
```

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+
+        if self.length >= self.threshold:
+            self.resize()
+
+        index_ = int(hash(key) % self.capacity)
+
+        while self.base_data[index_] is not None:
+
+            if self.base_data[index_][0] == key and\
+                    self.base_data[index_][1] == hash(key):
```
IvanRamyk, line 19: Calculating a cache could take a lot of time
So it's better to calculate it once and store in the variable before the cycle

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+    def __setitem__(self, key, value):
+
+        if self.length >= self.threshold:
+            self.resize()
+
+        index_ = int(hash(key) % self.capacity)
+
+        while self.base_data[index_] is not None:
+
+            if self.base_data[index_][0] == key and\
```
IvanRamyk, line 18: In `if` you should compare hashes first because it's faster than comparing keys (e.x. they could be long tuples).

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+                break
+            index_ = (index_ + 1) % self.capacity
+
+        self.base_data[index_] = [key, hash(key), value]
+        self.length += 1
+
+    def __getitem__(self, key):
+        index_ = hash(key) % self.capacity
+        while self.base_data[index_] is not None\
+                and (index_ < self.capacity):
```
IvanRamyk, line 31:
```suggestion
                and (index_ < self.capacity):
```
index_ won't reach `self.capacity` value because you take the remainder of the division in line 36.

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #87, opened 2022-08-26
'Solution'

dmitriyKostiuk:


app/main.py
```diff
@@ -1,2 +1,81 @@
...
+            raise KeyError
+        need_item = [
+            tup for tup in self.hash_table
+            if tup is not None and tup[1] == hash_key and tup[0] == item]
+        if len(need_item) > 0:
+            return need_item[0][2]
+        raise KeyError("Invalid key")
+
+    def __len__(self):
+        return len([item for item in self.hash_table if item is not None])
```
IvanRamyk, line 62: it's bad practice to calculate this number each time using comprehension. It takes a lot of time. It's better to save length in protected varaible

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+                        self.hash_table[ind] = (key, hash_key, value)
+                        continue
+                    new_index = self.get_need_index(ind, key, hash_key)
+                    self.hash_table[new_index] = (key, hash_key, value)
+
+    def __setitem__(self, key, value):
+        hash_key = hash(key)
+        try:
+            ind = hash_key % len(self.hash_table)
+        except KeyError:
```
IvanRamyk, line 43: What could lead to the `KeyError`?

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+
+    def get_need_index(self, ind, key, hash_key):
+        if self.hash_table[ind] is None:
+            return ind
+        if self.hash_table[ind][1] == hash_key \
+                and self.hash_table[ind][0] == key:
+            return ind
+
+        # collision
+        need_item = [
```
IvanRamyk, line 13: This algorithm is too slow. Try to use open addressing

IvanRamyk (changes requested):

app/main.py
IvanRamyk: You haven't fixed this algorithm yet. It works but it works slowly

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+        if len(need_item) > 0:
+            return self.hash_table.index(need_item[0])
+
+        for new_ind in range(ind + 1, len(self.hash_table)):
+            if self.hash_table[new_ind] is None:
+                return new_ind
+        for new_ind in range(0, ind):
+            if self.hash_table[new_ind] is None:
+                return new_ind
+        raise KeyError("Logical error")
```
IvanRamyk, line 24: Also, change error message to specify what's wrong

app/main.py
IvanRamyk: You should avoid using such comprehensions.

IvanRamyk: They iterate over all items which leads to the increasing the time complexity
Ask in slack if you have questions about it.

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+    def __setitem__(self, key, value):
+        hash_key = hash(key)
+        try:
+            ind = hash_key % self.capacity
+        except TypeError:
+            return
+        new_index = self.get_need_index(ind, key, hash_key)
+        self.hash_table[new_index] = (key, hash_key, value)
+        self.__size = len([item for item in self.hash_table
+                           if item is not None])
```
DmytroSvirsa, line 47: represent it as counter, rise it by 1 when adding element and reduce by 1 when deleting

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #96, opened 2022-09-14
Solution

UaMermaid:


app/main.py
```diff
@@ -1,2 +1,83 @@
...
+        if 6 < self.length <= self.capacity * (1 / 3):
+            self.resize(0.5)
+        hash_key = hash(key)
+        index = hash(key) % self.capacity
+        while self.hash_table[index]:
+            if self.hash_table[index][1] == hash_key \
+                    and self.hash_table[index][0] == key:
+                self.hash_table[index] = []
+            index = (index + 1) % self.capacity
+        raise KeyError(key)
```
Abnormaltype, line 60: this method should return value

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+
+    def pop(self, key):
+        self.__delitem__(key)
```
Abnormaltype, line 63: incorrect implementation

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+
+    def get(self, key):
+        self.__getitem__(key)
```
Abnormaltype, line 66: incorrect implementation

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+    def update(self, other):
+        if isinstance(other, Dictionary):
+            other_table_hash = other.hash_table
+            for element in other_table_hash:
+                if len(element) != 0:
+                    self.__setitem__(element[0], element[2])
+        if isinstance(other, dict):
+            for k, v in other.items():
+                self.__setitem__(k, v)
+        raise TypeError(other)
```
Abnormaltype, line 77: incorrect implementation

Abnormaltype (changes requested):

app/main.py
UaMermaid: How to do correct?

Abnormaltype: you should return an item.

Also, add this test
```
@pytest.mark.timeout(5)
def test_deletion():
    dictionary = Dictionary()

    for key, value in items:
        dictionary[key] = value

    for key, value in items:

    for key, value in items:
        del dictionary[key]

    assert len(dictionary) == 0
```

UaMermaid: Ok, thanks

app/main.py
```diff
@@ -1,2 +1,92 @@
...
+            other_table_hash = other.hash_table
+            for element in other_table_hash:
+                if len(element) != 0:
+                    self.__setitem__(element[0], element[2])
+            return self
+        if isinstance(other, dict):
+            for k, v in other.items():
+                self.__setitem__(k, v)
+            return self
+        raise TypeError(f"{other} must be dictionary")
```
Abnormaltype, line 81: please, check the original method

app/main.py
Abnormaltype: ?

app/main.py
Abnormaltype: ?

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+
+    def get(self, key, value=None):
+        hash_key = hash(key)
+        index = hash(key) % self.capacity
+        while self.hash_table[index]:
+            if self.hash_table[index][1] == hash_key \
+                    and self.hash_table[index][0] == key:
+                return self.hash_table[index][2]
+            index = (index + 1) % self.capacity
+        return value
```
Abnormaltype, line 55: use getitem here

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+        hash_key = hash(key)
+        index = hash(key) % self.capacity
+        while self.hash_table[index]:
+            if self.hash_table[index][1] == hash_key \
+                    and self.hash_table[index][0] == key:
+                self.hash_table[index] = []
+                self.length -= 1
+                return self
+            index = (index + 1) % self.capacity
+
```
Abnormaltype, line 72: should raise KeyError if not found a key

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+                    and self.hash_table[index][0] == key:
+                value = self.hash_table[index][2]
+                self.hash_table[index] = []
+                self.length -= 1
+                return value
+            index = (index + 1) % self.capacity
+        if default is not None:
+            return default
+        else:
+            raise KeyError(f"{key} not in dictionary")
```
Abnormaltype, line 89: use delitem here

Abnormaltype (changes requested):

app/main.py
UaMermaid: Yes, but if this method has raise KeyError test is faild. I don't know why. Maybe I have some problem with loop? I use the similar loop in __getitem__

Abnormaltype: you don't need a resize here

UaMermaid: Deleted resize but it doesn't solve the problem

Abnormaltype: try to write like this:

```
    def __getitem__(self, key):
        hashed_value = hash(key)
        index = hashed_value % self.capacity

        if self.length:
            while self.storage[index]:
                if self.storage[index][0] == key \
                        and self.storage[index][2] is True:
                    raise KeyError

                if hashed_value == self.storage[index][2] and \
                        self.storage[index][0] == key:
                    return self.storage[index][1]

        raise KeyError
```

Abnormaltype: it's only an example, don't use this code, you should write it for your logic

UaMermaid: Ok, thanks

UaMermaid: And this example didn't help me

Abnormaltype: what's the error you got?

UaMermaid: KeyError

app/main.py
Abnormaltype: because you have incorrect, resize

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+                    self.hash_table[index][1] == hash_key:
+                self.hash_table[index][2] = value
+                break
+            index = (index + 1) % self.capacity
+
+    def resize(self, coefficient):
+        old_table_hash = self.hash_table
+        self.capacity = int(self.capacity * coefficient)
+        if self.capacity < 8:
+            self.capacity = 8
```
Abnormaltype, line 30: remove this

Abnormaltype (changes requested):

app/main.py
Abnormaltype: but anyway strange, push the code

app/main.py
UaMermaid: Ok

app/main.py
UaMermaid: 👌

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+                if self.hash_table[index][0] != key \
+                        and self.hash_table[index][1] is True:
+                    raise KeyError(f"{key} not in dictionary")
+                if hash_key == self.hash_table[index][1] and \
+                        self.hash_table[index][0] == key:
+                    self.hash_table[index] = []
+                    self.length -= 1
+                    return self
+                index = (index + 1) % self.capacity
+        raise KeyError(f"{key} not in dictionary")
```
Abnormaltype, line 68: I've given you an example, please, rewrite this code for your logic

Abnormaltype (changes requested):

UaMermaid:
1. I did not understand how to use the example of code, so I wrote differently.
2. I haven't ideas how to write method pop using delitem because one returns a value and the other returns nothing

IvanRamyk (approved):
Well done!

PR #97, opened 2022-09-19
Create class dict

Foggy-Cat:


app/main.py
```diff
@@ -1,2 +1,111 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.length_hash = 8
+        self.length_dict = 0
+        self.capacity = [[] for _ in range(self.length_hash)]
+
+    def __setitem__(self, key, value):
+        if self.length_dict >= self.length_hash * 2 // 3:
```
IvanRamyk, line 8: move it to the separate method

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+            temp_capacity = self.capacity.copy()
+            self.length_hash *= 2
+            self.capacity = [[] for _ in range(self.length_hash)]
+            for item in temp_capacity:
```
IvanRamyk, line 12: It's actually the logic of setitem method. So call it from the new `resize` method you will create

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+                        elif new_index == self.length_hash - 1:
+                            new_index = 0
+                        else:
+                            new_index += 1
+            del temp_capacity
+        new_index = hash(key) % self.length_hash
+        hash_index = new_index
+        while True:
+            if self.capacity[new_index]:
+                if hash_index == self.capacity[new_index][1] and\
```
IvanRamyk, line 28:
```suggestion
                if <hash value not index> == self.capacity[new_index][1] and\
```

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+                        self.capacity[new_index][0] == key:
+                    if self.capacity[new_index][2] != value:
+                        self.capacity[new_index][2] = value
+                        return
+                elif new_index == self.length_hash - 1:
+                    new_index = 0
+                else:
+                    new_index += 1
+            else:
+                self.capacity[new_index] = [key, hash(key), value]
```
IvanRamyk, line 38: Do not calculate hash multiple times in one method

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+                self.length_dict += 1
+                return
+
+    def __getitem__(self, key):
+        hash_ = hash(key)
+        new_index = hash_ % self.length_hash
+        count = 0
+        while True:
+            if count == self.length_hash:
```
IvanRamyk, line 47: You should not visit all cells but stop when you face the first empty one

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+    def __len__(self):
+        return self.length_dict
+
+    def __delitem__(self, key):
+        hash_index = hash(key) % len(self.capacity)
+        count = 0
+        while True:
+            if count == len(self.capacity):
+                raise KeyError(f"{key} is not founded")
+            if self.capacity[hash_index][0] == key:
```
IvanRamyk, line 68: Check hash first

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+                self.capacity[hash_index] = []
```
IvanRamyk, line 69: If you just assign an empty cell then open addressing won't work

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+                        self.capacity[new_index][2] = value
+                        return
+                elif new_index == self.length_hash - 1:
+                    new_index = 0
+                else:
+                    new_index += 1
+            else:
+                self.capacity[new_index] = [key, hash_, value]
+                self.length_dict += 1
+                return
```
IvanRamyk, line 27: redundant return

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+                raise KeyError(f"{key} is not founded")
+            if self.capacity[new_index] and\
+                    self.capacity[new_index][0] == key and\
+                    self.capacity[new_index][1] == hash_:
+                return self.capacity[new_index][2]
+            elif new_index == self.length_hash - 1:
+                new_index = 0
+            else:
+                new_index += 1
+            count += 1
```
IvanRamyk, line 44: Why do you need this variable?

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        hash_ = hash(key)
+        hash_index = hash_ % len(self.capacity)
+        while True:
+            if not self.capacity[hash_index]:
+                raise KeyError(f"{key} is not founded")
+            if self.capacity[hash_index] and \
+                    self.capacity[hash_index][1] == hash_ and\
+                    self.capacity[hash_index][0] == key:
+                pop_item = self.capacity[hash_index]
+                self.capacity[hash_index].clear()
```
IvanRamyk, line 59: You cannot just clear the corresponding cell

IvanRamyk: image

IvanRamyk: Comment where I describe this issue

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        return f"{{{', '.join(new_string)}}}"
+
+    def __resize_hash(self):
+        temp_capacity = self.capacity.copy()
+        self.length_hash *= 2
+        self.capacity = [[] for _ in range(self.length_hash)]
+        for item in temp_capacity:
+            if item:
+                new_index = item[1] % self.length_hash
+                while True:
```
IvanRamyk, line 82: Use `__setitem__` here, DRY

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #98, opened 2022-09-20
Solution

Paul-Starodub:


app/main.py
```diff
@@ -1,2 +1,65 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.capacity = 8
+        self.length = 0
+        self.load = int(self.capacity * 2 / 3)
+        self.tab = [[] for _ in range(self.capacity)]
```
IvanRamyk, line 6:
```suggestion
        self.table = [[] for _ in range(self.capacity)]
```
Use more descriptive names

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+
+    def __setitem__(self, key, value):
+        # preparing index of tab
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        while True:
+            # if new key length increases
+            if not len(self.tab[index]):
```
IvanRamyk, line 14:
```suggestion
            if self.tab[index] == []:
```
It's better to check explicitly

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+                self.tab[index] = [key, value, hashed_key]
+                self.length += 1
+                break
+                # if key in dictionary
+            if self.tab[index][0] == key and self.tab[index][2] == hashed_key:
```
IvanRamyk, line 19: check hash value first

IvanRamyk (changes requested):

app/main.py
Paul-Starodub: it was my first variant)

Paul-Starodub: pycharm wants "is"

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+                break
+            index = (index + 1) % self.capacity
+        if self.length == self.load:
+            self.resize()
+
+    def resize(self):
+        self.capacity *= 2
+        self.length = 0
+        self.load = int(self.capacity * 2 / 3)
+        prep_table = self.table
```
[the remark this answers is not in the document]
IvanRamyk: change the name to make clear what you mean

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+                self.__setitem__(element[0], element[1])
+
+    def __len__(self):
+        return self.length
+
+    def __getitem__(self, key):
+        index = hash(key) % self.capacity
+        element = self.table[index]
+        while element:
+            if element[2] == hash(key) and element[0] == key:
```
IvanRamyk, line 41: do not calculate hash multiple times inside cycle
store this value at the beginning

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+                return element[1]
+            index = (index + 1) % self.capacity
+            element = self.table[index]
+        raise KeyError()
```
IvanRamyk, line 45: add an error message

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+        self.length = 0
+        self.load = int(self.capacity * 2 / 3)
+        self.table = [[] for _ in range(self.capacity)]
+
+    def __delitem__(self, key):
+        index = hash(key) % self.capacity
+        element = self.table[index]
+        while element:
+            if element[2] == hash(key) and element[0] == key:
+                self.table[index] = []
```
IvanRamyk, line 58: you cannot just clear the cell
the detailed information is in the following screenshot:
image

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+                self.length -= 1
+                break
+            index = (index + 1) % self.capacity
+            element = self.table[index]
+        raise KeyError
```
IvanRamyk, line 63: also, add a message, please

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+
+    def get(self, key, value=None):
+        list_actual_keys = (_[0] for _ in self.table if _ != [])
```
IvanRamyk, line 66: This comprehension requires O(size) operations to get through all elements 
Use getitem and `try`-`except` block only. It takes O(1) operations

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+        if key not in list_actual_keys:
+            return value
```
IvanRamyk, line 68: You should raise an error if key doesn't present and user didn't pass the default value

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+        return self.__getitem__(key)
+
+    def pop(self, __key, default=None):
+        list_actual_keys = (_[0] for _ in self.table if _ != [])
```
IvanRamyk, line 72: O(n) complexity

IvanRamyk: Use the same approach described for `get` method

IvanRamyk (changes requested):

app/main.py
Paul-Starodub: This method returns the value for the given key. If the key is not available, the default value None is returned. Not?

IvanRamyk (changes requested):
What about deletion?
Other fixes are OK

IvanRamyk (approved):
Good job!

PR #101, opened 2022-09-23
Solution

KatyaVasylieva:


app/main.py
```diff
@@ -1,2 +1,99 @@
...
+
+        # rewrite the value of existing key:
+        for ind, element in enumerate(bucket):
+            if element[0] == key:
+                bucket[ind] = (key, hashed_value, value)
+                break
+
+        # create new key-value pair and check for the need to resize
+        if (key, hashed_value, value) not in bucket:
+            fullness = (sum(1 for el in self.buckets if el) + 1) / self.size
```
DmytroSvirsa, line 26: not the best way to calculate, you make this long math, try to find a faster way

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+    def __getitem__(self, key):
+        hashed_value = hash(key)
+        bucket = self.buckets[hashed_value % self.size]
+        for current_key, hashed_value, value in bucket:
+            if current_key == key:
+                return value
+        raise KeyError
+
+    def __len__(self):
+        return len(sum(self.buckets, []))
```
DmytroSvirsa, line 47: the same here

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+
+    def __setitem__(self, key, value):
+        hashed_value = hash(key)
+        bucket = self.buckets[hashed_value % self.size]
+
+        # rewrite the value of existing key:
+        for ind, element in enumerate(bucket):
+            if element[0] == key:
+                bucket[ind] = (key, hashed_value, value)
+                break
```
DmytroSvirsa, line 22: You did not handle "collision" - when elements have the same `hashed_value % self.size` but the keys are different

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #104, opened 2022-09-27
Custom dict + few additional methods

wQuelS:


AriohBelskij (approved):
good!

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+
+    def create_hash_table(self):
+        return [[] for _ in range(self.capacity)]
+
+    def clear(self):
+        self.hash_table = self.create_hash_table()
+        return self
+
+    def __delitem__(self, key):
+        for ind, val in enumerate(self.hash_table):
```
Nattalli, line 60: I just wanted to praise you, but here you used abbreviations for names
Abbreviations are appropriate for long names, in this case, you shortened by 2-3 letters, so is it significant? It is better to complete the word completely, it will be more readable

Nattalli (changes requested):
left comments with advice, and in general, very good work. Well done!

Nattalli (approved):
Excellent! 🌞

PR #105, opened 2022-09-27
Develop

AriohBelskij:


wQuelS (approved):
Looks great. But i don't know about deepcopy. Copy() would easely handle your problem, but anyway, let's see what metors say.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
-    pass
+    def __init__(self):
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
+        self.table = [[] for _ in range(self.capacity)]
+        self.length = 0
+
+    def __getitem__(self, key):
+        hash_key = hash(key)
+        get_ind = hash_key % self.capacity
```
Nattalli, line 13: `get_ind` is something like `get_index`? I think that if you add 2 letters, it will look more readable
because if you shortened like that, then 2 letters is a minor reduction
WDYT?

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+            except IndexError:
+                raise KeyError
+            get_ind = (get_ind + 1) % self.capacity
+
+    def __setitem__(self, key, value):
+        if self.length == self.threshold:
+            self.upgrade_table()
+
+        hash_key = hash(key)
+        ind = hash_key % self.capacity
```
Nattalli, line 32: is it `index`?

Nattalli (changes requested):
left some comments, correct at your discretion. Nice work, I love it!

app/main.py
AriohBelskij: i fix it, sorry

Nattalli (approved):
Amazing 😻

PR #106, opened 2022-09-27
Solution

IvanStored:


wQuelS (changes requested):
check your get method.
what if i am trying to get() unexisting key with second parameter? 
For example -> your_dict("not_a_key", "?") -> here, method should return "?" if "not_a_key" does not exist

app/main.py
```diff
@@ -49,5 +76,8 @@ def clear(self):
         self.hash_table = [None for _ in range(self.capacity)]
         return self
 
-    def get(self, key):
-        return self.__getitem__(key)
+    def get(self, key, value="None"):
```
wQuelS, line 79: None should not be a string, i guess:)

wQuelS (changes requested):

app/main.py
IvanStored: yes, you are right :)))

wQuelS (approved):
Looks nice

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        if self.size == self.threshold:
+            self.resize()
+        hash_ = hash(key)
+        index = hash_ % self.capacity
+        while True:
+            if not self.hash_table[index]:
+                self.hash_table[index] = [key, hash_, value]
+                self.size += 1
+                break
+            if self.hash_table[index][0] == key \
```
MasterpieceElbow, line 19: Avoid using `\`. Use `( )` for the whole expression. Remake it everywhere.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #109, opened 2022-09-29
Solution

sbublinskyi:


mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+        return self.hash_table[position][2]
+
+    def __len__(self):
+        return self._size
+
+    def clear(self):
+        self.hash_table.clear()
+
+    def get(self, key):
+        return self[key]
```
mltdll, line 62: Dict get() method returns None if key is not found. You should implement this behavior here.

mltdll (changes requested):
Sorry, I forgot to mention something you should fix.

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+
+    def __delitem__(self, key) -> None:
+        position = self.__find_key_in_table_if_exists(key)
+        del self.hash_table[position]
+        self.__refill_table()
+
+    def clear(self) -> None:
+        self.hash_table.clear()
+
+    def get(self, key):
```
mltdll, line 68: Also, since you started to add type annotations: all `keys` may me annotated as Hashable and all values may be `Any`. Those types can be imported from `typing` module.

mltdll (changes requested):

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        position = self.__find_key_in_table_if_exists(key)
+        del self.hash_table[position]
+        self.__refill_table()
+
+    def clear(self) -> None:
+        self.hash_table.clear()
+
+    def get(self, key: Hashable) -> Any:
```
MasterpieceElbow, line 71: `get` also can receive the default value that is returned when a key isn't found.

MasterpieceElbow (changes requested):

TPolina (approved):
Well done!

PR #110, opened 2022-09-29
Solution

pazyn404:


app/main.py
```diff
@@ -1,2 +1,98 @@
...
+        for pair in iter(self):
+            ind = hash(pair) % len(new_hash_table)
+            new_hash_table[ind].append(pair)
+        self._hash_table = new_hash_table
+
+    def clear(self) -> None:
+        for i in range(self.capacity):
+            self._hash_table[i] = []
+
+    def pop(self, key: Hashable) -> None:
```
mltdll, line 51: Dict's `pop()` methods returns deleted value. You should implement this behavior here.

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+            self._resize()
+
+    def _resize(self) -> None:
+        new_hash_table = [[] for _ in range(self.capacity * 2)]
+        for pair in iter(self):
+            ind = hash(pair) % len(new_hash_table)
+            new_hash_table[ind].append(pair)
+        self._hash_table = new_hash_table
+
+    def clear(self) -> None:
```
mltdll, line 47: Dict's `clear()` method also changes its capacity back to 8. Can't you just `[[]] * self._INITIAL_CAPACITY`?

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+
+
+# I implement dictionary via chaining
 class Dictionary:
-    pass
+    _INITIAL_CAPACITY = 8
+    _LOAD_FACTOR = 2 / 3
+
+    def __init__(self) -> None:
+        self._hash_table = [[] for _ in range(Dictionary._INITIAL_CAPACITY)]
```
mltdll, line 32: Also, I think that here `[[]] * self._INITIAL_CAPACITY` would be a better way to do it. Ignore this comment if you disagree.

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+
+    def _ensure_capacity(self) -> None:
+        self._size += 1
+        if self._size > int(self.capacity * Dictionary._LOAD_FACTOR):
+            self._resize()
+
+    def _resize(self) -> None:
+        new_hash_table = [[] for _ in range(self.capacity * 2)]
+        for pair in iter(self):
+            ind = hash(pair) % len(new_hash_table)
```
mltdll, line 43: I haven't ever seen `ind` as abbreviation for `index` in Python code. I'd suggest to use the whole word if you aren't completely sure it's necessary. It only saves you 2 out of 5 characters anyway.

mltdll (changes requested):

app/main.py
pazyn404: Thanks for comment, but it doesn't work in such way because all list will have the same reference.

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+        return res
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def __get__(self, key: Hashable) -> Any:
+        ind = hash(key) % self.capacity
```
mltdll, line 63: There are also `ind`s used in oher functions: here...

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+        for i, pair in enumerate(self._hash_table[ind]):
+            if pair == key:
+                return pair.value
+        raise KeyError(f"Key {key} does not exist")
+
+    def __set__(self, key: Hashable, value: Any) -> None:
+        ind = hash(key) % self.capacity
```
mltdll, line 70: Here...

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+        self._ensure_capacity()
+
+    def __getitem__(self, item: Hashable) -> Any:
+        return self.__get__(item)
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        self.__set__(key, value)
+
+    def __delitem__(self, key: Hashable) -> None:
+        ind = hash(key) % self.capacity
```
mltdll, line 85: ... And here. Sorry for not mentioning it right away.

mltdll (approved):

app/main.py
mltdll: Whoops. I forgot about that.

Nattalli (changes requested):
The comments about `ind` variables is good. Please, fix this moment

Y-Havryliv (approved):
Well done!

PR #111, opened 2022-09-29
Solution

StasOkhrym:


app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        return [[] for _ in range(self.capacity)]
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.length = 0
+        self.capacity *= 2
+        self.threshold = int(self.capacity * self.LOAD_FACTOR)
+        self.hash_table = self.make_hash_table()
+        for element in old_hash_table:
+            if len(element) != 0:
```
mltdll, line 23: For lists, `len(list) != 0` is equivalent to `not list`. I'd suggest to change this check to the latter.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+                   and self.hash_table[hash_index][1] == hash_key:
+                    return self.hash_table[hash_index][2]
+            except IndexError:
+                raise KeyError
+            hash_index = (hash_index + 1) % self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.length == self.threshold:
+            self.resize()
+        hash_key = (hash(key))
```
mltdll, line 41: Since English is an analytic language rather than synthetic, order of words matters. Therefore, i would suggest to rename this variable to key_hash, so that it would look like "hash that belongs to key" rather than "key that belongs to hash".

mltdll (changes requested):

mltdll (approved):

Nattalli (approved):
Amazing 😻

PR #112, opened 2022-09-29
Solution

VladyslavKlevanskyi:


app/main.py
```diff
@@ -1,2 +1,91 @@
...
+        index = self._hashed_value_calculation(input_key)
+        while self.buckets[index] is not None:
+            key, value = self.buckets[index]
+            if key == input_key:
+                return value
+            index = (index + 1) % len(self.buckets)
+        raise KeyError
+
+    def __len__(self):
+        return len(self.buckets) - self.buckets.count(None)
```
DmytroSvirsa, line 37: This method spends a lot of time on calculations every time, find the faster way

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+
+    def clear(self):
+        self.__init__()
```
DmytroSvirsa, line 40: The clear method must clear the dictionary "buckets", but do not resize it to default

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+        while self.buckets[index] is not None:
+            key, value = self.buckets[index]
+            if key == input_key:
+                for_returning = self.buckets[index][1]
+                self.buckets[index] = None
+                return for_returning
+            index = (index + 1) % len(self.buckets)
+        if args:
+            return args
+        else:
```
DmytroSvirsa, line 72: unnecessary return

DmytroSvirsa (changes requested):

app/main.py
VladyslavKlevanskyi: Why? I have to return the argument, if it was entered, in the event that the key is not found.

app/main.py
DmytroSvirsa: no You must clear it to state self.buckets = [None] * len of dict before clear

app/main.py
DmytroSvirsa: You can avoid else when have return statement.
Good example:

```
def is_adult(age: int) -> str:
    if age >= 18:
        return "adult"
    return "not an adult"
```

Bad example:

```
def is_adult(age: int) -> str:
    if age >= 18:
        return "adult"
    else:
        return "not an adult"
```

DmytroSvirsa (changes requested):
read coments

app/main.py
VladyslavKlevanskyi: ok, but it means unnecessary `else` not `return` )

app/main.py
```diff
@@ -1,2 +1,101 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.length = 0
+        self.capacity = 8
+        self.buckets = [None] * self.capacity
+
+    def _hashed_value_calculation(self, key):
```
Abnormaltype, line 7: add type annotation

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        hashed_value = hash(key)
+        return hashed_value % len(self.buckets)
+
+    def _resize(self):
```
Abnormaltype, line 11: add type annotation

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        self.length = 0
+        old_buckets = self.buckets.copy()
+        self.buckets = [None] * self.capacity * 2
+        self.capacity *= 2
+        for i in old_buckets:
```
Abnormaltype, line 16: i - it's a bad name for a variable

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+            if i is not None:
+                self.__setitem__(i[0], i[1])
+
+    def __setitem__(self, key, value):
+        index = self._hashed_value_calculation(key)
+        while self.buckets[index] is not None:
+            if key == self.buckets[index][0]:
```
Abnormaltype, line 23: also, you should the hash

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def __getitem__(self, input_key):
+        index = self._hashed_value_calculation(input_key)
+        while self.buckets[index] is not None:
+            key, value = self.buckets[index]
+            if key == input_key:
+                return value
+            index = (index + 1) % self.capacity
+        raise KeyError
+
```
Abnormaltype, line 41: also, you should check the hash

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+    def __delitem__(self, input_key):
+        index = self._hashed_value_calculation(input_key)
+        while self.buckets[index] is not None:
+            key, value = self.buckets[index]
+            if key == input_key:
+                self.buckets[index] = None
+                self.length -= 1
+                return None
+            index = (index + 1) % self.capacity
+        raise KeyError
```
Abnormaltype, line 58: also, yous should check the hash

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def get(self, input_key, args=None):
+        index = self._hashed_value_calculation(input_key)
+        while self.buckets[index] is not None:
+            key, value = self.buckets[index]
+            if key == input_key:
+                return value
+            index = (index + 1) % self.capacity
+        return args
```
Abnormaltype, line 67: use getitem here

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+            key, value = self.buckets[index]
+            if key == input_key:
+                for_returning = self.buckets[index][1]
+                self.buckets[index] = None
+                self.length -= 1
+                return for_returning
+            index = (index + 1) % self.capacity
+        if args:
+            return args
+        raise KeyError
```
Abnormaltype, line 81: use delitem here

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def update(self, inp):
+        for key in inp:
+            self.__setitem__(key, inp.__getitem__(key))
```
Abnormaltype, line 85: incorrect implementation

Abnormaltype (changes requested):
add type annotation

app/main.py
VladyslavKlevanskyi: Why? I tested it with the dictionary {"7745": "1600", "color": "White", "12": "34"}, it works. Or should I test this method with a dictionary created from my class?

Abnormaltype: `    def update(self, _m=None, **kwargs):` - compare with original arguments

app/main.py
Abnormaltype: image
different hash, but index the same

You should firstly check the hash

Abnormaltype (changes requested):

app/main.py
VladyslavKlevanskyi: I still don't understand what you mean by checking the hash?
I am calculating the hash and converting it to an index for the table in the `_hashed_value_calculation function`.
The hash itself is not stored anywhere in my dictionary.

app/main.py
VladyslavKlevanskyi: I'm confused. Can you give a little more detail please?

app/main.py
Abnormaltype: delitem -- should return a value)

app/main.py
Abnormaltype: and&) you should store the hash in your dictionary)

app/main.py
Abnormaltype: provide value by default, and add **kwargs

VladyslavKlevanskyi: Sorry, can you explain in more detail? My method works with both built-in dictionaries and dictionaries created in my class. I checked.
While I do not understand your hints.

VladyslavKlevanskyi: Thank you! 
I implemented the function this way, but I don't understand what should go into kwargs. That is, in what cases should it be used?
изображение

Abnormaltype: what can store in **kwargs?

VladyslavKlevanskyi: key-value pairs

Abnormaltype: create a loop, and use `setitem` for it)

Abnormaltype: update(inp={1: 2}, key1=value1, key2=value2)

VladyslavKlevanskyi:
I don't know why, but `point.py` did't pass check. It is strange, I even didn't commit this file, and I didn't change it.

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+                return args
+            raise KeyError
+
+    def update(self, inp=None, **kwargs):
+        if inp:
+            for key in inp:
+                self.__setitem__(key, inp.__getitem__(key))
+        if kwargs:
+            for k in kwargs:
+                self[k] = kwargs[k]
```
Abnormaltype, line 84:
```suggestion
            for key, value in kwargs.items():
                self.__setitem__(key, value)
```

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        return self
+
+    def __next__(self) -> any:
+        if self.current_element >= self.new_buckets_length:
+            raise StopIteration
+        result = self.new_buckets[self.current_element]
+        self.current_element += 1
+        print(result)
+        return result[0]
+
```
Abnormaltype, line 106: remove unnecessary prints)

Abnormaltype (changes requested):
the last one)

Abnormaltype (approved):
Well done!

PR #114, opened 2022-10-04
Solution

AVyha:


app/main.py
```diff
@@ -1,2 +1,55 @@
 class Dictionary:
-    pass
+    def __init__(self):
+        self.capacity = 8
+        self.load_factor = 0.66
```
mltdll, line 4: 0.67 is a better approximation than 0.66. :)
You may also use `2 / 3`, I think it's even better.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+                return
+
+    def __getitem__(self, key: (int, float, str, tuple, bool)) -> None:
+        for cell in self.list:
+            try:
+                if key == cell[0]:
+                    return cell[1]
+            except IndexError:
+                pass
+        raise KeyError
```
mltdll, line 41: It's not a very good practice to raise errors without messages. Python's dict uses the key it tried to find as message for `KeyError`. I think we should do the same.

mltdll (changes requested):

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+                self.list[key_hash] = [key, value, hash(key)]
+                return
+
+    def __getitem__(self, key: (int, float, str, tuple, bool)) -> None:
+        for cell in self.list:
+            try:
+                if key == cell[0]:
+                    return cell[1]
+            except IndexError:
+                pass
```
Nattalli, line 40: does `try except` block needed here?

Nattalli (changes requested):

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+            if cell:
+                count += 1
+        return count
+
+    def increase_capacity(self) -> None:
+        old_list = [char for char in self.list if char]
+        self.capacity *= 2
+        self.list = [[] for _ in range(self.capacity)]
+        for char in old_list:
+            self[char[0]] = char[1]
```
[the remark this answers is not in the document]
AVyha: in this function i use __setitem__ for `self[char[0]] = char[1]`, i dont understand what i must rewrite

app/main.py
AVyha: in this cycle i trying to get key from cell, but many cells are empty, without `try: except:` this code shouldn't work, i think try dont interfere with the method.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+            self.list[key_hash] = [key, value, hash(key)]
+            return
+
+        for i in range(len(self.list)):  # check if we have key in self.list
+            try:
+                if self.list[i][0] == key:
+                    self.list[i][1] = value
+                    return
+            except IndexError:
+                pass
```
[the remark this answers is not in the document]
AVyha: im try to find key if that key exist in my custom dictionary, when i dont find a key in next loop i find next empty cell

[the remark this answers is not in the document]
Nattalli: okay, now I understand
Sorry, I don't know how I wrote my previous comment 🥲

app/main.py
Nattalli: if they are empty - will not work something like `if cell[0]`, so that means that it is not empty

app/main.py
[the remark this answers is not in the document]
Nattalli: okay, then let it be in this way

Nattalli (approved):

PR #115, opened 2022-10-05
Solution

AndriyKy:


Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,48 @@
...
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.fullness = 0
+        self.buckets = [None] * self.capacity
+
+    def __resizer(self) -> None:
+        self.capacity *= 2
+        self.fullness, old_buckets = 0, self.buckets
```
IvanRamyk, line 13: It's better to define a single variable in one line

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.fullness = 0
+        self.buckets = [None] * self.capacity
+
+    def __resizer(self) -> None:
```
IvanRamyk, line 11:
```suggestion
    def __resize(self) -> None:
```
Name functions using verbs (classes using nouns)

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        while True:
+            if self.buckets[index] is None:
+                self.buckets[index] = [key, value, hashed_key]
+                self.fullness += 1
+                break
+            if self.buckets[index][0] == key and \
+                    self.buckets[index][2] == hashed_key:
```
IvanRamyk, line 32: It's better to compare hashes first because comparing integers is faster then e.x. comparing huge tuples

IvanRamyk (approved):
Well done!

PR #118, opened 2022-10-06
Solution

luxuriant777:


app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        for element in old_hash_table:
+            if element:
+                self.__setitem__(element[0], element[2])
+
+    def __getitem__(self, key: Hashable) -> Any:
+        key_hash = hash(key)
+        hash_index = key_hash % self.capacity
+        while True:
+            try:
+                if self.hash_table[hash_index][0] == key\
```
MasterpieceElbow, line 31: Avoid using `try` if it is not necessary

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        if self.length == self.threshold:
+            self.resize()
+        key_hash = (hash(key))
+        hash_index = key_hash % self.capacity
+        while True:
+            if not self.hash_table[hash_index]:
+                self.hash_table[hash_index] = [key, key_hash, value]
+                self.length += 1
+                break
+            if self.hash_table[hash_index][0] == key and\
```
MasterpieceElbow, line 48: Avoid using `\`, use parentheses.

MasterpieceElbow (changes requested):

app/main.py
luxuriant777: changed

app/main.py
luxuriant777: deleted

TPolina (approved):
Well done!

PR #120, opened 2022-10-07
Solution)

Gr1zzZzzly:


app/main.py
```diff
@@ -1,2 +1,41 @@
...
+    def resize(self) -> None:
+        temp_ = [el for el in self.dict_ if el is not None]
+        if len(temp_) > (self.size * (2 / 3)):
+            self.size *= 2
+            self.dict_ = [None for _ in range(self.size)]
+            for el in temp_:
+                self.__setitem__(el[0], el[2])
+
+    def __len__(self) -> int:
+        return len([el for el in self.dict_ if el is not None])
```
mltdll, line 18: I think it would worth it to create an attribute to store current length so that the dict doesn't have to compute this every time you try to get its length. Especially since that happens every time you add an element...
Oh, wait. You compute the same thing in the `resize()` method even though you have defined the `__len__()`. Well, you shouldn't.

app/main.py
```diff
@@ -1,2 +1,41 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size = 8
+        self.dict_ = [None for i in range(self.size)]
+
+    def resize(self) -> None:
+        temp_ = [el for el in self.dict_ if el is not None]
+        if len(temp_) > (self.size * (2 / 3)):
```
mltdll, line 11: Move this check up and use `len(self)` in it. Also, I would argue that it would be even better to move this check into `__setitem__()` and only call `resize()` when you need, but that's just my preference, I guess.

app/main.py
```diff
@@ -1,2 +1,41 @@
...
+                return
+                self.resize()
+            index_ = (index_ + 1) % self.size
+        self.dict_[index_] = [key, hash_, value]
+        self.resize()
+
+    def __getitem__(self, key: Hashable) -> None:
+        index_ = hash(key) % self.size
+        counter = 0
+        while counter <= (self.size + 1):
```
mltdll, line 35: I'm not sure why you wouldn't use a for loop here. If there's a reason, please do tell.

mltdll (changes requested):

app/main.py
Gr1zzZzzly: thx for all tips, it's beneficial to me))

app/main.py
```diff
@@ -1,21 +1,25 @@
...
         self.size = 8
+        self.lenght = 0
         self.dict_ = [None for i in range(self.size)]
 
     def resize(self) -> None:
-        temp_ = [el for el in self.dict_ if el is not None]
-        if len(temp_) > (self.size * (2 / 3)):
+        if len(self) > (self.size * (2 / 3)):
+            self.lenght = 0
+            temp_ = deepcopy(self.dict_)
```
mltdll, line 14: Don't use deepcopy here. You probably don't want the objects in your dict to suddenly get replaced with copies when it gets resized.

app/main.py
```diff
@@ -1,21 +1,25 @@
+from copy import deepcopy
 from typing import Any, Hashable
 
 
 class Dictionary:
     def __init__(self) -> None:
         self.size = 8
+        self.lenght = 0
```
mltdll, line 8: You mean length? :)

mltdll (changes requested):

app/main.py
mltdll: You're welcome. :)
There are a few more things with the new version.

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,41 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size = 8
+        self.length = 0
+        self.dict_ = [None for i in range(self.size)]
```
MasterpieceElbow, line 8: If you don't use the loop variable - use `_` as a name.  
It is better to name it what it is - `hash_table`.

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,2 +1,41 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size = 8
+        self.length = 0
+        self.hash_table = [None for _ in range(self.size)]
+
+    def resize(self) -> None:
+        if len(self) > (self.size * (2 / 3)):
```
MasterpieceElbow, line 11: Don't make 1 more level of depth, it is better to write 
```python
if not len(...):
    return
...
...
```

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #123, opened 2022-10-09
Solution

kateryna3k:


app/main.py
```diff
@@ -1,2 +1,58 @@
...
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = self.make_hash_table()
+
+    def make_hash_table(self) -> list:
+        return [[] for _ in range(self.capacity)]
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __getitem__(self, key: Hashable) -> Any:
```
mltdll, line 17: How does this method behave when the passed key is not in dictionary and there's no collision? Is this a constant time operation in this case?

app/main.py
```diff
@@ -1,2 +1,58 @@
...
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.size = 0
+        self.capacity *= 2
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = self.make_hash_table()
+        for bucket in old_hash_table:
+            if bucket:
+                self.__setitem__(bucket[0], bucket[2])
```
mltdll, line 39: There's a certain shortcut for the `__setitem__()` method you are definitely aware of. You might as well use it.

mltdll (changes requested):

app/main.py
kateryna3k: I don't understand so much about how to make my code better, I think it works pretty well enough, maybe after the mentor review I'll have to change smth, but for now, I think it's ok) thx for the review

mltdll: What I'm talking about in this comment is `self.[bucket[0]] = bucket[2]`. Since it's a magic method, you have no real reason to call it directly, there's a shortcut.

app/main.py
mltdll: Ah, yes, it seems that it is. It's just achieved in a way that is unexpected for me.

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,58 @@
...
+        return [[] for _ in range(self.capacity)]
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __getitem__(self, key: Hashable) -> Any:
+        hash_key = hash(key)
+        hash_index = hash_key % self.capacity
+        while True:
+            try:
```
MasterpieceElbow, line 21: Avoid using `try` if it is not necessary.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #124, opened 2022-10-09
'Solution'

hasslieb3:


app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        self.hash_table = self.make_hash_table()
+        for element in old_hash_table:
+            if element:
+                self.__setitem__(element[0], element[2])
+
+    def __getitem__(self, key: Hashable) -> Any:
+        hash_key = hash(key)
+        hash_index = hash_key % self.capacity
+        while True:
+            try:
```
MasterpieceElbow, line 30: Avoid using `try` if it is not necessary.

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+                if self.hash_table[hash_index][0] == key\
```
MasterpieceElbow, line 31: Avoid using `\`.

PR #126, opened 2022-10-09
Solution!

ketstap162:


app/main.py
```diff
@@ -1,2 +1,109 @@
...
+    def __init__(self) -> None:
+        self._capacity = 8
+        self._load_factor = 2 / 3
+        self._threshold = int(self._capacity * self._load_factor)
+        self._size = 0
+        self._source = [None for _ in range(self._capacity)]
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = hash(key) % self._capacity
+        is_exist = False
```
mltdll, line 23: What do you need this variable for? Can't you just increase `_size` when you set it to `True`?

app/main.py
```diff
@@ -1,2 +1,109 @@
...
+                is_exist = True
+                break
+            if index == self._capacity - 1:
+                index = 0
+            else:
+                index += 1
+        self._source[index] = DictItem(key, value)
+        if not is_exist:
+            self._size += 1
+        if self._size > self._threshold:
```
mltdll, line 35: Resizing is not necessary until the item is added. You can change condition to == and move it to the top of the method.

app/main.py
```diff
@@ -1,2 +1,109 @@
...
+        self._source = [None for _ in range(self._capacity)]
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = hash(key) % self._capacity
+        is_exist = False
+        while (item := self._source[index]) is not None:
+            if item.key == key:
+                is_exist = True
+                break
+            if index == self._capacity - 1:
```
mltdll, line 28: Can't you just use `%` here?

app/main.py
```diff
@@ -1,2 +1,109 @@
...
+        self.iter = 0
+        return self
+
+    def __next__(self) -> Any:
+        while self._source[self.iter] is None:
+            self.iter += 1
+            if self.iter == self._capacity:
+                del self.iter
+                raise StopIteration
+        result = self._source[self.iter]
```
mltdll, line 66: Do you return `DictItem` here? `__next__()` should return a key. Returning type should be Hashable.

app/main.py
```diff
@@ -1,2 +1,109 @@
...
+        self.iter += 1
+        return result
+
+    def __len__(self) -> int:
+        return self._size
+
+    def _resize(self) -> None:
+        self._size = 0
+        items = [self._source[index]
```
mltdll, line 75: You can iterate over `_source` directly.

app/main.py
```diff
@@ -1,2 +1,109 @@
...
+                 for index in range(self._capacity)
+                 if self._source[index] is not None]
+        self._capacity *= 2
+        self._threshold = int(self._capacity * self._load_factor)
+        self._source = [None for _ in range(self._capacity)]
+        for item in items:
```
mltdll, line 81: It looks like you could just copy `_source` earlier and iterate over the list of non-empty nodes, thus avoiding iterating twice.

app/main.py
```diff
@@ -1,2 +1,109 @@
...
+            self[item.key] = item.value
+
+    def clear(self) -> None:
+        self._capacity = 8
```
mltdll, line 85: It's probably better to create a class attribute to store initial capacity to avoid using magic constant twice: here and in `__init__()`.

app/main.py
```diff
@@ -1,2 +1,109 @@
...
+        self._source = [None for _ in range(self._capacity)]
+
+    def pop(self, key: Hashable) -> Any:
+        value = self[key]
+        del self[key]
+        self._size -= 1
+        return value
+
+    def get(self, key: Hashable, default: Any) -> Any:
```
mltdll, line 94: `default` argument should default to `None`.

mltdll (changes requested):

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,104 @@
...
+    _load_factor = 2 / 3
+    _threshold = 5
+
+    def __init__(self) -> None:
+        self._size = 0
+        self.storage = [None for _ in range(self._capacity)]
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            return self[key]
```
Nattalli, line 35: let me understand: do your values on `self.storage`? So what are you doing here?

Nattalli (changes requested):

app/main.py
ketstap162: You call magic method `__getitem__` and if you haven't key in dict, function `get` will return `None`, no exception `KeyError`

Nattalli (approved):
Good job!

PR #127, opened 2022-10-09
Solution

hai-le-di:


app/main.py
```diff
@@ -1,2 +1,66 @@
...
+        for item in temp_table:
+            if len(item) != 0:
+                self.__setitem__(item[0], item[1])
+
+    def __getitem__(self, input_key: Hashable) -> Any:
+        key_hash = hash(input_key)
+        index = key_hash % self.capacity
+        item = self.hash_table[index]
+        while True:
+            try:
```
IvanRamyk, line 43: do not use `try` block here
it's better to use `if` to verify that the cell isn't empty

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+        while True:
+            if len(self.hash_table[index]) == 0:
+                self.hash_table[index] = [key, value, key_hash]
+                self.length += 1
+                break
+            if self.hash_table[index][0] == key and \
```
IvanRamyk, line 20: it's better to compare hashes first. comparing hashes that are represented by integers is faster than comparing complex keys (e.x. tuples)

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+            if len(item) != 0:
+                self.__setitem__(item[0], item[1])
+
+    def __getitem__(self, input_key: Hashable) -> Any:
+        key_hash = hash(input_key)
+        index = key_hash % self.capacity
+        while True:
+            if not self.hash_table[index]:
+                raise KeyError("Key Error")
+            if self.hash_table[index][0] == input_key:
```
IvanRamyk, line 44: check hashes first
it could save a lot of time because comparing of integers is faster than comparing keys (e.x. long tuples)

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+
+    def __len__(self) -> int:
+        return self.length
+
+    def clear(self) -> None:
+        self.hash_table.clear()
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = hash(key) % self.capacity
+        self.hash_table[index] = []
```
IvanRamyk, line 56: if you just set deleted cell to an empty list you will lose the chain and after that, some of the elements at the end of the chain won't be found using the cycle in line 41 due to true value of condition in line 42

IvanRamyk (changes requested):
fix errors described in comments

app/main.py
```diff
@@ -53,7 +55,7 @@ def clear(self) -> None:
 
     def __delitem__(self, key: Hashable) -> None:
         index = hash(key) % self.capacity
-        self.hash_table[index] = []
+        self.hash_table[index] = [None, None, hash(key)]
```
DmytroSvirsa, line 59: what if key dose dot exist?

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #128, opened 2022-10-10
Solution

HStakhiv:


app/main.py
```diff
@@ -1,2 +1,54 @@
+from copy import copy
+from typing import Hashable, Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [[]] * self.capacity
```
mltdll, line 9: You can't do that with list of mutable objects: it will produce a list of lists with the same reference.
```python
>>> x = [[]] * 2
>>> x[0] is x[1]
True
```
Your solution still works because you reassign those list items later, but it's bad practice anyway. Use `[[] for _ in range()]` instead.

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+                self.hash_table[index][1] = value
+                break
+
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, item: Hashable) -> Any:
+        index = hash(item) % self.capacity
+
+        for _ in range(len(self.hash_table)):
+            if self.hash_table[index] and self.hash_table[index][0] == item:
```
mltdll, line 35: You should immediately raise `KeyError` when `self.hash_table[index]` is `[]`, otherwise it's not a constant time operation while it should be.

mltdll (changes requested):

LiudmylaKulinchenko (approved):
Looks good for me 👀

mltdll:
> Looks good for me eyes

:)

app/main.py
HStakhiv: I agree

app/main.py
HStakhiv: You're right

mltdll (approved):

Nattalli (approved):
well done!

PR #129, opened 2022-10-10
ready

maxitoska:


app/main.py
```diff
@@ -1,2 +1,57 @@
...
+
+    def resize_capacity(self) -> None:
+        this_hash_table = self.hash_table
+        self.capacity *= 2
+        self.hash_table = self.create_hash_table()
+        self.length = 0
+        self.threshold = int(self.capacity * 2 / 3)
+
+        for item in this_hash_table:
+            if item:
```
IvanRamyk, line 22:
```suggestion
            if len(item) > 0:
```
> Explicit is better than implicit.

https://peps.python.org/pep-0020/

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        hash_value = hash(key)
+        hash_index = hash_value % self.capacity
+
+        while True:
+            if not self.hash_table[hash_index]:
+                self.hash_table[hash_index] = [key, hash_value, value]
+                self.length += 1
+                break
+            if (self.hash_table[hash_index][0] == key
+                    and self.hash_table[hash_index][1] == hash_value):
```
IvanRamyk, line 38: it's better to compare hashes first because comparing integers could be mush faster than comparing complex objects such as long tuples

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+
+            hash_index = (hash_index + 1) % self.capacity
+
+    def __getitem__(self, key: Hashable) -> Any:
+
+        hash_value = hash(key)
+        hash_index = hash_value % self.capacity
+
+        while self.hash_table[hash_index]:
+            if (self.hash_table[hash_index][0] == key
```
IvanRamyk, line 50: same here

PR #130, opened 2022-10-10
Solution

VadymShkarbul:


app/main.py
```diff
@@ -1,2 +1,46 @@
+from typing import Hashable, Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size = 0
+        self.capacity = 8
+        self.threshold = int(self.capacity * 0.666)
```
mltdll, line 8: I think it's better to use expression `2 / 3` as a constant. It's easier to perceive and also a bit more accurate.

app/main.py
```diff
@@ -1,2 +1,46 @@
...
+                self.size += 1
+                break
+            if self.hash_table[hash_index][0] == hash_key and \
+                    self.hash_table[hash_index][1] == key:
+                self.hash_table[hash_index][2] = value
+                break
+            hash_index = (hash_index + 1) % self.capacity
+
+    def __getitem__(self, key: Hashable) -> Any:
+        for el in self.hash_table:
```
mltdll, line 40: The idea of python dictionary is to have constant time operations for setting and getting values. It doesn't make sense to just use loop here. You should apply the same search principle as in `__setitem__()` method you've defined.

mltdll (changes requested):

app/main.py
VadymShkarbul: Thank you for th review. Yes you are right. It was just the fastest in terms of my own resource at the time ;)

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [None for _ in range(self.capacity)]
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.size = 0
+        self.capacity *= 2
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [None for _ in range(self.capacity)]
+        for el in old_hash_table:
```
Nattalli, line 17: I don't like `el` variable name, but that is not mistake

Nattalli (approved):
Well done!

PR #132, opened 2022-10-10
Solved the task

Cerne13:


app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __setitem__(self, key: Hashable, value: any) -> None:
+        hash_ = hash(key)
+        index_ = hash_ % self.capacity
+
+        while self.items[index_]:
+            if self.items[index_][0] == key and self.items[index_][2] == hash_:
```
FranchukS, line 19: There is no need to use 'and' part
When the first part is True, the second part always will be True either, because hash_ can't be different If keys are equal

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
+        if self.length == int(self.capacity * (2 / 3)):
+            self.resize_hashtable()
+
+    def __getitem__(self, key: Hashable) -> any:
+        hash_ = hash(key)
+        index_ = hash_ % self.capacity
+
+        while self.items[index_]:
+            if self.items[index_][0] == key and self.items[index_][2] == hash_:
```
FranchukS, line 37: Same here

FranchukS (changes requested):

app/main.py
Cerne13: Oh, you're right! Thank you!

FranchukS (approved):
Well done!

app/main.py
```diff
@@ -1,2 +1,53 @@
+from typing import Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 0
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
```
TPolina, line 8: why do you need this attribute if you never use it?

TPolina (changes requested):

app/main.py
Cerne13: Thank you!
Forgot that I decided to use just a calculation later in the code.

TPolina (approved):
Good job!

PR #133, opened 2022-10-10
Solution

shurushku:


VadymShkarbul (approved):
self.table = [None] * self.capacity - nice!

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+                break
+            if self.table[index][0] == key:
+                self.table[index][2] = value
+                break
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: Hashable) -> Any:
+        for elem in self.table:
+            if elem and elem[0] == key:
+                return elem[2]
```
MasterpieceElbow, line 33: What is the point of complicated logic in `__setitem__` if you iterate over the whole list anyway? Remake it with O(1) complexity instead of O(n).

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #134, opened 2022-10-10
Solution

LiudmylaKulinchenko:


HStakhiv (approved):
Everything looks good to me

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+
+        raise KeyError
+
+    def __len__(self) -> int:
+        return self.size
+
+    def resize(self) -> None:
+        threshold = self.capacity * self.LOAD_FACTOR
+
+        if self.size > threshold:
```
MasterpieceElbow, line 43: Don't add another level of depth. It is better to write:
```python
if not ...:
   return
...
...
```

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,2 +1,61 @@
...
+        self.capacity *= 2
+        old_dictionary = self.buckets
+
+        self.buckets = [[] for _ in range(self.capacity)]
+
+        for element in old_dictionary:
+            if element:
+                key, value = element
+                self.__setitem__(key, value)
+                self.size -= 1
```
MasterpieceElbow, line 55: It is not very explicit to decrease self.size for each element. It would be better to assign it to a zero once.

MasterpieceElbow (changes requested):

TPolina (approved):

PR #135, opened 2022-10-10
Solution

matechris:


app/main.py
```diff
@@ -1,2 +1,60 @@
...
+        self.size = 0
+        self.threshold = int(self.capacity * 2 / 3)
+        new_table = self.hash_table.copy()
+        self.hash_table = [None for _ in range(self.capacity)]
+
+        for element in new_table:
+            if element:
+                self.__setitem__(element[0], element[2])
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
HStakhiv, line 22: The `Key` is probably `Hashable` not just `Any`. What do you think?

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+                break
+
+            if (self.hash_table[index][0] == key
+                    and self.hash_table[index][1] == hash_value):
+                self.hash_table[index][2] = value
+                break
+
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: Any) -> Any | KeyError:
```
HStakhiv, line 42: The same

HStakhiv (changes requested):

app/main.py
matechris: right, my bad

HStakhiv (approved):

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+        self.capacity = 8
+        self.size = 0
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [None for _ in range(self.capacity)]
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        self.size = 0
+        self.threshold = int(self.capacity * 2 / 3)
+        new_table = self.hash_table.copy()
```
Nattalli, line 15: `new_table` saves the old table? I think it is a bad name

Nattalli (changes requested):

Nattalli (approved):
Good job!

PR #136, opened 2022-10-11
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -1,2 +1,60 @@
+from typing import Hashable, Any
+
+
 class Dictionary:
-    pass
+    SIZE = 0
+    CAPACITY = 8
```
Nattalli, line 6: why do use `SIZE` and `CAPACITY` as a const? They are the start values when we initialize `Dictionary` class, so it is needed at once. Then it will be changed all the time

Nattalli (changes requested):
You have an awesome work! 😎 
But please, look at my comment and reject or approve it

app/main.py
olia-trofymchuk: i thought that we need to put it as a constant cause it's like an initial value and we always resizing using this standard parameters. 

for example, we always multiply the capacity on load factor which is equal to 2/3.

Nattalli: yes, that's why `load factor` can be a const. But size and capacity are not really needed

Nattalli: if you want to use size and capacity as default value every time when you create a `Dictionary` object - then you can just in the `__init__` method make something like 
```python
self.size = 0
self.capacity = 8
```
And it will be the same

Nattalli (approved):
Good job, but I will notice that before, one of the constants was relevant, so it was even possible to leave it
but even now the solution is good

PR #137, opened 2022-10-11
'Solution'

OvDima1:


app/main.py
```diff
@@ -1,2 +1,56 @@
...
+            self.resize()
+        hash_key = (hash(key))
+        hash_index = hash_key % self.capacity
+
+        while True:
+            if not self.hash_table[hash_index]:
+                self.hash_table[hash_index] = [key, hash_key, item]
+                self.length += 1
+                break
+            elif self.hash_table[hash_index][0] == key and (
```
FranchukS, line 49: Same here with 'and' part

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+            if element:
+                self.__setitem__(element[0], element[2])
+
+    def __getitem__(self, key: Hashable) -> Any:
+        hash_key = hash(key)
+        hash_index = hash_key % self.capacity
+
+        while self.hash_table[hash_index]:
+            if (self.hash_table[hash_index][0] == key
+                    and self.hash_table[hash_index][1] == hash_key):
```
FranchukS, line 33: There is no need to compare hash_keys if the keys are equal
It would be the same without 'and' part

FranchukS (changes requested):

FranchukS (approved):
Well done!

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.hash_table = self.create_hash_table()
+
+    def create_hash_table(self) -> list:
+        return [[] for _ in range(self.capacity)]
+
+    def resize(self) -> None:
+        hash_table = self.hash_table
+        self.length = 0
+        self.capacity *= 2
+        self.threshold = int(self.capacity * self.load_factor)
```
Y-Havryliv, line 20: do you need this line here?

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        while self.hash_table[hash_index]:
+            if self.hash_table[hash_index][0] == key:
+                return self.hash_table[hash_index][2]
+            hash_index = (hash_index + 1) % self.capacity
+        raise KeyError
+
+    def __setitem__(self, key: Hashable, item: Any) -> None:
+        if self.length == self.threshold:
+            self.resize()
+        hash_key = (hash(key))
```
Y-Havryliv, line 40: `hash_key = hash(key)` remove extra parentheses

Y-Havryliv (changes requested):

PR #139, opened 2022-10-11
Solution

skrypka-dmytro:


app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        self.size = 0
+        self.capacity *= 2
+        self.hash_table = [[] for _ in range(self.capacity)]
+        self.threshold = int(self.capacity * self.LOAD_FACTOR)
+
+        for elements in new_hash_table:
+            if elements:
+                self.__setitem__(elements[0], elements[2])
+
+    def get(self, key: Hashable) -> Any:
```
mltdll, line 55: This method is not just a shortcut for `__getitem__()`. It should return a default value instead of raising `KeyError` when key is not found. Default value defaults to `None` and can be passed as an optional `default` argument.

mltdll (changes requested):

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size == self.threshold:
+            self._resize()
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+
+        while True:
+            if not self.hash_table[index]:
+                self.hash_table[index] = [key, hash_key, value]
+                self.size += 1
```
mltdll, line 34: By the way, you may also move call to `_resize()` here (before the assignment) since this is the only case where it's needed.

mltdll (approved):

app/main.py
mltdll: Sorry, you should move `_resize()` back. I didn't think this through before suggesting...

mltdll (changes requested):
Oops, I made a mistake in my advice.

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+
+        for elements in new_hash_table:
+            if elements:
+                self.__setitem__(elements[0], elements[2])
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return default
```
Nattalli, line 59: why do you need this method? You have a `__getitem__()` method, it is enough

Nattalli (changes requested):

app/main.py
mltdll: Sorry for intrusion, but it's int the extra tasks. :)

>
>  you can implement other methods of the dict interface (both regular and magic):
>   - `clear`
>   - `__delitem__`
>   - `get`
>   - `pop`
>   - `update`
>   - `__iter__`

Nattalli: yes, I see now. Thanks!

Nattalli (approved):
Good job!

PR #141, opened 2022-10-11
Solution

taras-andruschenko:


app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        self.length = 0
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.length == self.threshold:
+            self.resize_buckets()
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        while True:
+
+            if not self.buckets[index]:
```
mltdll, line 19: You may move call to `self.resize_buckets()` here, since this is the only case where it's actually needed.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: Hashable) -> Any:
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+
+        while True:
+            if not self.buckets[index]:
+                raise KeyError
+            try:
```
mltdll, line 36: Do you still need this `try ... except` block? It seems to me that you implemented different solution and forgot to remove it.

mltdll (changes requested):
Also, I don't think you should push changes to `point.py`. I would suggest to rollback them.

app/main.py
taras-andruschenko: It's a good idea. Thank you!

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #142, opened 2022-10-11
Solution

Andriy-Makarenko:


mltdll (approved):
Looks good.

app/point.py
```diff
@@ -17,7 +17,7 @@ def __hash__(self) -> int:
         # which is NOT a best practice, but you will be able to predict
         # a hash value by coordinates of the point and its index
         # in the hashtable as well
-        return hash((self.x, self.y))
+        return int(self.x + self.y)
```
MasterpieceElbow, line 20: Don't change Point class!

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+            index = (index + 1) % self.capacity
+        raise KeyError
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size == self.threshold:
+            self.table_resize()
+
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        nodes = self.table
```
MasterpieceElbow, line 41: Why do you save `self.table` into `nodes`?

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.table = [[] for _ in range(self.capacity)]
+
+        for node in hash_table:
+            if node:
+                self[node[0]] = node[2]
+
+    def __getitem__(self, key: Hashable) -> Any:
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        nodes = self.table
```
MasterpieceElbow, line 27: Why do you save `self.table` into `nodes`?

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.load_factor = 2 / 3
+        self.threshold = int(self.capacity * self.load_factor)
+        self.table = [[] for _ in range(self.capacity)]
+
+    def table_resize(self) -> None:
+        hash_table = self.table.copy()
```
MasterpieceElbow, line 14: You don't need to use `.copy()` here.

MasterpieceElbow (changes requested):

app/point.py
Andriy-Makarenko: I don't rememder I changed it. Should I replace int with hash?

app/main.py
Andriy-Makarenko: For me it's easier to perceive a code(logicalwise), and line places in range 79 characters

app/main.py
Andriy-Makarenko: For me it's easier to perceive a code(logicalwise), and line places in range 79 characters

app/main.py
Andriy-Makarenko: Ok

MasterpieceElbow (approved):
I approve but you are not supposed to save attributes into another variable. For long lines use parentheses.

PR #143, opened 2022-10-11
Solution

KatrinLazarenko:


app/main.py
```diff
@@ -1,2 +1,52 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity_table = 8
+        self.threshold_table = int(self.capacity_table * (2 / 3))
+        self.size_table = 0
+        self.hash_table = [[] for i in range(self.capacity_table)]
```
mltdll, line 9: The convention for unused variables in for loops is to call them `_`.

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+    def __len__(self) -> int:
+        return self.size_table
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size_table == self.threshold_table:
+            self.resize_table()
+        hash_key = hash(key)
```
mltdll, line 17: I think it's better to rename this variable into `key_hash` or something like that, since it would be a more natural way to describe it in English.

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        self.fill_table(key, value, hash_key)
+
+    def __getitem__(self, key: Hashable) -> list | None:
+        hash_key = hash(key)
```
mltdll, line 21: Same here.

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        hash_ind = hash_key % self.capacity_table
```
mltdll, line 22: I don't think `ind` is a very good abbreviation. It only saves you two characters anyway. So why don't you rename the variable into `hash_index`? Same at the line 43.

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+    def resize_table(self) -> None:
+        self.capacity_table *= 2
+        self.threshold_table = int(self.capacity_table * (2 / 3))
+        self.size_table = 0
+        prev_table = self.hash_table
+        self.hash_table = [[] for i in range(self.capacity_table)]
+        for prev_item in prev_table:
+            if prev_item:
+                self.__setitem__(prev_item[0], prev_item[1])
```
mltdll, line 38: You may also use the usual interface. `self[prev_item[0]] = prev_item[1]`. I also don't think that `prev_item` is a good name. Just `item` might be better.

mltdll (changes requested):

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #144, opened 2022-10-11
Dictionary

VadimSabadyr:


app/main.py
```diff
@@ -1,2 +1,55 @@
...
+    CAPACITY = 8
+    LOAD_FACTOR = 2 / 3
+
+    def __init__(self) -> None:
+        self.threshold = int(self.CAPACITY * self.LOAD_FACTOR)
+        self.table = [[] for _ in range(self.CAPACITY)]
+        self.length = 0
+
+    def resize_table(self) -> None:
+        table_copy = self.table.copy()
```
mltdll, line 15: You don't need a copy here. Just assign `table` to the temporary variable.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        self.length = 0
+        self.CAPACITY *= 2
+        self.threshold = int(self.CAPACITY * self.LOAD_FACTOR)
+        self.table = [[] for _ in range(self.CAPACITY)]
+        for x in table_copy:
```
mltdll, line 20: `x` is a bad name. It's better to use something like `element`, `item`, or a common abbreviation.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+            if x:
+                self.__setitem__(x[0], x[1])
```
mltdll, line 22: You can use the usual interface for assigning items here, no need to call the magic method directly.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+
+            if self.table[index][0] == key and \
+                    self.table[index][2] == _hash:
+                self.table[index][1] = value
+                return
+
+            index = (index + 1) % self.CAPACITY
+
+    def __getitem__(self, key: Hashable) -> Any:
+        _hash = hash(key)
```
mltdll, line 45: Isn't `_` usually added to the end of the name?

mltdll (changes requested):

mltdll (approved):

Nattalli (approved):
well done!

PR #147, opened 2022-10-11
Solution

Vasyl-Poremchuk:


app/main.py
```diff
@@ -1,2 +1,76 @@
...
+
+        while self.hash_table[index]:
+            key_, value = self.hash_table[index]
+            if key == key_:
+                return value
+            index = (index + 1) % self.capacity
+
+        raise KeyError
+
+    def pop(self, key: Hashable) -> Any:
```
FranchukS, line 62: Pop raises AttributeError ,  because del ("__delitem__") is not defined
And if it did, it would cause problems with indexes - they shift to the side of an empty place

All the other parts look good, (and now I need to refactor my code =))) 
So you can just  "pop" the pop method for approval (as it is not a must-have), or you can try to do it

FranchukS (changes requested):

FranchukS (approved):
👍

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+            self.capacity *= 2
+            old_hash_table = self.hash_table
+
+            self.hash_table = [[] for _ in range(self.capacity)]
+
+            for item in old_hash_table:
+                if item:
+                    key, value = item
+                    self.__setitem__(key, value)
+                    self.length -= 1
```
Y-Havryliv, line 35: you could just make `self.length = 0` before resizing

Y-Havryliv (approved):
GJ!

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.hash_table_index(key)
+
+        while self.hash_table[index]:
+            key_, value_ = self.hash_table[index]
+            if key == key_:
+                self.hash_table[index] = (key, value)
```
Y-Havryliv, line 43: sorry) you need to add hash as well:
`Also, not forgot to store (key, hash, value) as node in hash table.`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #148, opened 2022-10-11
Solution

YevheniiDavydenko:


app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        self.capacity = 8
+        self.size = 0
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [None for _ in range(self.capacity)]
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.size == self.threshold:
+            self.resize()
+        hash_ = hash(key)
+        index_ = hash_ % self.capacity
```
mltdll, line 15: `index` is not a reserved name, you may use it without `_`.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        if self.hash_table[index_] is None:
+            self.hash_table[index_] = (key, hash_, value)
+            self.size += 1
+            return
+        if self.hash_table[index_][0] == key:
+            self.hash_table[index_] = (key, hash_, value)
+            return
+        for _ in range(self.capacity):
```
mltdll, line 23: It's safe to use `whlie True` here. By the way, you may combine the contents of this loop with the code above to avoid code duplication.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [None for _ in range(self.capacity)]
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
KatrinLazarenko, line 11: key value can`t be Any, maybe better to change.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+            index_ = (index_ + 1) % self.capacity
+            if self.hash_table[index_] is None:
+                self.hash_table[index_] = (key, hash_, value)
+                self.size += 1
+                return
+            if self.hash_table[index_][0] == key:
+                self.hash_table[index_] = (key, hash_, value)
+                return
+
+    def __getitem__(self, key: Any) -> Any:
```
KatrinLazarenko, line 33: The same with key annotation

app/main.py
mltdll, line 11: The correct type annotation for `key` is `Hashable`.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        hash_ = hash(key)
+        index_ = hash_ % self.capacity
+        for _ in range(self.capacity):
+            if self.hash_table[index_] is None:
+                raise KeyError
```
mltdll, line 38: Add messages to your errors. In CPython, `__getitem__()` uses the key as message, e.g. `raise KeyError(key)`.

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+        raise KeyError
+
+    def __len__(self) -> int:
+        return self.size
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        self.size = 0
+        self.threshold = int(self.capacity * 2 / 3)
+        old_table = self.hash_table.copy()
```
mltdll, line 51: Don't use `copy()` here. Just assign the `hash_table` to the variable.

KatrinLazarenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+                self.size += 1
+                return
+            if self.hash_table[index_][0] == key:
+                self.hash_table[index_] = (key, hash_, value)
+                return
+
+    def __getitem__(self, key: Any) -> Any:
+        hash_ = hash(key)
+        index_ = hash_ % self.capacity
+        for _ in range(self.capacity):
```
mltdll, line 36: You may rewrite this loop as follows:
```python
while self.hash_table[index] is not None:
```
This will allow you to remove the first check.

mltdll (changes requested):

mltdll (approved):

PR #150, opened 2022-10-12
Solution

SorokaIgor:


app/main.py
```diff
@@ -1,2 +1,69 @@
...
+        return self.load_slots
+
+    def __delitem__(self, key: Hashable) -> Any:
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        while True:
+            if len(self.free_slots[index]) != 0 and\
+                    key == self.free_slots[index][0]\
+                    and self.free_slots[index][1] == hashed_key:
+                self.free_slots[index] = []
```
IvanRamyk, line 64: if you just make it empty then the chain will be broken and get cycle in line 46 can raise an error

app/main.py
```diff
@@ -1,2 +1,69 @@
...
+    def __len__(self) -> int:
+        return self.load_slots
+
+    def __delitem__(self, key: Hashable) -> Any:
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        while True:
+            if len(self.free_slots[index]) != 0 and\
+                    key == self.free_slots[index][0]\
+                    and self.free_slots[index][1] == hashed_key:
```
IvanRamyk, line 63: compare first hash values
it's much faster than comparing complex keys e.x. long tuples 

fix it everywhere

IvanRamyk (changes requested):

IvanRamyk (changes requested):
still haven't fixed

DmytroSvirsa (approved):

PR #151, opened 2022-10-12
Solution

serviza:


Nikyxa (approved):
GJ!

app/main.py
```diff
@@ -1,2 +1,46 @@
...
-    pass
+    def __init__(self) -> None:
+        self.size = 0
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [[] for _ in range(self.capacity)]
+
+    def __setitem__(self, key: int, value: Any) -> None:
+        if self.size == self.threshold:
+            old_hash_table = self.hash_table
```
Y-Havryliv, line 13: it would be better to make another method for resizing

Y-Havryliv (changes requested):

app/main.py
serviza: Ok. I created 'resize' method

PR #152, opened 2022-10-12
Solution

graandda:


app/main.py
```diff
@@ -1,2 +1,61 @@
...
+        self._length = 0
+        self._increase_value = 2
+        self.table = [[] for _ in range(self._capacity)]
+
+    def __len__(self) -> int:
+        return self._length
+
+    def __resize_table(self) -> None:
+
+        table_ = self.table.copy()
```
mltdll, line 18: Don't use `copy()` here, just assign hash table to a temporary variable.
By the way, is `table` a reserved name? Why `table_`?

app/main.py
```diff
@@ -1,2 +1,61 @@
...
+        for cell in table_:
+            if cell:
+                self[cell[0]] = cell[1]
+
+    def __getitem__(self, item: Hashable) -> list or None:
+        hash_ = hash(item)
+        index = hash_ % self._capacity
+
+        while True:
+            if not self.table[index]:
```
mltdll, line 33: If you want, you can move this check to `while` statement, and raise an error upon exit.

app/main.py
```diff
@@ -1,2 +1,61 @@
...
+
+
 class Dictionary:
-    pass
+
+    def __init__(self) -> None:
+        self._capacity = 8
+        self._threshold = int(self._capacity * 2 / 3)
+        self._length = 0
+        self._increase_value = 2
```
mltdll, line 10: What is this attribute for? You don't seem to use it.

mltdll (changes requested):

app/main.py
graandda: my inaccuracy, tnx)

app/main.py
graandda: looks better

app/main.py
graandda: thank you)

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #153, opened 2022-10-13
Solution

mariia-movchan:


app/main.py
```diff
@@ -1,2 +1,59 @@
...
+        self.threshold = int(self.capacity * 2 / 3)
+        self.buckets = [[] for _ in range(self.capacity)]
+        self.size = 0
+
+    def __getitem__(self, key: Hashable) -> Any:
+        hash_item = hash(key)
+        idx = hash_item % self.capacity
+
+        while self.buckets[idx]:
+            if (self.buckets[idx][1] == key
```
FranchukS, line 16: If keys are equal, hash_items can't be different. When first part is True, second part always will be True either.
So there is no need to use 'and' part

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+
+        hash_item = hash(key)
+        idx = hash_item % self.capacity
+
+        while True:
+            if not self.buckets[idx]:
+                self.buckets[idx] = [hash_item, key, value]
+                self.size += 1
+                break
+            if (self.buckets[idx][1] == key
```
FranchukS, line 36: Same about 'and' part

FranchukS (changes requested):

PR #155, opened 2022-10-13
Solution

FranchukS:


mltdll (changes requested):
I can see that you implemented a different approach to collision handling, and I think it's fine, but it would be great if you provided a brief explanation for reviewers. :)

mltdll (approved):

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+        hashed_key = hash(key)
+        index = hashed_key % self._capacity
+
+        if self._hash_table[index]:
+            for cell in self._hash_table[index]:
+                if cell[0] == key:
+                    cell[1] = value
+                    return
+
+        self._hash_table[index].append([key, value, hashed_key])
```
TPolina, line 25: Do you try to add a few elements in one cell in your dict? It's not how the dict works. Or did I not understand sth?

TPolina (changes requested):

app/main.py
FranchukS: example
I've tried to implement kind of chain method for resolving collisions

PR #156, opened 2022-10-13
'Solution'

Ilyakson:


app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        self.table = [[] for _ in range(self.capacity)]
+        self.length = 0
+        self.threshold = int(self.capacity * self.load_factor)
+
+    def __getitem__(self, key: Hashable) -> Any:
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+
+        while self.table[index]:
+            if (self.table[index][0] == key
```
FranchukS, line 17: There is no need to compare hash_keys if the keys are equal
It would be the same without 'and' part

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        if self.length == self.threshold:
+            self.resize()
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+        while True:
+            if not self.table[index]:
+                self.table[index] = [key, value, hash_key]
+                self.length += 1
+                break
+            if (key == self.table[index][0]
```
FranchukS, line 34: Same here with 'and' part

FranchukS (changes requested):

FranchukS (approved):
GJ!

PR #157, opened 2022-10-13
Dictionary

LavrykN:


app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        hash_ = hash(key)
+        index = hash_ % self.capacity
+
+        while True:
+            if not self.table[index]:
+                self.table[index] = [key, value, hash_]
+                self.length += 1
+                return
+
+            if self.table[index][0] == key and self.table[index][2] == hash_:
```
IvanRamyk, line 26: compare hashes first
comparing integers is faster than comparing complex keys, e.x. long tuples

IvanRamyk: fix everywhere

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #158, opened 2022-10-13
dictionary

VitaliyBychevoy:


app/main.py
```diff
@@ -1,2 +1,84 @@
...
+        self.hash_table = [[] for _ in range(8)]
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        threshold = (len(self.hash_table) * 2) // 3
+        if self.length == threshold:
+            self.resize()
+        hash_key = hash(key)
+        index = self.get_index(hash_key)
+
+        if self.hash_table[index]:
```
FranchukS, line 20: You don't need this if/else condition, it all works the same inside "if" block

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+
+            while self.hash_table[index]:
+                if key == self.hash_table[index][0]:
+                    self.hash_table[index][1] = value
+                    return
+                index += 1
+                if index == len(self.hash_table):
+                    index = 0
+            self.hash_table[index] = [key, value, key.__hash__()]
```
FranchukS, line 29: You can use  variable hash_key, no need to use hash function again

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+            self.length += 1
+        else:
+            self.hash_table[index] = [key, value, key.__hash__()]
+            self.length += 1
+
+    def __getitem__(self, key: Any) -> Any:
```
FranchukS, line 35: You need to use hash func as well to find the item because there is no difference with the list in this solution
You can do the same as in setitem method but instead of rewriting value - return it when the keys are the same

FranchukS (changes requested):

app/main.py
VitaliyBychevoy: It does not work without key.__hash__() or hash(key)

app/main.py
FranchukS: This looks better but still has the same logic - you trying to find an item by looking at every element in a list.
Instead, you need to use a hash function to get the index of the element you need to start looking from.

FranchukS (changes requested):

FranchukS (approved):
GJ!

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+
+        while self.hash_table[index]:
+            if (self.hash_table[index][0] == key
+                    and self.hash_table[index][1] == hash_key):
+                return self.hash_table[index][2]
+            index += 1
+            if index == len(self.hash_table):
+                index = 0
+            if index == len(self.hash_table):
+                return self.hash_table[index][2]
```
TPolina, line 41: What is this check for? It is never executed because of 38-39 lines

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+            if item:
+                count += 1
+        return count
+
+    def resize(self) -> None:
+
+        self.length = 0
+        self.capacity *= 2
+        new_list = [[] for _ in range((len(self.hash_table) * 2))]
+        # hash_list_copy = deepcopy(self.hash_table)
```
TPolina, line 63: delete unnecessary comment

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+        hash_list_copy = self.hash_table
+        self.hash_table = new_list
+        for item in hash_list_copy:
+            if item:
+                self.__setitem__(item[0], item[2])
+
+    def total_len(self) -> None:
+        print(len(self.hash_table))
```
TPolina, line 71: what is this method for?

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+
+    def get_index(self, hash_key: int) -> int:
+        return hash_key % self.capacity
+
+    def get_capacity(self) -> int:
+        count = 0
+        for item in self.hash_table:
+            if item:
+                count += 1
+        return count
```
TPolina, line 56: what is this method for? `__len__` does the same

app/main.py
```diff
@@ -1,2 +1,74 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+
+    def __init__(self,) -> None:
+        self.capacity = 8
+        self.threshold = (self.capacity * 2) // 3
```
TPolina, line 8: You don't need this attribute because you never use it

TPolina (changes requested):

app/main.py
```diff
@@ -37,7 +36,6 @@ def __getitem__(self, key: Any) -> Any:
             index += 1
             if index == len(self.hash_table):
                 index = 0
-            if index == len(self.hash_table):
                 return self.hash_table[index][2]
```
TPolina, line 39: this line is also unnecessary because when the index becomes 0 you also need to compare the element under that index with a key and hash and move forward if necessary. In your code, you always return an element under the 0 index without checking it

TPolina (changes requested):

PR #162, opened 2022-10-14
Solution

pavlo-kapitanets:


misha-doroshenko (approved):
Well done

app/main.py
```diff
@@ -1,2 +1,54 @@
-class Dictionary:
-    pass
+from typing import Any, Hashable
+
+
+class Node(object):
+    def __init__(self, key: Hashable, value: Any) -> None:
+        self.key = key
+        self.value = value
```
TPolina, line 7: you need to store hash as well

TPolina (changes requested):

app/main.py
```diff
@@ -29,6 +30,7 @@ def __setitem__(self, key: Hashable, value: Any) -> None:
                 break
             index = (index + 1) % self.initial_capacity
         self.size += 1
+        node.hash = index
```
TPolina, line 33: hash and index is not the same

TPolina (changes requested):

app/main.py
```diff
@@ -44,6 +44,7 @@ def resize(self) -> None:
         for node in table:
             if node is not None:
                 self.__setitem__(node.key, node.value)
+                node.hash = self.get_hash(node.key)
```
TPolina, line 47: you didn't fix the problem: your `get_hash` function returns index and not hash. You just need to use the `hash` function

TPolina (changes requested):

TPolina (approved):

PR #164, opened 2022-10-16
Solution

ShevchukVitalii:


app/main.py
```diff
@@ -1,2 +1,53 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.initial_capacity = 8
```
mltdll, line 6: This attribute is not really properly named. The moment you change it, it's no longer initial. Therefore, I'd suggest to rename it to just `capacity`.

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
+    def resize(self) -> None:
+        prev_hash_table = self.hash_table
+        self.initial_capacity *= 2
+        self.size = 0
+        self.hash_table = [[] for _ in range(self.initial_capacity)]
+        self.threshold = int(self.initial_capacity * 2 / 3)
+        for item in prev_hash_table:
+            if item:
+                self.__setitem__(item[0], item[1])
```
mltdll, line 19: You may use the usual dict assignment interface here, e.g. `self[item[0]] = item[1]`.

mltdll (changes requested):

mltdll (approved):

PR #169, opened 2022-10-18
'Solution'

IvanGLS:


DmytroSvirsa (changes requested):
the idea of this task is to implement your own realization of the dictionary, by using hash and storing it in list

DmytroSvirsa (approved):

PR #173, opened 2022-10-23
Solution

Amirammi:


app/main.py
```diff
@@ -1,2 +1,23 @@
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.hash_table = [[] for _ in range(8)]
+        self.length = 0
+
+    def __setitem__(self, key: object, value: object) -> None:
+        index = hash(key) % 8
+        for data in self.hash_table[index]:
```
Abnormaltype, line 8: This loop takes O(n) time for every search for the missing key. You should optimize it :(

app/main.py
```diff
@@ -1,2 +1,23 @@
...
+            if data[0] == key:
+                data[1] = value
+                return
+        self.hash_table[index].append([key, value])
+        self.length += 1
+
+    def __getitem__(self, key: object) -> None:
+        index = hash(key) % 8
+        for data in self.hash_table[index]:
```
Abnormaltype, line 17: This loop takes O(n) time for every search for the missing key. You should optimize it :(

app/main.py
```diff
@@ -1,2 +1,23 @@
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.hash_table = [[] for _ in range(8)]
+        self.length = 0
+
+    def __setitem__(self, key: object, value: object) -> None:
+        index = hash(key) % 8
+        for data in self.hash_table[index]:
+            if data[0] == key:
```
Abnormaltype, line 9: also, you should check the hash

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+    def __setitem__(self, key: object, value: object) -> None:
+        if self.length == self.treshold:
+            self.resize_hash()
+        index = hash(key) % self.hash_size
+        while True:
+            if not self.hash_table[index]:
+                self.hash_table[index] = [key, value]
+                self.length += 1
+                return
+            if key in self.hash_table[index]:
```
Abnormaltype, line 27: you should also check the hash

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+                self.hash_table[index][1] = value
+                return
+
+            index = (index + 1) % self.hash_size
+
+    def __getitem__(self, key: object) -> object:
```
Abnormaltype, line 33:
```suggestion
    def __getitem__(self, key: object) -> object:
```
you can also return an error

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        index = hash(key) % self.hash_size
+        while self.hash_table[index]:
+            if key in self.hash_table[index]:
```
Abnormaltype, line 36: you should check the hash

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+    def clear(self) -> None:
+        self.hash_size = 8
+        self.treshold = int(self.hash_size * 2 / 3)
+        self.length = 0
+        self.hash_table = [[] for _ in range(self.hash_size)]
+
+    def __delitem__(self, key: object) -> None:
+        index = hash(key) % self.hash_size
+        while self.hash_table[index]:
+            if key in self.hash_table[index]:
```
Abnormaltype, line 53: you should check the hash

Abnormaltype: also, raise an error, when you cannot find the element

Abnormaltype (changes requested):

app/main.py
Amirammi: But I did it at the beginning of this method, should I do the same inside the loop, or just move it from the begging to the loop? Cause, then it will be a duplicate of the code

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+        self.hash_table = [[] for _ in range(self.hash_size)]
+        for item in old_hash_table:
+            if item:
+                self.__setitem__(item[0], item[1])
+
+    def __setitem__(self, key: object, value: object) -> None:
+        if self.length == self.treshold:
+            self.resize_hash()
+        if isinstance(key, (list, dict)):
+            raise TypeError(f"unhashable type: {type(key)}")
```
Abnormaltype, line 22: Unnecessary here

app/main.py
Abnormaltype: you should check the hash, now you check only the key, add another condition to this `if`

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+                self.length += 1
+                return
+            if key in self.hash_table[index]:
+                self.hash_table[index][1] = value
+                return
+            index = (index + 1) % self.hash_size
+
+    def __getitem__(self, key: object) -> object:
+        if self.length == self.treshold:
+            self.resize_hash()
```
Abnormaltype, line 36: Unnecessary here

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+
+    def clear(self) -> None:
+        self.hash_size = 8
+        self.treshold = int(self.hash_size * 2 / 3)
+        self.length = 0
+        self.hash_table = [[] for _ in range(self.hash_size)]
+
+    def __delitem__(self, key: object) -> None:
+        if self.length == self.treshold:
+            self.resize_hash()
```
Abnormaltype, line 55: Unnecessary here

Abnormaltype (changes requested):

app/main.py
Abnormaltype: fix type annotation

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+    def __len__(self) -> int:
+        return self.length
+
+    def clear(self) -> None:
+        self.hash_size = 8
+        self.threshold = int(self.hash_size * 2 / 3)
+        self.length = 0
+        self.hash_table = [[] for _ in range(self.hash_size)]
+
+    def __delitem__(self, key: object) -> None:
```
Abnormaltype, line 54: the same

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+                self.hash_table[index] = [key, key_hash, value]
+                self.length += 1
+                return
+            if self.hash_table[index][0] == key and \
+                    self.hash_table[index][1] == key_hash:
+                self.hash_table[index][2] = value
+                return
+            index = (index + 1) % self.hash_size
+
+    def __getitem__(self, key: (str, int, tuple)) -> object:
```
Abnormaltype, line 38:
```suggestion
    def __getitem__(self, key: (str, int, tuple)) -> object:
```
not correct for key, also you can raise an error

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+    def __len__(self) -> int:
+        return self.length
+
+    def clear(self) -> None:
+        self.hash_size = 8
+        self.threshold = int(self.hash_size * 2 / 3)
+        self.length = 0
+        self.hash_table = [[] for _ in range(self.hash_size)]
+
+    def __delitem__(self, key: (str, int, tuple)) -> None:
```
Abnormaltype, line 57: the same

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+    def __len__(self) -> int:
+        return self.length
+
+    def clear(self) -> None:
+        self.hash_size = 8
+        self.threshold = int(self.hash_size * 2 / 3)
+        self.length = 0
+        self.hash_table = [[] for _ in range(self.hash_size)]
+
+    def __delitem__(self, key: Any) -> None:
```
Abnormaltype, line 57:
```suggestion
    def __delitem__(self, key: Any) -> Union[None, KeyError]:
```

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+                self.hash_table[index] = [key, key_hash, value]
+                self.length += 1
+                return
+            if self.hash_table[index][0] == key and \
+                    self.hash_table[index][1] == key_hash:
+                self.hash_table[index][2] = value
+                return
+            index = (index + 1) % self.hash_size
+
+    def __getitem__(self, key: Any) -> object:
```
Abnormaltype, line 38:
```suggestion
    def __getitem__(self, key: Any) -> Union[object, KeyError]:
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #176, opened 2022-11-07
Solution

Terrrya:


app/main.py
```diff
@@ -1,2 +1,88 @@
...
+            self.nodes.append([])
+        for node in temp:
+            for item in node:
+                if item:
+                    self.__write(item[0], item[1])
+
+    def __setitem__(self,
+                    key: Union[int, float, bool, str, tuple],
+                    value: Any) -> None:
+        if self.nodes.count([]) <= 1 / 3 * len(self.nodes):
```
DmytroSvirsa, line 25: This loop takes O(n) time for every check. You should optimize it :(

app/main.py
```diff
@@ -1,2 +1,88 @@
...
+                return node[1]
+        raise KeyError
+
+    def __len__(self) -> int:
+        result = 0
+        for node in self.nodes:
+            for item in node:
+                if item:
+                    result += 1
+        return result
```
DmytroSvirsa, line 53: This loop takes O(n^2) time for every check.
it returns not the correct result
You should optimize it to work as O(1) :(

app/main.py
```diff
@@ -1,2 +1,88 @@
...
+                key: Union[int, float, bool, str, tuple],
+                value: Any) -> None:
+        write = True
+        for item in self.nodes[hash(key) % len(self.nodes)]:
+            if key in item:
+                item[1] = value
+                write = False
+                break
+        if write:
+            self.nodes[hash(key) % len(self.nodes)].append([key, value])
```
DmytroSvirsa, line 39: what about colision when you have edged situation: hash(key) % len(self.nodes) == len(self.nodes) - 1 and this node is fufiled

app/main.py
```diff
@@ -1,2 +1,88 @@
...
+            for item in node:
+                if item:
+                    result += 1
+        return result
+
+    def clear(self) -> None:
+        self.nodes = []
+
+    def __delitem__(self, key: Any) -> None:
+        for node in self.nodes[hash(key) % len(self.nodes)]:
```
DmytroSvirsa, line 59: len(self.nodes) You should optimize it to work as O(1)

app/main.py
Terrrya: "and this node is fufiled" 
I didn't understand how it can be. Exactly how is node fufiled?

app/main.py
Terrrya: It always returns correct result. But i agree it takes O(nˆ2) time

Terrrya:
Code was passing time-test to these improvments :( I geting mark.time(5) now :(

app/main.py
Terrrya: Done

app/main.py
Terrrya: O(1) done

app/main.py
Terrrya: Done

DmytroSvirsa (approved):

PR #180, opened 2022-11-11
checklist initial

DmytroSvirsa:


checklist.md
```diff
@@ -0,0 +1,32 @@
+# Check Your Code Against the Following Points
+
+## Code Style
+
+Make sure that you provide message when exception rises.
```
Y-Havryliv, line 5: raises

checklist.md
```diff
@@ -0,0 +1,32 @@
...
+## Code Efficiency
+
+### Make sure that your methods work with **O(1)** time complexity.
+
+The exceptions:
+- `__init__` and `clear` methods 
+- handling the collision in any method
+- resizing the dictionary
+
+Bad example:
```
Y-Havryliv, line 16: you forgot a good one

Abnormaltype (changes requested):
add this point:
If you wanna check your `__delitem__` method use these tests:

```
@pytest.mark.timeout(5)
def test_deletion():
    dictionary = Dictionary()
    for key, value in items:
        dictionary[key] = value
    for key, value in items:
    for key, value in items:
        del dictionary[key]
    print(len(dictionary))
```

Also, `delitem` should return nothing.

checklist.md
DmytroSvirsa: it is nof fair to give tham the solution, no?

checklist.md
```diff
@@ -0,0 +1,53 @@
...
+        
+    # in any method that add or delete item we change self.length
+
+    def __len__(self) -> int:
+        return self.length
+```
+
+**Note:** any loop have O(n) time complexity (list comprehension is also a loop)
+
+If you want check your `__delitem__` method
```
Abnormaltype, line 47: Also, delitem should return nothing.

checklist.md
```diff
@@ -0,0 +1,53 @@
...
+        self.length = 0
+        self.hash_table: list = [None] * 8
+        
+    # in any method that add or delete item we change self.length
+
+    def __len__(self) -> int:
+        return self.length
+```
+
+**Note:** any loop have O(n) time complexity (list comprehension is also a loop)
```
Abnormaltype, line 45: add here note about O(1) at best

Abnormaltype (changes requested):

checklist.md
```diff
@@ -0,0 +1,53 @@
...
+```python
+class Dictionary:
+    def __init__(self) -> None:
+        self.hash_table: list = [None] * 8
+
+    def __len__(self) -> int:
+        return len([cell for cell in self.hash_table if cell])
+```
+
+Good example:
```
Y-Havryliv, line 30: good examples are always before bad ones in other theories

Y-Havryliv (approved):

Abnormaltype (approved):

PR #189, opened 2022-12-01
Pytest was passed

Polyakiv-Andrey:


YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+                self.hash_table[index][1] = value
+                break
+
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: int) -> Any:
+        hash_ = hash(key)
+        index = hash_ % self.capacity
+        while self.hash_table[index]:
+            if self.hash_table[index][2] == hash_ and \
```
katevasilenko, line 44: It’s better to break line before `and` https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

katevasilenko (changes requested):

Vasyl-Smutok (approved):
well done!

PR #193, opened 2022-12-08
Solution

the-andrew-007:


app/main.py
```diff
@@ -1,2 +1,45 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 0
+        self.hash_table: list = [None] * 8
+        self.size = len(self.hash_table)
+
+    def __setitem__(
+        self, key: int | str | float | tuple | bool, value: Any
```
Abnormaltype, line 11: what about class and functions? would be better hashable

app/main.py
```diff
@@ -1,2 +1,45 @@
...
+    ) -> None:
+        index = hash(key) % self.size
+        while (
+            self.hash_table[index] is not None
+            and self.hash_table[index][0] != key
```
Abnormaltype, line 16: you should check the input hash as well

app/main.py
```diff
@@ -1,2 +1,45 @@
...
+        if self.length > self.size * (2 / 3):
+            self.resize()
+
+    def __getitem__(self, key: int | str | float | tuple | bool) -> Any:
+        index = hash(key) % self.size
+        try:
+            while self.hash_table[index][0] != key:
+                index = (index + 1) % self.size
+            return self.hash_table[index][1]
+        except TypeError:
```
Abnormaltype, line 31: write without try/except

Abnormaltype (changes requested):

OleksandrPozden (approved):

PR #194, opened 2022-12-09
DONE

DevSpaciX:


YuriiKindrat (changes requested):
All tests should pass - fix the flake8 issues

app/main.py
```diff
@@ -1,2 +1,35 @@
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 0
+        self.size = 8
+        self.hash_table: list = [[] for _ in range(self.size)]
+
+    def __setitem__(self, key: str | int, value: str | int) -> None:
```
YuriiKindrat, line 7: key could be any unmmutable type, value could be any type

app/main.py
```diff
@@ -1,2 +1,35 @@
...
+        if self.hash_table[index]:
+            for item in self.hash_table[index]:
+                if item[0] == key and hash(item[0]) == hash(key):
+                    item[1] = value
+                    return
+        self.hash_table[index].append([key, value])
+        self.length += 1
+
+    def resize(self) -> None:
+        self.size += 2
```
YuriiKindrat, line 18: u should resize it in *2

YuriiKindrat: do u resize self.hash_table?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+                self.hash_list[index][2] = value
+                break
+            index = (index + 1) % self.size
+
+    def resize(self) -> None:
+        self.size *= 2
+        self.critical_size = int(self.size * 2 / 3)
+        self.length = 0
+        old_data = self.hash_list
+        self.hash_list = [[] for _ in range(self.size)]
```
OleksandrPozden, line 32:
```suggestion
        old_data = self.hash_list
        self.hash_list = [[]]*self.size
```

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size = 8
+        self.length = 0
+        self.critical_size = int(self.size * 2 / 3)
+        self.hash_list = [[] for _ in range(self.size)]
+
```
OleksandrPozden, line 10: here the same

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.length == self.critical_size:
```
OleksandrPozden, line 12:
```suggestion
        if self.length > self.critical_size:
```

OleksandrPozden (changes requested):

YuriiKindrat (approved):

PR #196, opened 2022-12-12
Solution Maksym Maksimov

MKeSiMu:


app/main.py
```diff
@@ -1,2 +1,57 @@
...
+            value: Any
+    ) -> None:
+        hashed_value = hash(key)
+        index = hashed_value % self.table_size
+        while self.hash_table[index] is not None:
+            saved_key, saved_hashed_value, saved_value = self.hash_table[index]
+            if saved_key == key and hashed_value == saved_hashed_value:
+                self.hash_table[index] = (key, hashed_value, value)
+                break
+            else:
```
YuriiKindrat, line 22: unnecessary else here

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+
+    def __len__(self) -> int:
+        return self.length
+
+    def resize(self) -> None:
+        if self.length > 2 / 3 * self.table_size:
+            old_hash_table = []
+            for container in self.hash_table:
+                if container is not None:
+                    old_hash_table.append(container)
```
YuriiKindrat, line 52: u can use list comprehension here - it'll be less of code

app/main.py
```diff
@@ -1,2 +1,53 @@
...
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 0
+        self.table_size = 8
+        self.hash_table: list = [None] * self.table_size
+
+    def __setitem__(
+            self,
+            key: (int, float, bool, tuple, str),
```
IvanRamyk, line 12:
```suggestion
            key: Hashable,
```
import from typing

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+            value: Any
+    ) -> None:
+        hashed_value = hash(key)
+        index = hashed_value % self.table_size
+        while self.hash_table[index] is not None:
+            saved_key, saved_hashed_value, saved_value = self.hash_table[index]
+            if saved_key == key and hashed_value == saved_hashed_value:
+                self.hash_table[index] = (key, hashed_value, value)
+                break
+            print(f"The key {key} collided with {self.hash_table[index]}")
```
IvanRamyk, line 22: remove print from the final version

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+    def __setitem__(
+            self,
+            key: (int, float, bool, tuple, str),
+            value: Any
+    ) -> None:
+        hashed_value = hash(key)
+        index = hashed_value % self.table_size
+        while self.hash_table[index] is not None:
+            saved_key, saved_hashed_value, saved_value = self.hash_table[index]
+            if saved_key == key and hashed_value == saved_hashed_value:
```
IvanRamyk, line 19: compare hashes first because comparing integers can be performed much faster than complex keys like tuples

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+                    return self.hash_table[index][2]
+            except TypeError:
+                raise KeyError
+            index = (index + 1) % self.table_size
+
+    def __len__(self) -> int:
+        return self.length
+
+    def resize(self) -> None:
+        if self.length > 2 / 3 * self.table_size:
```
IvanRamyk, line 47: I believe you should put the check into the method that calls resize because when you call the `resize` method you expect that the table will be extended anyway (not only if the size reached some value)

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+            self.resize()
+
+    def __getitem__(
+            self,
+            key: (int, float, bool, tuple, str)
+    ) -> Any:
+        hashed_value = hash(key)
+        index = hashed_value % self.table_size
+        while True:
+            try:
```
IvanRamyk, line 36: do not use `try` `except` where you can just check if the cell is empty.
it makes the code less readable

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #198, opened 2022-12-20
Solution

mish-sk:


app/main.py
```diff
@@ -1,2 +1,48 @@
...
+            hash_key = (hash_key + 1) % self.capacity
+
+    def __getitem__(self, key: Any) -> Any:
+        hash_key = hash(key) % self.capacity
+        while True:
+            if self.hash_table[hash_key]:
+                if self.hash_table[hash_key][0] == key:
+                    return self.hash_table[hash_key][1]
+            else:
+                raise KeyError
```
OleksandrPozden, line 33: don't use nested ```if's```
And remove ```else``` as it is redundant

OleksandrPozden (changes requested):

YuriiKindrat (approved):

PR #200, opened 2022-12-26
Solution

Vanya2389:


app/main.py
```diff
@@ -1,2 +1,46 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8, length_table: int = 0) -> None:
```
YuriiKindrat, line 5: remove unnecessary arguments from __init__ - user shouldn't be able to create exp of a dict with capacity and lenght_table more then initial
image

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #206, opened 2023-01-08
Solution

evgenijmartynuk07:


katevasilenko (changes requested):
You can’t use built-in dict in this task. Please implement hash table with its capacity and reindexing features

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+                    self.load += 1
+                    return
+            for ind in range(len(self.items)):
+                if self.items[ind] is None:
+                    self.items[ind] = [key, hash(key), value]
+                    self.load += 1
+                    break
+
+    def __getitem__(self, key: Any) -> Any:
+        if key == "missing_key":
```
katevasilenko, line 39: How do you know this? This checking is bad idea

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+            raise KeyError
+        for index in self.items:
+            if index is not None and key == index[0]:
+                return index[-1]
+
+    def reload(self) -> None:
+        create_new = [None] * len(self.items)
```
katevasilenko, line 46: As you are using `len(self.items) ` several times does it make sense to set it as an attribute of class?

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+                    create_new[index] = item
+                if create_new[index] is not None:
+                    count = index
+                    for ind in range(len(self.items[index:]) - 1):
+                        count += 1
+                        if create_new[count] is None:
+                            create_new[count] = item
+                            return
+                    if create_new[count] != item:
+                        for ind in range(len(self.items)):
```
katevasilenko, line 60: It is too many nested loops so your code has higher complexity than should. Please read checklist

katevasilenko (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #207, opened 2023-01-08
Dictionary

MarynaProkhorenko:


app/main.py
```diff
@@ -1,2 +1,47 @@
...
+                self.hash_table[index][2] = value
+                break
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: str) -> Any:
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+        while self.hash_table[index] is not None:
+            if self.hash_table[index][0] == hash_key \
+                    and self.hash_table[index][1] == key:
```
olena-hul, line 32: Use `()` to split a line

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+        while True:
+            if self.hash_table[index] is None:
+                self.hash_table[index] = [hash_key, key, value]
+                self.length += 1
+                break
+            if self.hash_table[index][1] == key \
+                    and self.hash_table[index][0] == hash_key:
```
olena-hul, line 22: It's better to use parentheses `()` instead of `\` to split a line

olena-hul (changes requested):

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+        while True:
+            if self.hash_table[index] is None:
+                self.hash_table[index] = [hash_key, key, value]
+                self.length += 1
+                break
+            if (
+                self.hash_table[index][1] == key
+                and self.hash_table[index][0] == hash_key
```
IvanRamyk, line 23:
```suggestion
                self.hash_table[index][0] == hash_key
                and self.hash_table[index][1] == key
```

IvanRamyk: it's better to check hash values first because comparing of the integers could be much faster than complex keys such as tuples. and if the first check fails the second would be skipped

IvanRamyk (changes requested):

IvanRamyk (approved):

app/main.py
MarynaProkhorenko: yes, of course

PR #210, opened 2023-01-11
1st try sol

AnnaKabatova:


app/main.py
```diff
@@ -1,2 +1,45 @@
...
+        if self.length > self.capacity * 2 / 3:
+            self.resize()
+        index = hash(key) % self.capacity
+        if self.hash_table[index]:
+            for item in self.hash_table[index]:
+                if item[0] == key and hash(item[0]) == hash(key):
+                    item[1] = value
+                    return
+        self.hash_table[index].append([key, value])
+        self.length += 1
```
Serhii-Leonenko, line 23: It doesn't work with collision when we have different keys with the same hash, use a while loop and change index when the cell is not empty. Also, it is better to add a hash with the key and value in the list.

app/main.py
```diff
@@ -1,2 +1,45 @@
...
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        self.hash_table *= 2
+        temporary_list = []
+        for storage in self.hash_table:
+            for item in storage:
+                temporary_list.append(item)
+        for item in temporary_list:
+            self.__setitem__(*item)
```
Serhii-Leonenko, line 33: remove the redundant loop here, just make a copy of your hash table, create a new hash table with new capacity, and then iterate through values in the copied table and use setitem for every non empty values

app/main.py
```diff
@@ -1,2 +1,45 @@
...
+
+    def __getitem__(self, key: str | int) -> list:
+        index = hash(key) % self.capacity
+        if not self.hash_table[index]:
+            raise KeyError
+        for item in self.hash_table[index]:
+            if item[0] == key:
+                return item[1]
```
Serhii-Leonenko, line 41: also, fix this method according to cases when the collision happens.

Serhii-Leonenko (changes requested):

OleksandrPozden (approved):

PR #220, opened 2023-01-25
Solution

yuliia-stopkyna:


app/main.py
```diff
@@ -1,2 +1,63 @@
...
+            if hash_node is not None:
+                key, hash_num, value = hash_node
+                self.hash_table[index] = None
+                self.add_hash_node(key, value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        try:
+            hash_slot = self.search_key(key)
+            self.hash_table[hash_slot] = (key, hash(key), value)
+        except KeyError:
```
katevasilenko, line 36: It’s not clear to me, how you can got KeyError while setting values. You.  Need to change this logic

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+            for index, node in enumerate(self.hash_table):
+                if node is not None and node[0] == key:
+                    return index
+        raise KeyError("Key is not in a dictionary")
+
+    def __getitem__(self, key: Any) -> Any:
+        try:
+            hash_slot = self.search_key(key)
+        except KeyError:
+            raise KeyError("Key is not in a dictionary")
```
katevasilenko, line 60: It is a bad style to catch exception and than raise it. You need to change this logic

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+        except KeyError:
+            self.size += 1
+            if self.size <= self.capacity * self.loadfactor:
+                self.add_hash_node(key, value)
+            else:
+                self.resize_hash_table()
+                self.add_hash_node(key, value)
+            self.keys.append(key)
+
+    def search_key(self, key: Any) -> int:
```
katevasilenko, line 45: I don’t get this method. What the Logic you try to realise here?

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+        else:
+            for i in range(self.capacity):
+                if self.hash_table[i] is None:
+                    self.hash_table[i] = (key, hash(key), value)
+                    break
+
+    def resize_hash_table(self) -> None:
+        self.hash_table += [None] * self.capacity
+        self.capacity = self.capacity * 2
+        for index, hash_node in enumerate(self.hash_table):
```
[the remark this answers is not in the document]
yuliia-stopkyna: why all nodes are None? no, I add additional None nodes to the old table: `self.hash_table += [None] * self.capacity`

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+        self.index += 1
+        return self.keys[index]
+
+    def get(self, key: Any, value: Any = None) -> Any:
+        try:
+            self.__getitem__(key)
+        except KeyError:
+            if value is None:
+                return None
+            self.__setitem__(key, value)
```
ArsenPidhoretskyi, line 82: It is not logical. See screenshot
image

ArsenPidhoretskyi: result will be:
b
b
But must return
b
None

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+            self.hash_table[hash_slot] = (key, hash_num, value)
+        else:
+            while self.hash_table[hash_slot] is not None:
+                if hash_slot == self.capacity - 1:
+                    hash_slot = 0
+                else:
+                    hash_slot += 1
+            self.hash_table[hash_slot] = (key, hash_num, value)
+
+    def resize_hash_table(self) -> None:
```
ArsenPidhoretskyi, line 26: it must be private or protected method

app/main.py
```diff
@@ -1,2 +1,89 @@
...
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.loadfactor = 2 / 3
+        self.size = 0
+        self.hash_table = [None] * self.capacity
+        self.keys = []
+
+    def add_hash_node(self, key: Any, value: Any) -> None:
```
ArsenPidhoretskyi, line 13: It must be private or protected method

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+        hash_num = hash(key)
+        hash_slot = hash_num % self.capacity
+        if self.hash_table[hash_slot] is None:
+            self.hash_table[hash_slot] = (key, hash_num, value)
+        else:
+            while self.hash_table[hash_slot] is not None:
+                if hash_slot == self.capacity - 1:
+                    hash_slot = 0
+                else:
+                    hash_slot += 1
```
ArsenPidhoretskyi, line 23: use inline if

ArsenPidhoretskyi: Also, as I see it possible to never end, yeah?

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+                self.add_hash_node(key, value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        hash_num = hash(key)
+        try:
+            self.__getitem__(key)
+        except KeyError:
+            self.size += 1
+            if self.size <= self.capacity * self.loadfactor:
+                self.add_hash_node(key, value)
```
ArsenPidhoretskyi, line 42: Duplicate code with line 45. Move if after if

ArsenPidhoretskyi (changes requested):

app/main.py
yuliia-stopkyna: why never end? __add_hash_node method is executed only after checking the capacity or resizing - the None is always present, I think

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+        self.size = 0
+        self.hash_table = [None] * self.capacity
+        self.keys = []
+
+    def __add_hash_node(self, key: Any, value: Any) -> None:
+        hash_num = hash(key)
+        slot = hash_num % self.capacity
+        if self.hash_table[slot] is None:
+            self.hash_table[slot] = (key, hash_num, value)
+        else:
```
ArsenPidhoretskyi, line 18: This code is redundant. Can you try if all will work without this? Just while

app/main.py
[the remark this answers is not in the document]
ArsenPidhoretskyi: @katevasilenko, no. It works right

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+                self.hash_table[index] = None
+                self.__add_hash_node(key, value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        hash_num = hash(key)
+        try:
+            self.__getitem__(key)
+        except KeyError:
+            self.size += 1
+            if self.size > self.capacity * self.loadfactor:
```
ArsenPidhoretskyi, line 38: create a property for `self.size > self.capacity * self.loadfactor`. It will increase readability 
```
@property
def is_need_resize(self):
    return self.size > self.capacity * self.loadfactor
```

app/main.py
ArsenPidhoretskyi: As for me, it's okay. Because first, we need to check if key exist in dictionary before set it

ArsenPidhoretskyi (changes requested):

olena-hul (approved):

PR #231, opened 2023-02-06
Solution

6aPa6aH:


app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
+        self.resize = 2
+        self.length = 0
+        self.hash = [[] for _ in range(self.capacity)]
+
+    def __len__(self) -> int:
+        return self.length
+
+    def get_index_and_hash(self, key: Any) -> Any:
```
katevasilenko, line 15:
```suggestion
    def get_index_and_hash(self, key: Hashable) -> Any:
```
Please fix all over the project

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        index = hash(key) % self.capacity
+        hash_of_item = hash(key)
+        return hash_of_item, index
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.length > self.threshold:
+            self.resize_table()
+        hash_of_item, index = self.get_index_and_hash(key)
```
katevasilenko, line 23: Good idea!

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        node = [key, hash_of_item, value]
```
katevasilenko, line 24: I would prefer hash at the first place

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
+        self.resize = 2
+        self.length = 0
+        self.hash = [[] for _ in range(self.capacity)]
```
katevasilenko, line 10: It better to call hash_table or something like that to avoid misunderstanding

katevasilenko (changes requested):

vasylhnatiuk (approved):

PR #232, opened 2023-02-06
'Solution'

timyr4iq:


app/main.py
```diff
@@ -1,2 +1,96 @@
...
+        while (self.hash_table[index] is not None
+               or self.hash_table[index] == 1):
+            if self.hash_table[index] == 1 or self.hash_table[index][0] == key:
+                self.length -= 1
+                break
+            index = (index + 1) % self.capacity
+        self.hash_table[index] = (key, value)
+        self.length += 1
+        if len(self) > self.load_factor * self.capacity:
+            self.capacity *= self.resize
```
IvanRamyk, line 23: let's move this logic to the separate method

IvanRamyk: e.x. `resize`

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+        self.load_factor = 2 / 3
+        self.resize = 2
+        self.length = 0
+        self.hash_table = [None] * self.capacity
+
+    def __setitem__(self, key: Callable, value: Callable) -> None:
+        index = hash(key) % self.capacity
+        while (self.hash_table[index] is not None
+               or self.hash_table[index] == 1):
+            if self.hash_table[index] == 1 or self.hash_table[index][0] == key:
```
IvanRamyk, line 16: also, compare hashes first - this operation performs faster because it's comparing not complex keys (tuples) but integers

IvanRamyk (changes requested):

app/main.py
timyr4iq: I didn’t quite understand what I should compare earlier - the conditions in the if or something else?

Vasyl-Smutok (approved):
Well Done 👍

PR #233, opened 2023-02-08
Solution

ilushasm:


app/main.py
```diff
@@ -1,2 +1,99 @@
...
+            position += 1
+
+            if position > self.hash_table_threshold - 1:
+                position = 0
+
+        if self.length == self.hash_table_threshold:
+            self.rehash()
+
+    def rehash(self) -> None:
+        hash_table = self.hash_table
```
katevasilenko, line 33: I would prefer something like old_hash_table to avoid possible misleading

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+        while True:
+            if self.hash_table[position] is None:
+                raise KeyError
+
+            if self.hash_table[position][0] == item:
+                return self.hash_table[position][2]
+
+            count += 1
+            if count == self.hash_table_threshold:
+                raise KeyError
```
katevasilenko, line 56: Why?

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+        self.hash_table_capacity *= 2
+        self.hash_table_threshold = round(self.hash_table_capacity * (2 / 3))
+        self.hash_table = [None] * self.hash_table_capacity
+        self.length = 0
+
+        for item in hash_table:
+            if item is not None:
+                self.__setitem__(item[0], item[2])
+
+    def __getitem__(self, item: Any) -> Any:
```
katevasilenko, line 43: Key would be better name for variable here

katevasilenko (changes requested):

app/main.py
ilushasm: Loop checks each cell in hash table, if key doesn't match it goes to the next cell. If I will remove count check than in cases when key atribute is not in the dict - cycle will run without ending. Count variable counts how many cell were checked, if it's bigger then len of hash table (meaning cycle checked all cells in the table and hasn't found the key) it raises error

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+            if position > self.hash_table_capacity - 1:
+                position = 0
+
+        if self.length > self.hash_table_threshold:
+            self.rehash()
+
+    def rehash(self) -> None:
+        old_hash_table = self.hash_table
+        self.hash_table_capacity *= 2
+        # self.hash_table_threshold = round(self.hash_table_capacity * (2 / 3))
```
Vasyl-Smutok, line 36: remove comment

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+        self.hash_table_capacity = 8
+        self.hash_table_threshold = 5
+
+    def __delitem__(self, key: Any) -> None:
+        position = hash(key) % self.hash_table_capacity
+        count = 0
+
+        while True:
+            if self.hash_table[position] is not None \
+                    and self.hash_table[position][0] == key:
```
Vasyl-Smutok, line 81:
```suggestion
            if (
                self.hash_table[position] is not None
                and self.hash_table[position][0] == key
            ):
```

Vasyl-Smutok (approved):
Good 👍 
also check comment

PR #235, opened 2023-02-14
solution

mkobylianskyi:


app/main.py
```diff
@@ -1,2 +1,51 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 0
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [[] for _ in range(self.capacity)]
```
Grayder0152, line 9: May better to create class `Node` and save (`key, value, hash`) as argument, and for empty cells use `None`

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+                self.hash_table[index][2] == hash_
+                and self.hash_table[index][0] == key
+            ):
+                return self.hash_table[index][1]
+            index = (index + 1) % self.capacity
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self.length
+
```
Grayder0152, line 42: I recomend implement some of this methods 
image

Grayder0152 (changes requested):

app/main.py
mkobylianskyi: what do you mean?

app/main.py
mkobylianskyi: thought not necessarily

katevasilenko (approved):
Looks good to me

PR #240, opened 2023-02-19
Solution

YehorKorn:


app/main.py
```diff
@@ -1,2 +1,123 @@
...
+
+        self._buckets = [[] for i in range(self._bucket_size)]
+        self._len_bucket = 0
+        self._assign_buckets(current_buckets)
+
+    def __setitem__(self, _key: Any, _value: Any) -> None:
+        self._len_bucket = len(self) + 1
+
+        if self._len_bucket > self._bucket_resize:
+            self._resize([(_key, _value)])
```
ArsenPidhoretskyi, line 54: it is not so good to do resize + add new element in one method
Must be Single Responsibility for method

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+        while self._buckets[index]:
+            _key, _value = self._buckets[index]
+            if _key == input_key:
+                if _value == "remote":
+                    raise KeyError(f'There is no such key: "{input_key}"')
+                return _value
+            index = (index + 1) % self._bucket_size
+
+    def __delitem__(self, input_key: Any) -> None:
+        self._assign_buckets([(input_key, "remote")])
```
ArsenPidhoretskyi, line 74: Why on delete we add key remote?

ArsenPidhoretskyi:
```
d = Dictionary()
d["key"] = "123"
print(d)
del d["key"]
print(d)
```
Test on this

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+            index = hashed_value % self._bucket_size
+
+            while self._buckets[index]:
+                if self._buckets[index][0] == _key:
+                    break
+                index = (index + 1) % self._bucket_size
+
+            self._buckets[index] = (_key, _value,)
+
+    def _get_buckets_full(self) -> list:
```
Bohdan-Salamakha, line 35: add additional typehint for list inner objects

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+        dict_repr = "{"
+        for _key, _value in self._get_buckets_full():
+            dict_repr += f"{_key}: {_value}, "
+        dict_repr += "}"
+        return dict_repr
+
+    def __len__(self) -> int:
+        return len(self._get_buckets_full())
+
+    def keys(self) -> list:
```
Bohdan-Salamakha, line 93: add additional typehint for list inner objects

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+        keys_list = []
+        for _key, _value in self._get_buckets_full():
+            keys_list.append(_key)
+        return keys_list
+
+    def values(self) -> list:
```
Bohdan-Salamakha, line 99: add additional typehint for list inner objects

Bohdan-Salamakha (changes requested):
v1

app/main.py
YehorKorn: If we delete this element and leave an empty cell, then the other values that had a collision with this cell will be lost

YehorKorn: Okay, I fixed it.

ArsenPidhoretskyi: If we remove the value we must lost this value totally

app/main.py
```diff
@@ -1,2 +1,131 @@
...
+        if elements is None:
+            elements = []
+        self._bucket_size = 8
+        self._bucket_resize = self._bucket_size * 2 / 3
+        while len(elements) > self._bucket_resize:
+            self._bucket_size *= 2
+            self._bucket_resize = self._bucket_size * 2 / 3
+        self._buckets = [[] for i in range(self._bucket_size)]
+        if len(elements):
+            self._assign_buckets(elements)
```
katevasilenko, line 16: Your init method looks a bit mess and it is hard to read. Why just don’t made all the checks into the methods below?

app/main.py
```diff
@@ -1,2 +1,131 @@
...
+
+    def _assign_buckets(self, elements: List[tuple]) -> None:
+        for _key, _value in elements:
+            if isinstance(_key, list | dict | set):
+                raise KeyError(
```
katevasilenko, line 21: TypeError is better here

app/main.py
```diff
@@ -1,2 +1,131 @@
+
```
katevasilenko, line 1: No need this white line

app/main.py
```diff
@@ -1,2 +1,131 @@
...
+    def _resize(
+            self,
+            new_element: List[tuple],
+            len_new_elements: int = 0
+    ) -> None:
+        current_buckets = self._get_buckets_full()
+        current_buckets += new_element
+
+        while len(self) + len_new_elements > self._bucket_resize:
```
katevasilenko, line 48: `len(self)` looks a bit strange. Does that make sense to set elements as an attribute of instance?

app/main.py
```diff
@@ -1,2 +1,131 @@
...
+        else:
+            self._assign_buckets(elements)
+
+
+d = Dictionary()
+d["key"] = "123"
+print(d)
+del d["key"]
+print(d)
+print(len(d))
```
katevasilenko, line 131: Please remove debugging from the final version

katevasilenko (changes requested):

app/main.py
YehorKorn: I don't understand what exactly needs to be done

app/main.py
YehorKorn: During initialization, a check must be performed to create the table, and then everything else.

Or you can do it another way?

Grayder0152: @katevasilenko , menched that you can create some method to check and then call it into init to reduce code in init method

app/main.py
Grayder0152: Create class attribute as `self.lenght=0` and than, after added or removed element from Dictionary change this attr

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,2 +1,123 @@
+from typing import List, Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, elements: List[tuple] = None) -> None:
```
Grayder0152, line 5:
```suggestion
    def __init__(self, elements: Optional[List[tuple]] = None) -> None:
```

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+                    break
+                index = (index + 1) % self._bucket_size
+
+            self._buckets[index] = (_key, _value,)
+            self.length += 1
+
+    def _get_buckets_full(self) -> list[Any]:
+        return [
+            full_bucket for full_bucket in self._buckets
+            if full_bucket and full_bucket[1] != KeyError
```
Grayder0152, line 42: What is this `full_bucket[1] != KeyError`?

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+        ]
+
+    def _resize(
+            self,
+            new_element: List[tuple],
+            len_new_elements: int = 0
```
Grayder0152, line 48: Lets get length of `new_element` automatically

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+        while self._buckets[index]:
+            _key, _value = self._buckets[index]
+            if _key == input_key:
+                if _value == KeyError:
+                    raise KeyError(f'There is no such key: "{input_key}"')
+                return _value
+            index = (index + 1) % self._bucket_size
+
+    def __delitem__(self, input_key: Any) -> None:
+        self._assign_buckets([(input_key, KeyError)])
```
Grayder0152, line 77: You must clear this `key` and `value` from `_buckets` not change value to `KeyError`

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+
+    def __repr__(self) -> str:
+        dict_repr = "{"
+        for _key, _value in self._get_buckets_full():
+            dict_repr += f"{_key}: {_value}, "
+        dict_repr += "}"
+        return dict_repr
+
+    def __len__(self) -> int:
+        return len(self._get_buckets_full())
```
Grayder0152, line 94: Why not return `self.length`?

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+
+    def keys(self) -> list[Any]:
+        keys_list = []
+        for _key, _value in self._get_buckets_full():
+            keys_list.append(_key)
+        return keys_list
```
Grayder0152, line 100: In this case you can use list comprehension

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+
+    def values(self) -> list[Any]:
+        values_list = []
+        for _key, _value in self._get_buckets_full():
+            values_list.append(_value)
+        return values_list
```
Grayder0152, line 106: The same as above

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+
+    def clear(self) -> None:
+        self._buckets = [[] for i in range(self._bucket_size)]
+
+    def get(self, _key: Any) -> Any:
+        _value = None
+        try:
+            _value = self.__getitem__(_key)
```
Grayder0152, line 114:
```suggestion
            _value = self[_key]
```

Grayder0152 (changes requested):

app/main.py
YehorKorn: It is for delitems

app/main.py
YehorKorn: this way the setitem will work faster

app/main.py
YehorKorn: Because that's the only way the tests go

app/main.py
YehorKorn: If we clear them, we will lose the other cells that were displaced by the collision

app/main.py
ArsenPidhoretskyi: It is wrong solution then
If something will be delated then it must be as start value

app/main.py
ArsenPidhoretskyi: But then you can not correct synced varialbes

app/main.py
ArsenPidhoretskyi: It is wrong again. If key delete we must delete this key-value not set KeyError
It is possible to do

ArsenPidhoretskyi: Now you are just making crutches, in real work it will be rejected immediately

app/main.py
ArsenPidhoretskyi: Because you out of sync your buckets and self.length
No need to make crutches!

app/main.py
```diff
@@ -1,2 +1,119 @@
...
+    def __getitem__(self, input_key: Any) -> Any:
+        hashed_value = hash(input_key)
+        index = hashed_value % self._bucket_size
+        if input_key not in self.keys():
+            raise KeyError(f'There is no such key: "{input_key}"')
+
+        while self._buckets[index]:
+            _key, _value = self._buckets[index]
+            if _key == input_key:
+                if _value == KeyError:
```
ArsenPidhoretskyi, line 71: One more crutches

ArsenPidhoretskyi (changes requested):

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+                self._buckets[index] = (_key,)
+                self.length -= 1
+
+            self._buckets[index] = (_key, _value,)
+            self.length += 1
+
+    def _get_buckets_full(self) -> list[Any]:
+        return [
+            full_bucket for full_bucket in self._buckets
+            if len(full_bucket) == 2
```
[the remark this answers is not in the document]
YehorKorn: because there may be cells that have been deleted and have a length of 1

app/main.py
```diff
@@ -1,2 +1,135 @@
...
+            index = (index + 1) % self._bucket_size
+        if not self._buckets[index]:
+            raise KeyError(f'There is no such key: "{_key}"')
+        self._buckets[index] = (_key,)
+        self.__length -= 1
+
+    def _get_buckets_full(self) -> list[Any]:
+        return [
+            full_bucket for full_bucket in self._buckets
+            if len(full_bucket) == 2
```
ArsenPidhoretskyi, line 61: BTW, You can create a method for example:
```
@staticmethod
def __is_bucket_filled(bucket):
    return len(bucket) == 2
```

ArsenPidhoretskyi: Reuse it in all places where == 1 or == 2

app/main.py
ArsenPidhoretskyi: Maybe resize will just resize it without adding new element?

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #242, opened 2023-02-20
Solution

Zoriana-Dvulit:


app/main.py
```diff
@@ -1,2 +1,110 @@
...
+
+    def clear(self) -> None:
+        self.table = [None] * self.size
+        self.count = 0
+
+    def get(self, key: Any, default: Any = None) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return default
```
Bohdan-Salamakha, line 79: use get() method to avoid KeyError without handling error

app/main.py
```diff
@@ -1,2 +1,110 @@
...
+
+    def __len__(self) -> int:
+        return self.count
+
+    def __contains__(self, key: Any) -> bool:
+        try:
+            self[key]
+            return True
+        except KeyError:
+            return False
```
Bohdan-Salamakha, line 52: use get() method to avoid KeyError without handling error

app/main.py
```diff
@@ -1,2 +1,110 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Any, default: Any = None) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
```
Bohdan-Salamakha, line 87: use get() method to avoid KeyError without handling error

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+            if node.key == key:
+                return node.value
+            node = node.next
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self.count
+
+    def __contains__(self, key: Any) -> bool:
+        return True if self.get(key) else False
```
ArsenPidhoretskyi, line 48:
```suggestion
        return bool(self.get(key))
```

ArsenPidhoretskyi (approved):
As for me looks so cool)

app/main.py
Zoriana-Dvulit: Thanks for the improvement issue , I adjusted it in my code. I committed the changes, and again pushed them to GitHub.

PR #243, opened 2023-02-21
Solution

SafonovVladimir:
add my Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,2 +1,146 @@
+from __future__ import annotations
+from typing import Any, Hashable
+
+INITIAL_CAPACITY = 8
+LOAD_FACTOR = 2 / 3
+RESIZE_FACTOR = 2
+
+
+class Node:
+    def __init__(self, key: Hashable, value: Any) -> None:
+        self.key = key
+        self.value = value
+        self.hash_key = hash(key)
+
+
 class Dictionary:
-    pass
+
+    def __init__(self, **kwargs: Any) -> None:
+        self.length = 0
+        self.dict_keys: list = []
+        self.dict_values: list = []
+        self.capacity: int = INITIAL_CAPACITY
+        self.hash_table: list = [None] * self.capacity
+        for key, value in kwargs.items():
+            self.__setitem__(key, value)
+
+    def __str__(self) -> str:
+        result_str = "{"
+
+        for i in range(len(self.hash_table)):
+
+            if self.hash_table[i] is not None:
+                result_str += (f"'{self.hash_table[i].key}': "
+                               f"{self.hash_table[i].value}, ")
+
+        return f"{result_str}"[:-2] + "}"
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        node = Node(key, value)
+        index = node.hash_key % self.capacity
+
+        if (index == len(self.hash_table)
+                and self.hash_table[index] is not None
+                and self.hash_table[index].key != key):
+            index = 0
+
+        while (self.hash_table[index] is not None
+               and self.hash_table[index].key != key):
+            index = (index + 1) % self.capacity
+
+        if self.hash_table[index] is None:
+            self.length += 1
+
+        self.hash_table[index] = node
+
+        if self.length / self.capacity >= LOAD_FACTOR:
+            self._resize()
+
+    def __getitem__(self, key: Hashable) -> Any:
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+
+        while (self.hash_table[index] is not None
+               and self.hash_table[index].key != key):
+            index = (index + 1) % self.capacity
+
+        if self.hash_table[index] is None:
+            raise KeyError(key)
+
+        return self.hash_table[index].value
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __delitem__(self, key: Hashable) -> None:
+        for i in range(self.length):
+            if (self.hash_table[i] is not None
+                    and self.hash_table[i].key == key):
+                self.hash_table[i] = None
+
+    def keys(self) -> Any:
+        for i in range(len(self.hash_table)):
+            if self.hash_table[i] is not None:
+                self.dict_keys.append(self.hash_table[i].key)
+        return self.dict_keys
+
+    def values(self) -> Any:
+        for i in range(len(self.hash_table)):
+            if self.hash_table[i] is not None:
+                self.dict_values.append(self.hash_table[i].value)
+        return self.dict_values
+
+    def items(self) -> list:
+        return [(self.hash_table[i].key, self.hash_table[i].value) for i in
+                range(len(self.hash_table)) if self.hash_table[i] is not None]
+
+    def clear(self) -> None:
+        self.hash_table = []
+        self.dict_keys = []
+        self.dict_values = []
+        self.length = 0
+
+    def get(self, key: Hashable) -> Any:
+        return self.__getitem__(key)
+
+    def pop(self, _key: Any) -> Any:
+        for i in range(len(self.hash_table)):
+
+            if (self.hash_table[i] is not None
+                    and self.hash_table[i].key == _key):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,2 +1,146 @@
...
+
+    def pop(self, _key: Any) -> Any:
+        for i in range(len(self.hash_table)):
+
+            if (self.hash_table[i] is not None
+                    and self.hash_table[i].key == _key):
+                return self.hash_table[i].value
+
+    def update(self, new_dict: Dictionary) -> Dictionary:
+        for key, value in new_dict.items():
```
bihunva, line 114: just add "self[key] = value" here and  remove other lines below. 
it will work the same way because u already defined "__setitem__"
no need to return anything

bihunva (changes requested):

app/main.py
SafonovVladimir: Thanks, it works.

bihunva: this method can be simplified further (in 3 lines) :)
just add "self[key] = value" to line 116, and everything else can be deleted
no need to return anything, just update current dict)

SafonovVladimir: Done

app/main.py
```diff
@@ -1,2 +1,146 @@
...
+            self.__setitem__(key, value)
+
+    def __str__(self) -> str:
+        result_str = "{"
+
+        for i in range(len(self.hash_table)):
+
+            if self.hash_table[i] is not None:
+                result_str += (f"'{self.hash_table[i].key}': "
+                               f"{self.hash_table[i].value}, ")
```
NEkropo1: You can simply iterate by non empty node, instead making range object for whole hash_table and iterating on table +few refactoring for more maintenance(instead reloading string whole time) and easier reading(in my opinion):
        items = []
        for node in self._hash_table:
            if node is not None:
                items.append(f"'{node.key}': {node.value}")
        return '{' + ', '.join(items) + '}'

app/main.py
```diff
@@ -1,2 +1,146 @@
...
+        return self.hash_table[index].value
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __delitem__(self, key: Hashable) -> None:
+        for i in range(self.length):
+            if (self.hash_table[i] is not None
+                    and self.hash_table[i].key == key):
+                self.hash_table[i] = None
```
NEkropo1: In this method, you could use the same loop condition as in the __getitem__ method to find the slot that contains the key, instead of iterating over the entire hash table, like this + it's good to raise exception(not sure if code works, its just fastprint):
def delitem(self, key):
        key_hash = hash(key)
        index = key_hash % self._capacity

        while (self._hash_table[index] is not None
               and self._hash_table[index].key != key):
            index = (index + 1) % self._capacity

        if self._hash_table[index] is None:
            raise KeyError(key)

        self._hash_table[index] = None
        self._length -= 1

app/main.py
```diff
@@ -1,2 +1,146 @@
...
+
+    def get(self, key: Hashable) -> Any:
+        return self.__getitem__(key)
+
+    def pop(self, _key: Any) -> Any:
+        for i in range(len(self.hash_table)):
+
+            if (self.hash_table[i] is not None
+                    and self.hash_table[i].key == _key):
+                return self.hash_table[i].value
```
NEkropo1: I guess it's good to raise a KeyError, if you don't find the key, not return None )))

NEkropo1 (changes requested):
My code is just an idea, not working sample, so it's on you if you want to implement something like this

app/main.py
SafonovVladimir: Yep, it's simplier.
Thank's.

app/main.py
SafonovVladimir: +

app/main.py
SafonovVladimir: Fixed it using magic method __delitem__ and return KeyError when key not found.
Thank's

NEkropo1 (approved):
looks ok )

app/main.py
```diff
@@ -1,2 +1,136 @@
...
+    def update(self, new_dict: Dictionary) -> None:
+        for key, value in new_dict.items():
+            self[key] = value
+
+    def __iter__(self) -> Any:
+        for i in range(len(self._hash_table)):
+            if self._hash_table[i] is not None:
+                yield self._hash_table[i]
+
+    def _resize(self) -> None:
```
YuriiKindrat, line 122: u can use setitem method here, just make a copy from ur old table and make a new clear table for dict, than use setitem method to fill it

app/main.py
```diff
@@ -1,2 +1,136 @@
...
+                self._hash_table[i] is not None]
+
+    def clear(self) -> None:
+        self.hash_table = []
+        self._dict_keys = []
+        self._dict_values = []
+        self._length = 0
+
+    def get(self, key: Hashable) -> Any:
+        return self.__getitem__(key)
```
YuriiKindrat, line 104: image

app/main.py
```diff
@@ -1,2 +1,136 @@
...
+
+    def pop(self, _key: Any) -> Any:
+        if self._hash_table[hash(_key) % self._capacity] is None:
+            raise KeyError(_key)
+        key_value = self._hash_table[hash(_key) % self._capacity].value
+        self.__delitem__(_key)
+        return key_value
```
YuriiKindrat, line 111: image

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: Thank's, it's easier and DRY.

app/main.py
SafonovVladimir: Done!

app/main.py
SafonovVladimir: Done

Morty67 (approved):

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, _key: Any, default: Any = None) -> Any:
+        try:
+            key_value = self[_key]
+            self.__delitem__(key_value)
+            return key_value
+        except KeyError:
+            return default
```
wQuelS, line 114: It returns default if `default` is provided. In other cases `KeyError` is raised

wQuelS (changes requested):

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+
+    def update(self, new_dict: Dictionary) -> None:
+        for key, value in new_dict.items():
```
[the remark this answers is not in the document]
SafonovVladimir: Done

app/main.py
SafonovVladimir: Fixed

app/main.py
```diff
@@ -1,2 +1,140 @@
...
+
+    def clear(self) -> None:
+        self._hash_table = []
+        self._dict_keys = []
+        self._dict_values = []
+        self._length = 0
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        # If key don't find in dictionary __getitem_raise a KeyError
+        return self.__getitem__(key)
```
wQuelS, line 107: `get()` never throws a key error:) That's why this method is so usefull

app/main.py
```diff
@@ -1,2 +1,140 @@
...
+
+    def pop(self, key: Any, default: Any = None) -> Any:
+        # If key don't find in dictionary __getitem_raise a KeyError
+        key_value = self[key]
+        self.__delitem__(key)
+        return key_value
```
wQuelS, line 113: what if default value is provided?

wQuelS (changes requested):

app/main.py
SafonovVladimir: Yep, I think I figured out.
Please, check my solution this methods

YuriiKindrat (approved):

PR #245, opened 2023-02-22
Solution

AlimFreight:


app/main.py
```diff
@@ -1,2 +1,53 @@
...
+                break
+            if (
+                self.table[index][0] == key
+                and self.table[index][1] == current_hash
+            ):
+                self.table[index][2] = value
+                break
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: Any) -> list:
```
Vasyl-Smutok, line 40:
```suggestion
    def __getitem__(self, key: Hashable) -> list:
```

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        self.capacity *= 2
+        self.threshold = int(self.capacity * self.load_factor)
+        self.size = 0
+        old_table = self.table
+        self.table = [None] * self.capacity
+        for cell in old_table:
+            if cell:
+                self.__setitem__(cell[0], cell[2])
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
Vasyl-Smutok, line 22:
```suggestion
    def __setitem__(self, Hashable: Any, value: Any) -> None:
```

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,2 +1,53 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 10, load_factor: float = 0.75) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
```
katevasilenko, line 7: Does that make sense to just multiply by 0.75 as this value never changes instead of creating attribute?

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        self.threshold = int(self.capacity * self.load_factor)
+        self.size = 0
+        self.table = [[]] * self.capacity
```
katevasilenko, line 10: Why do you need empty list in empty list?

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+                self.__setitem__(cell[0], cell[2])
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size > self.threshold:
+            self.resize()
+        current_hash = hash(key)
+        index = current_hash % self.capacity
+        while True:
+            if not self.table[index]:
+                self.table[index] = [key, current_hash, value]
```
katevasilenko, line 29: It’s picky but I would prefer do not separate key and value

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        self.threshold = int(self.capacity * self.load_factor)
+        self.size = 0
+        self.table = [[]] * self.capacity
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        self.threshold = int(self.capacity * self.load_factor)
+        self.size = 0
+        old_table = self.table
+        self.table = [None] * self.capacity
```
katevasilenko, line 17: Please use this way in init too

katevasilenko (changes requested):

katevasilenko (approved):
Looks good to me now

PR #246, opened 2023-02-22
solution

avkpol:


app/main.py
```diff
@@ -1,2 +1,47 @@
...
+    def __len__(self) -> int:
+        return self.size
+
+    def __setitem__(self, key: Any, value: Any) -> Any:
+        index = hash(key) % self.capacity
+        treshold = self.capacity * 2 / 3
+        while self.node[index] is not None:
+            if self.node[index].key == key:
+                self.node[index].value = value
+                return key, value
```
katevasilenko, line 20: You don’t need to return in set method

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+            index = (index + 1) % self.capacity
+        if self.size > treshold:
+            self.capacity *= 2
+            items = [item for item in self.node if item is not None]
+            self.size = 0
+            self.node = [None] * self.capacity
+            for item in items:
+                self.__setitem__(item.key, item.value)
```
katevasilenko, line 28: Please remove this to the separate method

app/main.py
```diff
@@ -1,2 +1,47 @@
...
-    pass
+    def __init__(self) -> None:
+        self.size: int = 0
+        self.capacity: int = 8
+        self.node = [None] * self.capacity
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __setitem__(self, key: Any, value: Any) -> Any:
```
katevasilenko, line 14: Key is Hashable

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+            self.size = 0
+            self.node = [None] * self.capacity
+            for item in items:
+                self.__setitem__(item.key, item.value)
+            self.__setitem__(key=key, value=value)
+            return key, value
+        self.size += 1
+        self.node[index] = Node(key=key, value=value, hash_=hash(key))
+
+    def __getitem__(self, key: Any) -> None:
```
katevasilenko, line 34: Key is hashable

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+            index = (index + 1) % self.capacity
+        if self.size > treshold:
+            self.capacity *= 2
+            items = [item for item in self.node if item is not None]
+            self.size = 0
+            self.node = [None] * self.capacity
+            for item in items:
+                self.__setitem__(item.key, item.value)
+            self.__setitem__(key=key, value=value)
+            return key, value
```
katevasilenko, line 30: No need to return here

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,50 @@
+import dataclasses
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size: int = 0
+        self.capacity: int = 8
+        self.node = [None] * self.capacity
```
ArsenPidhoretskyi, line 9: make node private in this case because it possible that someone set value
and then if self.node will be full your code stop worker
Or inside __setitem__ do check for size before setting

app/main.py
```diff
@@ -1,2 +1,50 @@
...
+        if self.size > treshold:
+            self.resize_hashtable()
+            self.__setitem__(key=key, value=value)
+            return
+        self.size += 1
+        self.node[index] = Node(key=key, value=value, hash_=hash(key))
+
+    def resize_hashtable(self) -> None:
+        self.capacity *= 2
+        items = [item for item in self.node if item is not None]
```
ArsenPidhoretskyi, line 31: use filter + list

ArsenPidhoretskyi (changes requested):

app/main.py
avkpol: `use filter + list` what does it mean?

Grayder0152:
```suggestion
        items = list(filter(lambda item: item is not None,self.node))
```

Grayder0152: but, for my opinion,  it is not necessary

Grayder0152 (approved):

avkpol:
Thank you!

PR #247, opened 2023-02-22
Solution

WDemidenko:


Danil228 (approved):
GJ

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+        if self.hash_table[index] is None:
+            raise KeyError(f"There is no element with this key: {key}")
+        elif self._check_key(key, index):
+            return self.hash_table[index][1]
+        else:
+            while self.hash_table[index] and not self._check_key(key, index):
+                index = (index + 1) % self.hash_capacity
+            if self.hash_table[index] or self._check_key(key, index):
+                return self.hash_table[index][1]
+            raise KeyError(f"There is no element with this key: {key}")
```
YuriiKindrat, line 56: make ur code DRY here, try to avoid KeyError dublication

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+                    if self.hash_table[i] and self._check_key(key, i):
+                        self.hash_table[i] = None
+        else:
+            raise KeyError(f"Wrong key: {key}")
+
+    def get(self, key: hashable) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return
```
YuriiKindrat, line 82: are u sure about this method, what about default value ?

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+
+    def pop(self, key: hashable, *args: str) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            if not args:
+                raise
+            else:
+                default, = args
+                return default
```
YuriiKindrat, line 92: try to avoid using else here, also are u sure about *args as arg ? maybe this is should only one param ?
image

app/main.py
```diff
@@ -1,2 +1,100 @@
...
+
+    def update(self, key: hashable, value: Any) -> None:
+        self[key] = value
```
YuriiKindrat, line 95: are u sure about arguments for this method, what if there will be more than 1 key,value pair ?

YuriiKindrat (changes requested):

wQuelS (approved):
Nice

PR #248, opened 2023-02-22
try this lil dict

NEkropo1:


app/main.py
```diff
@@ -1,2 +1,94 @@
...
+    def __len__(self) -> int:
+        return self.count
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.count / self.capacity >= self.load_factor:
+            self._resize()
+
+        index = self._hash(key) % self.capacity
+        while self.table[index] is not None and self.table[index][0] != key:
+            index = (index + 1) % self.capacity
```
WDemidenko, line 25: again, "is not None" could be removed

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+
+    def update(self, other: Mapping[Hashable, Any]) -> None:
+        for key, value in other.items():
+            self[key] = value
+
+    def _find_key(self, key: Hashable) -> int:
+        index = self._hash(key) % self.capacity
+        while self.table[index] is not None and self.table[index][0] != key:
+            index = (index + 1) % self.capacity
+        return index if self.table[index] is not None else None
```
WDemidenko, line 71: here as well

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+
+    def _resize(self) -> None:
+        self.capacity *= 2
+        new_table = [None] * self.capacity
+        for item in self.table:
+            if item is not None:
```
WDemidenko, line 77: and here

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+                index = item[1] % self.capacity
+                while (
+                    new_table[index] is not None
```
WDemidenko, line 80: and here

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+        self.table = new_table
+
+    def _hash(self, key: Hashable) -> int:
+        return hash(key)
+
+    def items(self) -> list:
+        return [(node[0], node[2]) for node in self.table if node is not None]
+
+    def clear(self) -> None:
+        self.table.clear()
```
WDemidenko, line 94: Why dont you setting self.count to 0?

WDemidenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+    def __init__(self, capacity: int = 8, load_factor: float = 0.75) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.count = 0
+        self.table = [None] * self.capacity
+
+    def __iter__(self) -> Any:
+        for item in self.table:
+            if item is not None:
+                yield item[0]
```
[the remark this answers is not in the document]
NEkropo1: What if item value == False?

app/main.py
NEkropo1: again

app/main.py
NEkropo1: yup

app/main.py
NEkropo1: and here

app/main.py
NEkropo1: and there

app/main.py
NEkropo1: Agree, missed that point.

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+
+    def pop(self, key: object = None, default: Any = None) -> Any:
+        if key is None:
+            raise ValueError(
+                "pop() method with `Dictionary` works only with defined key"
+            )
+
+        index = self._find_key(key)
+        if index is None:
+            return default
```
wQuelS, line 48: image

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+        else:
+            value = self.table[index][2]
+            self.table[index] = None
+            self.count -= 1
+            return value
+
+    def update(self, other: Mapping[Hashable, Any]) -> None:
+        for key, value in other.items():
+            self[key] = value
```
wQuelS, line 57: image

wQuelS (changes requested):
What about `get()`? Yeah, it is not obligatory, but I'm asking, 'cause you've implemented almost all additional methods..

app/main.py
NEkropo1: I'm sorry, but where you found that info? Here is my: https://docs.python.org/3.11/library/stdtypes.html#dict.update
image

wQuelS: Your source is correct, but update() for dict can also take `tuples`, for example. And tuples don't have `.items()`

YuriiKindrat (approved):

PR #249, opened 2023-02-24
Solution 1.0

ZabFTFT:


viktoria-rybenchuk (approved):

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+            elif element[0] == key:
+                return element[-1]
+        raise KeyError("Dictionary doesn't have any value with provided key!")
+
+    def get(self, key: Hashable) -> Any:
+        for element in self.data:
+            if element is None:
+                continue
+            elif element[0] == key:
+                return element[-1]
```
wQuelS, line 77: dict's `get()` returns `None` if there is no such key. You also have `__getitem__` which u can use here

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+
+    def __len__(self) -> int:
+        return self.size
+
+    def pop(self, key: Hashable, default: str = "Default value") -> Any | None:
```
wQuelS, line 82: let's make `default = None` as it implemented in built-in dictionary

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+                    continue
+                elif self.data[i][0] == key:
+                    pop_item = self.get(key)
+                    self.data[i] = None
+                    self.size -= 1
+                    return pop_item
+        return default
+
+    def __delitem__(self, key: Hashable) -> None:
+        for i in range(len(self.data)):
```
wQuelS, line 95: `i` is a bad variable name. Also I would use `enumerate` instead of `range`

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+        for element in iterable:
+            self.__setitem__(element[0], element[-1])
+
+    def __iter__(self) -> Dictionary:
+        self.index = -1
+        return self
+
+    def __next__(self) -> tuple:
+        self.index += 1
+        return self.data[self.index]
```
wQuelS, line 114: what if this method reaches the end of the dictionary?

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+CAPACITY_MULTIPLIER = 2
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = INITIAL_CAPACITY
+        self.size = 0
+        self._load_factor_calculation()
+        self.data = []
```
wQuelS, line 17: not really good to set this value to an empty list. Can you use one of your methods here?

wQuelS (changes requested):

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+            elif element[0] == key:
+                return element[-1]
+
+    def get(self, key: Hashable, default: None = None) -> Any:
+        for element in self.data:
+            if element is None:
+                continue
+            elif element[0] == key:
+                return element[-1]
+        return default
```
wQuelS, line 75: Why don't you just use your `__getitem__` ?

app/main.py
```diff
@@ -1,2 +1,120 @@
...
+        if self.__getitem__(key):
+            for i in range(len(self.data)):
+                if self.data[i] is None:
+                    continue
+                elif self.data[i][0] == key:
+                    pop_item = self.get(key)
+                    self.data[i] = None
+                    self.size -= 1
+                    return pop_item
+        return default
```
wQuelS, line 90: image

wQuelS (changes requested):

app/main.py
ZabFTFT: I thought we shouldn't raise KeyError in get function as I remember from docs.But my get_item method did that so from my perspective it was better to rewrite whole block then to use get_item ib try except =)

wQuelS: I bet you know how to catch errors:))

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+            elif element[0] == key:
+                return element[-1]
+
+    def get(self, key: Hashable, default: None = None) -> Any:
+        try:
+            value = self.__getitem__(key)
+        except KeyError:
+            return default
+        else:
+            return value
```
wQuelS, line 75: just `return` the value which is `__getitem__` returns to you when you call it. I mean, you don't need an additional variable here. Also, you will be able to get rid of 74 and 75 lines

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+                "Dictionary doesn't have any value with provided key!"
+            )
+
+        for element in self.data:
+            if element is None:
+                continue
+            elif element[0] == key:
+                return element[-1]
+
+    def get(self, key: Hashable, default: None = None) -> Any:
```
wQuelS, line 69: I guess `default` should be `Any`. What do you think?

wQuelS (changes requested):

app/main.py
ZabFTFT: Totaly agree, my bad.

PR #251, opened 2023-02-26
Solution

SergiiMachulin:


inrock1 (approved):

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+            if node is None:
+                self.hash_table[next_index] = Node(key, value, hash_value)
+                self.length += 1
+                if self.length > self._capacity * self._load_factor:
+                    self._resize()
+                return
+            elif node.key == key:
+                new_node = Node(key, value, hash_value)
+                self.hash_table[next_index] = new_node
+                return
```
YuriiKindrat, line 30: use break instead of blank returns here

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
```
YuriiKindrat, line 99: image
fix this

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+
+    def update(self, other: dict) -> None:
+        for key, value in other.items():
+            self[key] = value
```
YuriiKindrat, line 103: update also should work with ur class and interable with tuples inside

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        index = hash_value % self._capacity
+        for range_index in range(self._capacity):
+            next_index = (index + range_index) % self._capacity
+            node = self.hash_table[next_index]
+            if node is None:
+                raise KeyError(f"Key '{key}' not found")
+            if node.key == key:
+                self.hash_table[next_index] = None
+                self.length -= 1
+                self._rehash()
```
YuriiKindrat, line 63: no need to rehash after deleting item

YuriiKindrat (changes requested):

app/main.py
SergiiMachulin: The rehashing is still necessary because the deletion may cause the hash table to become unbalanced, and this can lead to hash collisions.

For example, consider a hash table where all keys have the same hash value, and several of them are deleted. In this case, many slots in the hash table will be set to None, and the remaining key-value pairs will cluster together in a small portion of the hash table. This can cause long search times and hash collisions, which can severely degrade performance.

By rehashing the hash table, we can distribute the remaining key-value pairs more evenly across the hash table, reducing the chance of collisions and improving performance.

Serhii-Leonenko: Agree, but I think you should change `_rehash()` method implementation, you create a new instance of the dictionary but in the result, you just store attributes from it. You already have the method with similar logic `_resize()` but with capacity changing. Try to unite these both methods into one universal for both cases

Serhii-Leonenko (changes requested):

wQuelS (approved):
GJ

PR #252, opened 2023-02-26
'Solution'

inrock1:


ostapT (approved):

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+            self.resize()
+        index = hash(key) % len(self.table)
+        bucket = self.table[index]
+        for i, (k, v) in enumerate(bucket):
+            if k == key:
+                bucket[i] = (key, value)
+                break
+        else:
+            bucket.append((key, value))
+            self.size += 1
```
YuriiKindrat, line 25: what about collisions ?

YuriiKindrat (changes requested):

app/main.py
inrock1: my variant works great with collisions, although not like others
image

inrock1: my 'bucket' is a mutable list, with 2 elements tuples.
I keep collisions there and have fast and simple access to them.
it is another way implement all requirements of the task

Serhii-Leonenko: do you know what collision is?) And where is the collision on your screenshot?)

inrock1:
> do you know what collision is?) And where is the collision on your screenshot?)

Collision is two different keys or objects have the same index from the hash.
In screenshot - data with keys (1, 9, 25) have index 1 and fall into the same cell.
But not overwritten. Stored separately and give correct values on request.

app/main.py
YuriiKindrat: it is a good way to implement the task, u are saving a few nodes in one bucket in case of collisions, but it is not how dict work in python ((
image

YuriiKindrat (changes requested):

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,50 @@
...
+        while self.hash_table[index] and self.hash_table[index].key != key:
+            index = (index + 1) % self.capacity
+        return index
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.calculate_index(key)
+        if self.hash_table[index] is None:
+            if self.size >= self.load_factor * self.capacity:
+                self.resize()
+                return self.__setitem__(key, value)
```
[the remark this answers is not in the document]
inrock1: it makes sense. Done.

Serhii-Leonenko (approved):
Well done!

PR #255, opened 2023-02-27
Solution

MaybeTI:


app/main.py
```diff
@@ -1,2 +1,75 @@
...
+        try:
+            item = self[key]
+            del self[key]
+            return item
+        except KeyError:
+            raise KeyError
+
+    def update(self, other: dict) -> None:
+        for key, value in other.items():
+            self[key] = value
```
YuriiKindrat, line 75: where is default args for pop\get methods ? update method should work with custom Dictionary classs/ default python class and iterable with tuples inside

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+        except TypeError:
+            raise KeyError
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __index__(self, key: Any) -> int:
+        index = hash(key) % self.capacity
+        while self.hash_table[index] is not None\
+                and self.hash_table[index][0] != key:
```
YuriiKindrat, line 36: avoid using linebreakers - use brackets instead

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Any, default: Any = None) -> Any:
+        try:
+            item = self[key]
+            del self[key]
+            return item
+        except KeyError:
+            return default
```
wQuelS, line 71: image
What if `default` is not provided?

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+
+    def update(self, other: dict) -> None:
+        for key, value in other.items():
+            self[key] = value
```
wQuelS, line 75: `update()` should work if `other` is also a `tuple`
image

wQuelS (changes requested):

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+
+    def clear(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.hash_table = [None] * self.capacity
+
+    def __delitem__(self, key: Any) -> None:
+        if self.hash_table[hash(key) % self.capacity] is None:
+            raise KeyError
+        self.hash_table[hash(key) % self.capacity] = None
```
Serhii-Leonenko, line 57: what about collisions, will it work correctly?

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+        self.capacity = 8
+        self.size = 0
+        self.hash_table = [None] * self.capacity
+
+    def __delitem__(self, key: Any) -> None:
+        index = self.__index__(key)
+        if self.hash_table[index] is None:
+            raise KeyError
+        self.hash_table[index] = None
+        self.size -= 1
```
Serhii-Leonenko, line 59: Imagine that there was a collision with keys and you have deleted one of them after that you want to get a value of the second key but you can't, you will get KeyError. Try to think why and fix)

Serhii-Leonenko (changes requested):

PR #257, opened 2023-02-27
Solution to Dict

Maytreyaya:


DmytryTong (approved):
Really hard task - good job)

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+        for item in self.hash_table:
+            if item is not None:
+                index = item[1] % self.capacity
+                while (
+                        new_table[index] is not None
+                        and new_table[index][0] != item[0]
+                ):
+                    index = (index + 1) % self.capacity
+                new_table[index] = item
+        self.hash_table = new_table
```
YuriiKindrat, line 27: u can use setitem method here, just copy old table to a variable

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+            index = (index + 1) % self.capacity
+        if self.hash_table[index] is None:
+            raise KeyError(key)
+        return self.hash_table[index][2]
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = hash(key) % self.capacity
+        if index is not None:
+            self.hash_table[index] = None
+            self.count -= 1
```
YuriiKindrat, line 58: what about collisions?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,58 @@
...
+            raise KeyError(f"Cannot find value for key: {key}")
+        return self.hash_table[index][2]
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = hash(key) % self.capacity
+        if index is not None:
+            self.hash_table[index] = None
+            self.length -= 1
+        else:
+            raise KeyError(f"{key} is not in {self}")
```
YuriiKindrat, line 58: u should fix this method, cause it didn't work with collisions

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+            index = (index + 1) % self.capacity
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+        return self.hash_table[index][2]
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = hash(key) % self.capacity
+        while (self.hash_table[index] is not None
+               and self.hash_table[index][0] != key):
+            index = (index + 1) % self.capacity
```
Serhii-Leonenko, line 56: First problem that you have code duplicating (the same code in the getitem method). Another problem is that when you make your cell None and you had a collision here before when you wrote keys with the same hash for example you can't get one of these keys with your getitem method, try to think why.

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #258, opened 2023-02-27
'Solution'

oksanaaam:


app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        self.hash_table = [None] * self.capacity
+        self.length = 0
+
+    def pop(self, key: Any, default: None = None) -> None:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
```
YuriiKindrat, line 72: image
what about KeyError?

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+
+    def update(self, other_dict: dict) -> None:
+        for key, value in other_dict.items():
+            self[key] = value
```
YuriiKindrat, line 76: make it possible to work with ur dict, also it should work with iterable with tuples inside

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,84 @@
+from __future__ import annotations
```
wQuelS, line 1: Do you really need it? If yes, consider to update your python version. This is not a case for reject but still..

wQuelS (changes requested):

app/main.py
```diff
@@ -1,2 +1,84 @@
...
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8, load_factor: float = 0.67) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.threshold = int(self.capacity * self.load_factor)
+        self.length = 0
+        self.hash_table = [None] * self.capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
wQuelS, line 13: Also, are u sure `key` can be `Any in a dict?

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+
+    def clear(self) -> None:
+        self.hash_table = [None] * self.capacity
+        self.length = 0
+
+    def pop(self, key: Any) -> None:
+        try:
+            self.pop(key)
+        except KeyError:
+            return
```
wQuelS, line 70: image

app/main.py
```diff
@@ -63,22 +62,27 @@ def clear(self) -> None:
...
-        try:
-            self.pop(key)
-        except KeyError:
-            return
+    def pop(self, key: Hashable, default: None = None) -> None:
+        if self[key]:
+            value = self[key]
+            del self[key]
+            return value
+        else:
```
wQuelS, line 70: do you really need this `else` statement after `return`? It will never go to this line if statement in line66 is True

wQuelS (changes requested):

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+        self.hash_table = [None] * self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = hash(key) % len(self.hash_table)
+        while self.hash_table[index] and self.hash_table[index][0] != key:
+            index = (index + 1) % len(self.hash_table)
+        if self.hash_table[index] and self.hash_table[index][0] == key:
+            self.hash_table[index] = (key, hash(key), value)
+        else:
+            self.hash_table[index] = (key, hash(key), value)
```
Serhii-Leonenko, line 19: no need to use every time hash() function, use it before the loop and write it into variable

wQuelS (approved):
Good

PR #260, opened 2023-02-27
'Solution'

Morty67:


oigod (approved):

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+            if not self.table[current_index]:
+                self.table[current_index] = [key, value, hashed]
+                self.size += 1
+                return
+            if (
+                key == self.table[current_index][0]
+                and hashed == self.table[current_index][2]
+            ):
+                self.table[current_index][1] = value
+                return
```
YuriiKindrat, line 29: no need to use blank return here, u can use break instead

app/main.py
```diff
@@ -1,2 +1,55 @@
...
+                and self.table[current_index][2] == hashed
+            ):
+                return self.table[current_index][1]
+        raise KeyError
+
+    def __len__(self) -> int:
+        return self.size
+
+    def resize(self) -> None:
+        new_copy = deepcopy(self.table)
```
YuriiKindrat, line 49: Don't use deepcopy() here. Just assign the hash_table to the variable.

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #261, opened 2023-02-27
solution

DavidGlusch:


app/main.py
```diff
@@ -1,2 +1,47 @@
...
+            else:
+                raise KeyError("Duplicate key")
+
+    def __getitem__(self, key: Any) -> Any:
+        index = self._get_index(key)
+        if self.table[index] is not None:
+            stored_key, _, stored_value = self.table[index]
+            if key == stored_key:
+                return stored_value
+        raise KeyError("Key not found")
```
oksana-feshchenko, line 32: if you had collision before, your key, value would not be  under this index, how you get your item?

PR #263, opened 2023-02-27
Solution

yura-hudzovskyi:


app/main.py
```diff
@@ -1,2 +1,114 @@
...
+
+    def pop(self) -> tuple:
+        for i, node in enumerate(self.table):
+            if node is not None:
+                self.table[i] = node.next
+                self.size -= 1
+                return node.key, node.value
+        raise KeyError("Dictionary is empty")
+
+    def update(self, other: dict) -> None:
```
amber-marichi, line 103: why dont you add make this method to update also from another instance of your class as it does from default python dict?

amber-marichi (approved):
overall seems good but im not sure about chaining implementation. Seems a bit complicated. Maybe you should check update method since i guess it can update from another instance of your Dictionary yet. unless of course you are planning to implement items method as well ;)

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+                index = self.hash(node.key)
+                if new_table[index] is None:
+                    new_table[index] = Node(node.key, node.value)
+                else:
+                    current = new_table[index]
+                    while current.next:
+                        current = current.next
+                    current.next = Node(node.key, node.value)
+                node = node.next
+        self.table = new_table
```
YuriiKindrat, line 68: u can write old table to a variable and than iterate through it using method setitem to simplify ur code

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+        except KeyError:
+            return default
+
+    def pop(self) -> tuple:
+        for i, node in enumerate(self.table):
+            if node is not None:
+                self.table[i] = node.next
+                self.size -= 1
+                return node.key, node.value
+        raise KeyError("Dictionary is empty")
```
YuriiKindrat, line 101: what about value that we want to pop? also what about default value and KeyError?
image

app/main.py
YuriiKindrat: image
@amber-marichi is right, make ur method work with ur class and iterable

YuriiKindrat (changes requested):

app/main.py
yura-hudzovskyi: Unfortunately, I have no idea how to do it

app/main.py
yura-hudzovskyi: I'll think about it, but for now I've removed this method

wQuelS (approved):

PR #266, opened 2023-02-27
Solution

oksana-feshchenko:


ZabFTFT (approved):

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+            self.hash_table[index] = None
+            self.length -= 1
+
+    def get(self, key: Hashable) -> Any:
+        index = self.calculate_index(key)
+        current_element = self.hash_table[index]
+        if current_element:
+            return current_element.value
+
+    def pop(self, key: Hashable) -> Any:
```
YuriiKindrat, line 90: what about default value for pop ?

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+        index = self.calculate_index(key)
+        current_element = self.hash_table[index]
+        if current_element is None:
+            raise KeyError("No element with this key")
+        value_to_return = current_element.value
+        self.hash_table[index] = None
+        self.length -= 1
+        return value_to_return
+
+    def update(self, other: Dictionary) -> None:
```
YuriiKindrat, line 100: image

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+        current_element = self.hash_table[index]
+        if current_element:
+            self.hash_table[index] = None
+            self.length -= 1
+
+    def get(self, key: Hashable) -> Any:
+        index = self.calculate_index(key)
+        current_element = self.hash_table[index]
+        if current_element:
+            return current_element.value
```
Serhii-Leonenko, line 88: what about the value for this method if there is no such key in the dictionary?

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+            self.hash_table[index] = None
+            self.length -= 1
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        index = self.calculate_index(key)
+        current_element = self.hash_table[index]
+        if current_element:
+            return current_element.value
+        if default:
+            return default
```
wQuelS, line 90: image

app/main.py
```diff
@@ -1,2 +1,116 @@
+from __future__ import annotations
```
wQuelS, line 1: do you really need it ?

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+from dataclasses import dataclass
+from typing import Any, Hashable, Iterable
+import math
```
wQuelS, line 4: I bet you don't need the whole math module, do you? You can import only `floor` if you really need it..

wQuelS (changes requested):

YuriiKindrat (approved):

PR #270, opened 2023-02-27
Solution

ostapT:


app/main.py
```diff
@@ -1,2 +1,50 @@
+from typing import Hashable, Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 0.6
```
dogmoony, line 7: load factor doesn`t equal 0.6

app/main.py
```diff
@@ -1,2 +1,50 @@
...
+        self.hash_table = [[]] * self.capacity
```
dogmoony, line 8: where is type annotation of hash_table?

dogmoony (changes requested):

wQuelS (approved):

PR #271, opened 2023-02-27
Solution

ada-krav:


destr34 (approved):

app/main.py
```diff
@@ -1,2 +1,53 @@
+from math import ceil
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size = 0
+        self.capacity = 11
```
YuriiKindrat, line 8: why self.capacity = 11 if it should be equal to 8 by default ?

YuriiKindrat (changes requested):

app/main.py
ada-krav: There is nothing in the README about the required size. I know it`s 8 by default, but I like number 11 better, and as it`s my implementation, why not? I added used ceil for load factor, so 2/3 will be marginally achieved.

Serhii-Leonenko: You should implement a custom dictionary but with the same logic as the built-in python dictionary

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        self.hash_table = [None] * self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size >= ceil(self.capacity * 2 / 3):
+            self._resize()
+
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
```
Serhii-Leonenko, line 16: you can create an extra method for this

Serhii-Leonenko (changes requested):

PR #274, opened 2023-02-27
Solution

TeliksPG:


MikhailLyvak (approved):
Да ти шо, пєсня.

usrofgh (approved):

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Any, default: Any = None) -> int | None:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
```
wQuelS, line 67: image

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+
+    def update(self, other: dict) -> None:
+        for key, value in other.items():
+            self[key] = value
```
wQuelS, line 71: image

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+
+    def clear(self) -> None:
+        self._length = 0
+        self.table = [None] * self.capacity
```
wQuelS, line 75: imagine having additional method for `table` creation instead of writing this same logic 3 times..

wQuelS (changes requested):

app/main.py
```diff
@@ -58,21 +58,8 @@ def get(self, key: Any, default: Any = None) -> int | None:
...
-            value = self[key]
-            del self[key]
-            return value
-        except KeyError:
-            return default
-
-    def update(self, other: dict) -> None:
-        for key, value in other.items():
-            self[key] = value
-
```
wQuelS: why do you give up on those methods? Your code was nice, I really loved it, you just had few mistakes.
Isn't it interesting to implement every method correctly? Your were so close...

P.S. Since these methods are not obligatory you can press re-review after my reject and i will approve it. But maybe try a bit harder on this? I know you can easily come up with right solution.

wQuelS (changes requested):

app/main.py
ProdyRodion: Thank you so much for this answer, but if to be honest I don't have a resources for that on this moment.
I will return for this one a bit later, but i really appreciate your feedback!

YuriiKindrat (approved):

PR #289, opened 2023-02-28
Solution

Litvinenko23:


app/main.py
```diff
@@ -1,2 +1,75 @@
...
+            self.size += 1
+        else:
+            if self.hash_table[index].key == key:
+                self.hash_table[index] = Node(key, value)
+            else:
+                while (
+                    self.hash_table[index] is not None
+                    and self.hash_table[index].key != key
+                ):
+                    index = (index + 1) % self.capacity
```
YuriiKindrat, line 41: u are repeating this part of code twice in setitem\getitem methods, maybe it is better to move this part to a calc index method ?

YuriiKindrat (changes requested):

wQuelS (approved):

PR #293, opened 2023-02-28
Solution

RomanRega:


app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.key = key
+        self.value = value
+
+    def __repr__(self) -> str:
+        return f"{self.key}: {self.value}"
+
+
 class Dictionary:
-    pass
+    NEW_LEVEL = 2 / 3
```
ZhAlexR, line 14: I would choose more descriptive name for this constant, something like "LOAD_FACTOR_THRESHOLD", for example.

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+        self.length = 0
+        for node in old_hash_table:
+            if node is not None:
+                self[node.key] = node.value
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = hash(key) % self.capacity
+        while self.hash_table[index] is not None:
```
ZhAlexR, line 32: You are using this algorithm twice inside your class. 
May be it would be better to create separate method and just use it instead repeating whole code?

ZhAlexR (approved):
Generally it looks good

PR #294, opened 2023-02-28
Solution

Zenionoi:


app/main.py
```diff
@@ -1,2 +1,117 @@
...
+    d = Dictionary()
+    d[15] = "A"
+    d[3] = "B"
+    d[5] = "C"
+    d[13] = "D"
+    d[9] = "E"
+    d[14] = "K"
+    d[30] = "F"
+    print(d)
+    print(d[30])
```
YuriiKindrat, line 117: remove unnecessary code

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+    def pop(self, key: Hashable) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return None
+
+    def update(self, other: dict | list) -> None:
+        for key, value in other.items():
```
YuriiKindrat, line 93: are u sure about list.items method ?

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+        except KeyError:
+            return None
+
+    def pop(self, key: Hashable) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return None
```
YuriiKindrat, line 90: what about default value for this methods ?

app/main.py
YuriiKindrat: also it should work with ur custom dict, it is also didn't have an items method

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
```
YuriiKindrat, line 94: what about KeyError in case if default is not specified ? 
image

YuriiKindrat (changes requested):

PR #300, opened 2023-03-07
Solution

StreetNik:


app/main.py
```diff
@@ -1,2 +1,62 @@
+import random
+
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    capacity = 8
```
ArsenPidhoretskyi, line 7: Move it to instance atribute

app/main.py
```diff
@@ -1,2 +1,62 @@
...
+                self.table[index] = [key, key_hash, value]
+                self.size += 1
+            else:
+                for index, el in enumerate(self.table):
+                    if el is not None and el[0] == key and el[1] == key_hash:
+                        self.table[index][2] = value
+                        break
+
+    def __getitem__(self, key: Hashable) -> Any:
+        for i in self.table:
```
ArsenPidhoretskyi, line 56: use hash for getting item
search must be for better key as const(1), for now you always do len(n)

ArsenPidhoretskyi (changes requested):

app/point.py
```diff
@@ -17,7 +17,7 @@ def __hash__(self) -> int:
         # which is NOT a best practice, but you will be able to predict
         # a hash value by coordinates of the point and its index
         # in the hashtable as well
-        return hash((self.x, self.y))
+        return hash((self.x + self.y))
```
ArsenPidhoretskyi, line 20: Also this must be as not plus

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.table = self.table_update()
+        self.size = 0
+        self.trashold = int(self.capacity * (2 / 3)) + 1
+
+    def table_update(self) -> list:
```
Grayder0152, line 13: it is incorrect name for function, better to use `get_empty_table`

Grayder0152 (changes requested):

PR #302, opened 2023-03-13
Solution v_01

bohdan-yatsyna:


app/main.py
```diff
@@ -1,2 +1,77 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+
+    load_factor = 2 / 3
```
vasylhnatiuk, line 6: data like this better write in Costant - `LOAD_FACTOR = 2 / 3`

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+            if node:
+                key, value, hash_value = node
+                self.__setitem__(key, value)
+
+    def __setitem__(self, key: int | float | str | bool, value: Any) -> None:
+        if self.length > int(self.capacity * self.load_factor):
+            self.resize()
+
+        index = hash(key) % self.capacity
+        while self.hash_table[index]:
```
vasylhnatiuk, line 28: this code repeat this cycle in each function, try to write separate function. do this like `def find_index`

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+        hash_table_to_resize = self.hash_table
+        self.capacity *= 2
+        self.hash_table: list = [None] * self.capacity
+        self.length = 0
+        for node in hash_table_to_resize:
+            if node:
+                key, value, hash_value = node
+                self.__setitem__(key, value)
+
+    def __setitem__(self, key: int | float | str | bool, value: Any) -> None:
```
vasylhnatiuk, line 23: write `key: Hashable ` instead of `int | float | str | bool`

vasylhnatiuk (changes requested):

app/main.py
bohdan-yatsyna: Thank you for all the remarks. Updated already.

ArsenPidhoretskyi (approved):
Well done!

PR #305, opened 2023-03-17
Add solution

ZaikaDaria:


app/main.py
```diff
@@ -1,2 +1,46 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, length: int = 0, capacity: int = 8) -> None:
+        self.length = length
+        self.capacity = capacity
+        self.load_factor = int(self.capacity * 2 / 3)
+        self.hash_table: list = [[]] * self.capacity
```
Grayder0152, line 9: it is incorrect way , you create a copy of one list, use list comp.

image

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,2 +1,46 @@
...
+        self.capacity *= 2
+        self.load_factor = int(self.capacity * (2 / 3))
+        prev_hash_table = self.hash_table
+        self.hash_table: list = [None] * self.capacity
+        for item in prev_hash_table:
+            if item:
+                key, _, value = item
+                self.__setitem__(key, value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
katevasilenko, line 22: Key is `Hashable`

app/main.py
```diff
@@ -1,2 +1,46 @@
...
+        if self.length > self.load_factor:
```
katevasilenko, line 23: And what if length is equal to load_factor and you write new element? I believe it’s better to check equality here
```suggestion
        if self.length == self.load_factor:
```

app/main.py
```diff
@@ -1,2 +1,46 @@
...
+            self.resize()
+        index = hash(key) % self.capacity
+        while True:
+            if not self.hash_table[index]:
+                self.hash_table[index] = [key, hash(key), value]
```
katevasilenko, line 28: It is just my feeling but I would prefer to do not separate key and value so it’s better to put hash to the start of end of the list.

app/main.py
```diff
@@ -1,2 +1,46 @@
...
+                self.length += 1
+                break
+            if self.hash_table[index][0] == key:
+                self.hash_table[index][2] = value
+                break
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, item: Any) -> Any:
```
katevasilenko, line 36:
```suggestion
    def __getitem__(self, key: Hashable) -> Any:
```

katevasilenko (changes requested):

app/main.py
ZaikaDaria: Did understand. Please clarify, thank you.

Grayder0152: `self.hash_table[index] = [key, value, hash(key)]`

app/main.py
Grayder0152: pls resolve comment

app/main.py
Grayder0152: pls resolve comment

Grayder0152 (changes requested):

app/main.py
ZaikaDaria: It doesn't work like this. Can I left like this?
self.hash_table[index] = [key, hash(key), value]
Thanks

Знімок екрана 2023-03-22 о 17 19 42

vasylhnatiuk (changes requested):
you didn't commit and push your changes

ZaikaDaria:
> 

Solution doesn't work any longer :
image

vasylhnatiuk (changes requested):
if your code is not working please write your problem in chat, or ask q&a and we will help you

Vasyl-Smutok (approved):
Good Job 👍

PR #310, opened 2023-03-27
Solution

Andrew13X:


app/main.py
```diff
@@ -1,2 +1,50 @@
...
+        self.hash_table: list = [[] for _ in range(self.capacity)]
+        self.cells_occupied: int = 0
+
+    def __getitem__(
+            self,
+            key: Any
+    ) -> Any:
+        for cell in self.hash_table:
+            if cell and cell[0] == key:
+                return cell[2]
```
ArsenPidhoretskyi, line 16: Is really dict works by iterating and check for key only???
Why we need hash?

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #311, opened 2023-03-29
Solution

sasha-semenenko:


app/main.py
```diff
@@ -1,2 +1,46 @@
...
+
 class Dictionary:
-    pass
+
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.current_table: list = [[] for _ in range(self.capacity)]
+        self.taken_columns: int = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
olena-hul, line 11: key is Hashable

olena-hul (changes requested):

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+            self.resize()
+
+    def __getitem__(self, key: Hashable) -> Any:
+        current_hash = hash(key)
+        index_ = hash(key) % self.capacity
+        while self.current_table[index_]:
+            if current_hash == self.current_table[index_][2] \
+                    and key == self.current_table[index_][0]:
+                return self.current_table[index_][1]
+            else:
```
ArsenPidhoretskyi, line 33: Remove else:
If you do return inside if else: not needed

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+                index_ = (index_ + 1) % self.capacity
```
ArsenPidhoretskyi, line 34:
```suggestion
            index_ = (index_ + 1) % self.capacity
```

ArsenPidhoretskyi (changes requested):

PR #312, opened 2023-03-30
solution

viktory-koroliova:


ArsenPidhoretskyi (changes requested):
Also, wrong type annotations:
1.  any is boolean function
2. You can use from typing import Any
3. key is hashable not Any

PR #316, opened 2023-04-09
'Solution'

MilManIT:


app/main.py
```diff
@@ -1,2 +1,94 @@
-class Dictionary:
-    pass
+from typing import Any
+
+
+class Dictionary(object):
```
Bohdan-Salamakha, line 4: u don't need to inherit Dictionary from object
it's not a Python2, all classes are already inherited from object by default

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+                yield item
+
+    def __iter__(self) -> iter:
+        for key_var in self.__iterate_kv():
+            yield key_var[0]
+
+    def keys(self) -> iter:
+        return self.__iter__()
+
+    def values(self) -> iter:
```
Bohdan-Salamakha, line 63: return type is typing.Generator

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+        storage_idx = hash(key) % self.size
+        for item in self.storage[storage_idx]:
+            if item[0] == key:
+                return True
+        return False
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __iterate_kv(self) -> iter:
```
Bohdan-Salamakha, line 49: return type is typing.Generator

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+        for sub_lst in self.storage:
+            if not sub_lst:
+                continue
+            for item in sub_lst:
+                yield item
+
+    def __iter__(self) -> iter:
```
Bohdan-Salamakha, line 56: return type is typing.Generator

Bohdan-Salamakha (changes requested):
v1

PR #320, opened 2023-04-13
Solution +

MaksNochvai:


app/main.py
```diff
@@ -1,2 +1,82 @@
...
+    def _get_index(self, key: Hashable) -> Any:
+        index = hash(key) % self.capacity
+        while True:
+            if (self.hash_table[index] is None
+                    or self.hash_table[index][0] == key):
+                return index
+            index = (index + 1) % self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._get_index(key)
```
ArsenPidhoretskyi, line 29: It already returns the index which we need, as for me. Maybe inside we do not need while loop?

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+        else:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> object:
+        if key in self:
+            value = self[key]
+            del self[key]
+            return value
+        else:
+            return default
```
ArsenPidhoretskyi, line 75: Can you check if you do pop is it must return None as default
example:
d = {}
d.pop(1) # Must raise an error
d.pop(1, None) will work okay

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #321, opened 2023-04-14
Solution

leetwinoff:


app/main.py
```diff
@@ -1,2 +1,46 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 0.6
+        self.size = 0
+        self.hashtable = [None] * self.capacity
+
+    def __setitem__(self, key: int, value: Any) -> Any:
```
vasylhnatiuk, line 17:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> Any:
```

app/main.py
```diff
@@ -1,2 +1,46 @@
+from typing import Any
```
vasylhnatiuk, line 1:
```suggestion
from typing import Any, Hashable
```

app/main.py
```diff
@@ -1,2 +1,46 @@
...
+
+    def resize(self) -> None:
+        current_cell = [cell for cell in self.hashtable if cell is not None]
+        self.capacity *= 2
+        self.hashtable = [None] * self.capacity
+        self.size = 0
+        for cell in current_cell:
+            self.__setitem__(cell.key, cell.value)
+
+    def __getitem__(self, key: Any) -> Any:
```
vasylhnatiuk, line 37:
```suggestion
    def __getitem__(self, key: Hashable) -> Any:
```

app/main.py
vasylhnatiuk: key can be not only int

vasylhnatiuk (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #324, opened 2023-04-19
Solution

goldenuni:


app/main.py
```diff
@@ -1,2 +1,48 @@
...
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 5
+        self.filled_cells = 0
+        self.hash_table = [None] * self.capacity
+
+    def __setitem__(self,
+                    key: int | float | str | bool | tuple,
+                    value: Any) -> None:
+        index = hash(key) % self.capacity
```
olena-hul, line 14: You should call resize() when size is equal to threshhold (threshhold is calculated as int(self.capacity * 2 / 3))

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
+    def resize(self) -> None:
+        hash_table_ = self.hash_table
+        self.capacity = self.capacity * 2
+        self.hash_table = [None] * self.capacity
+        self.load_factor = int(self.capacity * (2 / 3))
+        self.filled_cells = 0
+        for cell in hash_table_:
+            if cell is not None:
+                self[cell[0]] = cell[2]
```
olena-hul, line 45: Use `__set_item__` here

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        index = hash(key) % self.capacity
+        if self.hash_table[index] is None:
+            self.hash_table[index] = []
+        for cell in self.hash_table[index]:
+            if cell[0] == key:
+                cell[1] = hash(key)
+                cell[2] = value
+                return
+        self.hash_table[index].append([key, hash(key), value])
+        self.filled_cells += 1
```
olena-hul, line 23: Are you sure the collisions will be handled properly? The index should be incremented when the collission occurrs. Here the index remains unchanged

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #327, opened 2023-04-23
solution

nickkozlov90:


app/main.py
```diff
@@ -1,2 +1,52 @@
...
+                        new_hash_table[index][2] = value
+                        break
+                    index = (index + 1) % len(new_hash_table)
+        self.hash_table = new_hash_table
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        hash_key = hash(key)
+        if self.length > len(self.hash_table) * 2 / 3:
+            self.resize()
+        index = hash_key % len(self.hash_table)
```
olena-hul, line 29: index should be calculated like: `hash_key % capacity`
You can store capacity separately in different instance attribute)

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+                break
+            index = (index + 1) % len(self.hash_table)
+
+    def __getitem__(self, key: Hashable) -> None:
+        hash_key = hash(key)
+        index = hash_key % len(self.hash_table)
+        while True:
+            if self.hash_table[index] is None:
+                raise KeyError
+            if key == self.hash_table[index][0]:
```
olena-hul, line 47: We should check hash_key too before checking the key

olena-hul (changes requested):

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        self.hash_table: list = [None] * 8
+        self.capacity = 8
+        self.is_resizing = False
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        self.length = 0
+        temp_table = [item for item in self.hash_table if item is not None]
+        self.hash_table = [None] * self.capacity
+        self.is_resizing = True
```
ArsenPidhoretskyi, line 16: Why you need this?
As for me it just debug code

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        for item in temp_table:
+            key, hash_key, value = item
+            self.__setitem__(key, value)
+        self.is_resizing = False
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.length > self.capacity * 2 / 3:
+            if not self.is_resizing:
```
ArsenPidhoretskyi, line 24: This check is redundant
While resizing it must never happens

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: Hashable) -> None:
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+        while True:
+            if self.hash_table[index] is None:
+                raise KeyError
+            if hash_key == self.hash_table[index][1]:
+                if key == self.hash_table[index][0]:
```
ArsenPidhoretskyi, line 46: Use and statement

ArsenPidhoretskyi (changes requested):

PR #329, opened 2023-05-01
'Solution'

iishchenko:


Vasyl-Smutok (changes requested):
you implemented the logic wrong, what if your dictionary has 100 items or more?

iishchenko:
I changed logic

app/main.py
```diff
@@ -1,2 +1,162 @@
...
+            if item is not None:
+                self[item[0]] = item[2]
+
+    def _get_key_position(self, key: Any) -> int:
+        key_hash = hash(key) % self.capacity
+        position = key_hash
+        while \
+                (self.container[position] is not None) and \
+                (self.container[position][1] > position):
+            position += 1
```
Grayder0152, line 25: use `()` instaed of `\` in all place

app/main.py
```diff
@@ -1,2 +1,162 @@
...
+    def resize(self, size: int) -> None:
+        container = self.container
+        self.capacity = size
+        self.container = [None] * size
+        self.len = 0
+        for item in container:
+            if item is not None:
+                self[item[0]] = item[2]
+
+    def _get_key_position(self, key: Any) -> int:
```
Grayder0152, line 19: Pls try to reduce count of loops in this method as i find it so difficult

app/main.py
```diff
@@ -1,2 +1,162 @@
...
+        new_item = (key, key_hash, value)
+        position = self._get_key_position(key)
+        if \
+                (self.container[position] is not None) and \
+                (self.container[position][0] != key):
+            self._move_down(position)
+        elif self.container[position] is None:
+            self.len += 1
+        self.container[position] = new_item
+        if self.len > self.capacity * 2 / 3:
```
Grayder0152, line 109: add `self.capacity * 2 / 3` value to class argument

app/main.py
```diff
@@ -1,2 +1,162 @@
...
+            self.resize(self.capacity * 2)
+
+    def __getitem__(self, key: Any) -> Any:
+        position = self._get_key_position(key)
+        if \
+                (self.container[position] is not None) and \
+                (self.container[position][0] == key):
+            return self.container[position][2]
+        else:
+            raise KeyError(key)
```
Grayder0152, line 119:
```suggestion
         raise KeyError(key)
```

app/main.py
```diff
@@ -1,2 +1,162 @@
...
+
+    def get(self, key: Any, default: Any = None) -> Any:
+        position = self._get_key_position(key)
+        if \
+                (self.container[position] is not None) and \
+                (self.container[position][0] == key):
+            return self.container[position][2]
+        else:
+            return default
```
Grayder0152, line 144: so common code as in `__getitem__`

app/main.py
```diff
@@ -1,2 +1,162 @@
...
+    def pop(self, key: Any) -> Any:
+        position = self._get_key_position(key)
+        if \
+                (self.container[position] is not None) and \
+                (self.container[position][0] == key):
+            value = self.container[position][2]
+            self._move_up(position)
+            return value
+        else:
+            raise KeyError(key)
```
Grayder0152, line 155:
```suggestion
        raise KeyError(key)
```

app/main.py
```diff
@@ -1,2 +1,162 @@
...
+            return default
+
+    def pop(self, key: Any) -> Any:
+        position = self._get_key_position(key)
+        if \
+                (self.container[position] is not None) and \
+                (self.container[position][0] == key):
+            value = self.container[position][2]
+            self._move_up(position)
+            return value
```
Grayder0152, line 153: use `self[key]` to avoid code dublication

iishchenko:
I changed a lot

app/main.py
```diff
@@ -1,2 +1,153 @@
...
+    def __init__(self) -> None:
+        self.capacity = 4
+        self.overflow_limit = self.capacity * 2 // 3
+        self.len = 0
+        self.container = [None] * self.capacity
+
+    def resize(self, size: int) -> None:
+        container = self.container
+        self.capacity = size
+        self.overflow_limit = self.capacity * 2 // 3
```
ArsenPidhoretskyi, line 14: * 2 // 3 is some magic number. Create some function and use this magic number only in this function

app/main.py
```diff
@@ -1,2 +1,153 @@
...
+        self.container = [None] * size
+        self.len = 0
+        for item in container:
+            if item is not None:
+                self[item[0]] = item[2]
+
+    def _get_key_position(self, key: Any) -> int:
```
ArsenPidhoretskyi, line 21: key is Hashable

app/main.py
```diff
@@ -1,2 +1,153 @@
...
+        key_hash = hash(key) % self.capacity
+        position = key_hash
+
+        def skip_looped_shift(container_row: tuple[Any, int, Any]) -> bool:
```
ArsenPidhoretskyi, line 25:
```suggestion
        def skip_looped_shift(container_row: tuple[Hashable, int, Any]) -> bool:
```

ArsenPidhoretskyi: Fix in other places
You can use from typing import Hashable

app/main.py
```diff
@@ -1,2 +1,153 @@
...
+                    (self.container[position] is not None) and \
+                    checker(self.container[position]):
+                position += 1
+            if position >= self.capacity:
+                position = 0
+                while \
+                        (self.container[position] is not None) and \
+                        checker(self.container[position]):
+                    position += 1
+        return position
```
ArsenPidhoretskyi, line 51: This code is so hard....
Just use only ONE while loop
In this loop check all conditions and increment current index

ArsenPidhoretskyi: All others code also the same(move_up and move_down)
And so hard.... Please simplify it and remove duplicated code
Also no need to use `\` instead use brackets

ArsenPidhoretskyi: Remove this code for move_down and move_up as for me it makes code more harder(Or simplify it)

ArsenPidhoretskyi (changes requested):

iishchenko:
new solution

iishchenko:
new solution

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        self.value = value
+        self.next: Optional[Node] = None
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+            initial_capacity: int = 10,
+            load_factor: float = 0.75
```
vsmutok, line 15: this is not correct data, it should be like this

```python
self._capacity = 8
self._load_factor = round(self.capacity * 2 / 3)
```

vsmutok (changes requested):

iishchenko:
New version

ArsenPidhoretskyi (approved):
Well done!

PR #333, opened 2023-05-10
Solution

V-Shkrobatskyi:


app/main.py
```diff
@@ -1,2 +1,83 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, *args) -> None:
+        self.nodelist = []
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.hash_table = [[i] for i in range(self.capacity)]
```
ArsenPidhoretskyi, line 9: Why you need i inside list???

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+        self.reinit(args)
+
+    def reinit(self, args: list) -> None:
+        while len(args) > self.capacity * self.load_factor:
+            self.capacity *= 2
+        self.hash_table = [[i] for i in range(self.capacity)]
+        for key_value in args:
+            key = key_value[0]
+            value = key_value[1]
```
ArsenPidhoretskyi, line 18:
```suggestion
            key, value = key_value
```

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+            for i in range(self.capacity):
+                correction = 0
+                if current_node + i >= self.capacity:
+                    correction = self.capacity - 1
+                hash_node = self.hash_table[current_node + i - correction]
+                if len(hash_node) == 1:
+                    hash_node.append(
+                        [hash(key), key, value]
+                    )
+                    break
```
ArsenPidhoretskyi, line 29: Use while loop because you have hash of current_node
Dict works with hash not as loop

app/main.py
```diff
@@ -1,2 +1,83 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, *args) -> None:
+        self.nodelist = []
```
ArsenPidhoretskyi, line 6: This property is not used

ArsenPidhoretskyi (changes requested):
Please use while loop because dict works with hash, not with for loop

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+            hash_node = self.hash_table[current_node]
+            while len(hash_node) != 0:
+                node_step += 1
+                if current_node + node_step >= self.capacity:
+                    node_step -= self.capacity
+                hash_node = self.hash_table[current_node + node_step]
+            hash_node.append([hash(key), key, value])
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.__len__() + 1 > self.capacity * self.load_factor:
```
ArsenPidhoretskyi, line 28: You can use len(self) instead of `__len__`

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+            reinit_args = []
+            for i in range(self.capacity):
```
ArsenPidhoretskyi, line 30: Why you can not iterate through node?
for node in self.hash_table

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+        hash_node = self.hash_table[current_node]
+        while len(hash_node) != 0:
+            if hash_node[0][1] == key:
+                hash_node[0][2] = value
+                break
+            node_step += 1
+            if current_node + node_step >= self.capacity:
+                node_step -= self.capacity
+            hash_node = self.hash_table[current_node + node_step]
+        hash_node.append([hash(key), key, value])
```
ArsenPidhoretskyi, line 47: Logic with node step so hard....
```suggestion
        index = hash(key) % self.capacity
        hash_node = self.hash_table[index]
        while len(hash_node) != 0:
            if hash_node[0][1] == key:
                hash_node[0][2] = value
                break
            index = (index + 1) % self.capacity
            hash_node = self.hash_table[index]
        hash_node.append([hash(key), key, value])
```

ArsenPidhoretskyi: Make it more simple)

ArsenPidhoretskyi: The same in getitem

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+                node_step -= self.capacity
+            hash_node = self.hash_table[current_node + node_step]
+        raise KeyError
+
+    def __len__(self) -> int:
+        result = 0
+        for i in range(self.capacity):
+            if len(self.hash_table[i]) != 0:
+                result += 1
+        return result
```
ArsenPidhoretskyi, line 70:
```suggestion
        result = sum(bool(node) for node in self.hash_table)
        return result
```
You can use bool() if node len > 0 it will return true which equal to 1 otherwise it will return 0

ArsenPidhoretskyi (changes requested):

app/main.py
V-Shkrobatskyi: Much less code. Thanks

Vasyl-Smutok (approved):
Great 👍

PR #337, opened 2023-05-16
solution

neostyle88:


app/point.py
```diff
@@ -26,3 +27,12 @@ def x(self) -> float:
...
         return self._y
+
+
+d = Dictionary()
+d[Point(1, 2)] = 9
+d[Point(5, 2)] = 13
+print(d[Point(5, 2)])
+print(hash(d[Point(1, 2)]))
+print(hash(Point(4, 2)))
+print(len(d))
```
YukihiroSM, line 38: This testing should be removed from point.py file, because any time, the point is being imported - this code will be executed. If you want to test something - better to do it in main.py, using something like `if __name__ == "__main__":`

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        if self.size / self.capacity > self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: Any) -> Any:
+        bucket_index = self._get_bucket_index(key)
+        bucket = self.buckets[bucket_index]
+
+        for k, v in bucket:
+            if k == key:
+                return v
```
YukihiroSM, line 35: Could you, please, find more appropriate variable names. k and v is not really description of what they contain.

YukihiroSM (changes requested):

app/point.py
```diff
@@ -17,7 +17,7 @@ def __hash__(self) -> int:
         # which is NOT a best practice, but you will be able to predict
         # a hash value by coordinates of the point and its index
         # in the hashtable as well
-        return hash((self.x, self.y))
+        return hash((self.x + self.y))
```
olena-hul, line 20: Do not change this

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.buckets = [[] for _ in range(initial_capacity)]
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        bucket_index = self._get_bucket_index(key)
+        bucket = self.buckets[bucket_index]
+
+        for i, (k, _) in enumerate(bucket):
+            if k == key:
+                bucket[i] = (key, value)
+                return
```
olena-hul, line 21: Your code isn't handling collissions properly
You should regenerate the index when the collission occurs

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+
+        if self.size / self.capacity > self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: Any) -> Any:
+        bucket_index = self._get_bucket_index(key)
+        bucket = self.buckets[bucket_index]
+
+        for find_key, value in bucket:
+            if find_key == key:
```
olena-hul, line 34: We should check hash and key. Hash should be checked first

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        return hash(key) % self.capacity
+
+    def _resize(self) -> None:
+        self.capacity *= 2
+        new_buckets = [[] for _ in range(self.capacity)]
+
+        for bucket in self.buckets:
+            for key, value in bucket:
+                new_bucket_index = hash(key) % self.capacity
+                new_buckets[new_bucket_index].append((key, value))
```
olena-hul, line 52: Use set_item function here you implemented above

app/main.py
olena-hul: Also we need to store hash in buckets

EdAlekseiev (approved):
Good job

PR #339, opened 2023-05-17
'Solution'

SashaLee:


app/main.py
```diff
@@ -1,2 +1,45 @@
+from typing import Hashable, Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.load_factor = 2 / 3
+        self.capacity = 8
+        self.size = 0
+        self.hash_table = [None] * 8
```
EdAlekseiev, line 9: What does the mean `8`? I think, `self.hash_table = [None] * self.capacity` is more understandable

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #340, opened 2023-05-17
Solution

OlehOryshchuk:


app/main.py
```diff
@@ -1,2 +1,126 @@
...
+
 class Dictionary:
-    pass
+
+    def __init__(self) -> None:
+        self._length = 0
+        self._capacity = 8
+        self._hash_table = [None] * self._capacity
+
+    def _get_index_and_hash(self, key: Any) -> tuple:
```
EdAlekseiev, line 20:
```suggestion
    def _get_index_and_hash(self, key: Hashable) -> tuple:
```

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+    def __getitem__(self, key: Any) -> None:
+        index, _ = self._get_index_and_hash(key)
+        current_node = self._hash_table[index]
+        while current_node is not None:
+            if current_node.key == key:
+                return current_node.value
+            current_node = current_node.next_node
+        raise KeyError(f"There are no key {key}")
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
EdAlekseiev, line 45:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+            if node is not None:
+                return node.key, node.value
+        raise StopIteration
+
+    def clear(self) -> None:
+        self._length = 0
+        self._capacity = 8
+        self._hash_table = [None] * self._capacity
+
+    def get(self, key: Any) -> Any:
```
EdAlekseiev, line 107: It would be nice if `get` could take a default value as an argument.

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+        current.next_node = Node(key, value, keys_hash)
+        self._length += 1
+
+        if self._length >= self._capacity * 2 / 3:
+            self._resize()
+
+    def __len__(self) -> int:
+        return self._length
+
+    def __delitem__(self, key: Any) -> None:
```
EdAlekseiev, line 72:
```suggestion
    def __delitem__(self, key: Hashable) -> None:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+        for node in old_hash_table:
+            while node is not None:
+                self[node.key] = node.value
+                node = node.next_node
+
+    def __getitem__(self, key: Hashable) -> None:
+        index, _ = self._get_index_and_hash(key)
+        current_node = self._hash_table[index]
+        while current_node is not None:
+            if current_node.key == key:
```
olena-hul, line 40: We should check hash before the key too

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+
+            if current.next_node is None:
+                break
+            current = current.next_node
+
+        current.next_node = Node(key, value, keys_hash)
+        self._length += 1
+
+        if self._length >= self._capacity * 2 / 3:
+            self._resize()
```
olena-hul, line 67: This check should be at the beginning of the function

app/main.py
```diff
@@ -1,2 +1,126 @@
...
+    def _resize(self) -> None:
+        self._length = 0
+        self._capacity *= 2
+        old_hash_table = self._hash_table
+        self._hash_table = [None] * self._capacity
+
+        for node in old_hash_table:
+            while node is not None:
+                self[node.key] = node.value
+                node = node.next_node
```
olena-hul, line 34: You can use `__setitem__` here

olena-hul (changes requested):

app/main.py
OlehOryshchuk: Sorry @olena-hul i don't really understand . I'm already using `__setitem__` here `` self[node.key]` = node.value.` No?

app/main.py
EdAlekseiev: Still not fixed

app/main.py
EdAlekseiev: Still not fixed

app/main.py
EdAlekseiev: You can use `self[node.key]`, it is ok.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #343, opened 2023-05-24
solution

T-Vlad-96:


app/main.py
```diff
@@ -1,2 +1,115 @@
...
+            self,
+            key: Union[int, float, str, bool, tuple, Point],
+            value: Any
+    ) -> None:
+        hashed_key = hash(key)
+        index = hashed_key % self.capacity
+        while self.hash_table[index] is not None:
+            current_key, current_value = self.hash_table[index]
+            if current_key == key:
+                self.hash_table[index] = [current_key, value]
```
olena-hul, line 44: Store hash value too

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+            self.current_element += 1
+            if self.current_element > self.capacity - 1:
+                raise StopIteration
+        result = self.hash_table[self.current_element][0]
+        self.current_element += 1
+        return result
+
+    def __setitem__(
+            self,
+            key: Union[int, float, str, bool, tuple, Point],
```
olena-hul, line 36:
```suggestion
            key: Hashable,
```

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+
+    def __getitem__(
+            self,
+            item: Union[int, float, str, bool, tuple, Point]
+    ) -> any:
+        hashed_item = hash(item)
+        index = hashed_item % self.capacity
+        while self.hash_table[index] is not None:
+            key, value = self.hash_table[index]
+            if key == item:
```
olena-hul, line 61: We should check hash and key

app/main.py
olena-hul: Fix in all places

olena-hul (changes requested):

app/main.py
T-Vlad-96: when i store hash, my code doesn't pass the tests. I get "ValueError: too many values to unpack (expected 2)" in every single test

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+                key, value, hash_of_key = item
+                self.__setitem__(key, value)
+
+    def clear(self) -> None:
+        self.__init__()
+
+    def get(
+            self,
+            key: Hashable
+    ) -> Union[None, any]:
```
Bohdan-Salamakha, line 94: Any already contains None type
Any should be started with uppercase char

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #344, opened 2023-05-25
'Question'

Jhaivir:
I haven't finished the work yet. Need help. I do not understand why this method (adding_new_value) does not work correctly (it fills the entire hash list at once)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,2 +1,101 @@
+from typing import Union
+
+
 class Dictionary:
-    pass
+
+    def __init__(self) -> None:
+        self.hash_table: list = [
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None]
+        ]
+
+    def __setitem__(self, key: Union, value: Union) -> None:
+        index_hash = self.hashed(key)
+        if self.hash_table[index_hash][0] is key:
+            self.hash_table[index_hash][1] = value
+            return
+        if (
+            self.hash_table.count([None, None])
+            == int(len(self.hash_table) / 3 + 1)
+        ):
+            copy_old_table = self.hash_table.copy()
+            self.hash_table = [[None, None]] * len(self.hash_table) * 2
+            for cell in range(len(copy_old_table)):
+                if copy_old_table[cell] != [None, None]:
+                    new_cell = self.hashed(self.hash_table[cell][0])
+                    self.adding_new_value(
+                        new_cell,
+                        copy_old_table[cell][0],
+                        copy_old_table[cell][1]
+                    )
+        if self.hash_table[index_hash][0] is None:
+            self.hash_table[index_hash][0] = key
+            self.hash_table[index_hash][1] = value
+            return
+        self.adding_new_value(10, key, value)
+
+    def adding_new_value(self, index_hash_table, key, value) -> None:
+        for _ in range(len(self.hash_table)):
+            if not self.hash_table[index_hash_table][0]:
+                self.hash_table[index_hash_table][0] = key
+                self.hash_table[index_hash_table][1] = value
+                return
+            if index_hash_table == len(self.hash_table) - 1:
+                index_hash_table = 0
+            else:
+                index_hash_table += 1
+
+    def __getitem__(self, key: Union) -> Union:
+        if (
+            self.hash_table[self.hashed(key)][1] is None
+            and key is None
+        ):
+            raise KeyError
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        self.hash_table: list = [
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None],
+            [None, None]
+        ]
```
ArsenPidhoretskyi, line 16: Use loop

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def __setitem__(self, key: Union, value: Union) -> None:
+        index_hash = self.hashed(key)
+        if self.hash_table[index_hash][0] is key:
```
ArsenPidhoretskyi, line 20: `is` is not good to use in this vay use ==

app/main.py
ArsenPidhoretskyi: Also better to create list with None: [None, None, None, ....] instead of list of list with None

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+                        copy_old_table[cell][1]
+                    )
+        if self.hash_table[index_hash][0] is None:
+            self.hash_table[index_hash][0] = key
+            self.hash_table[index_hash][1] = value
+            return
+        self.adding_new_value(10, key, value)
+
+    def adding_new_value(self, index_hash_table, key, value) -> None:
+        for _ in range(len(self.hash_table)):
```
ArsenPidhoretskyi, line 44: It must works with while loop:
while (here is condition with check that cell is available to fill, if not available continue loop)
inside loop just do:
index_hash_table = (index_hash_table + 1) % len(self.has_table)

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        ]
+
+    def __setitem__(self, key: Union, value: Union) -> None:
+        index_hash = self.hashed(key)
+        if self.hash_table[index_hash][0] is key:
+            self.hash_table[index_hash][1] = value
+            return
+        if (
+            self.hash_table.count([None, None])
+            == int(len(self.hash_table) / 3 + 1)
```
ArsenPidhoretskyi, line 25: / 3 + 1
For what this magic numbers need?

ArsenPidhoretskyi (changes requested):

app/main.py
Jhaivir: Why isn't Slack working?

vasylhnatiuk (changes requested):
you pushed nothing

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+
+
+items = [(1, "one"), (2, "two"), (3, "tree"), (4, "four")]
+
+dictionary = Dictionary()
+for key, value in items:
+    dictionary[key] = value
+
+for key, value in items:
+    print(key, dictionary[key])
```
vsmutok, line 96: remove it or add  `if __name__ == '__main__':`

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+            values = []
+        self.length: int = 0
+        self.hash_table: list = [None] * 8
+        self.keys = keys
+        self.values = values
+        self.need_key: Union = None
+        for i in range(len(keys)):
+            self.__setitem__(keys[i], values[i])
+
+    def __setitem__(self, key_dict: Union, value_dict: Union) -> None:
```
vsmutok, line 22:
```suggestion
    def __setitem__(self, key_dict: Hashable, value_dict: Any) -> None:
```

vsmutok (changes requested):
also please use black for fix indentation

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+            self.hash_table = [None] * len(self.hash_table) * 2
+            for cell in copy_old_table:
+                if cell:
+                    self.need_key = cell[0]
+                    self.hash_table[self.__hash__()] = cell
+        self.need_key = key_dict
+        self.hash_table[self.__hash__()] = [key_dict, value_dict]
+        self.length = len(self.hash_table) - self.hash_table.count(None)
+
+    def __getitem__(self, key_dict: Union) -> Union:
```
vsmutok, line 40:
```suggestion
    def __getitem__(self, key_dict: Hashable) -> Any:
```

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍 

also try to use better commit names in the future

PR #348, opened 2023-05-26
Solution

vlamuha:


ArsenPidhoretskyi (approved):
Well done!

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.threshold = int(self.capacity * 2 / 3)
+        self.hash_table = [[] for _ in range(self.capacity)]
+        self.storage = 0
+
+    def __getitem__(self, key: (int, float, str, tuple, bool)) -> list:
```
Abnormaltype, line 11:
```suggestion
    def __getitem__(self, key: (int, float, str, tuple, bool)) -> list:
```

`Key` can be class.
You should put `hashable`

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+        while True:
+            try:
+                if self.hash_table[index][1] == key:
```
Abnormaltype, line 16: you should also compare the hash

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+        self.initial_capacity *= 2
+        self.hash_table = [None] * self.initial_capacity
+        self.size = 0
+        for node in nodes:
+            self.__setitem__(node[0], node[1])
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = hash(key) % self.initial_capacity
+        while self.hash_table[index] is not None:
+            if self.hash_table[index][0] == key:
```
ArsenPidhoretskyi, line 36: the hash must be also the same

ArsenPidhoretskyi: Save hash into hash_table

app/main.py
```diff
@@ -1,2 +1,47 @@
...
+            else:
+                index = (index + 1) % self.initial_capacity
+        else:
+            self.hash_table[index] = [key, value]
+            self.size += 1
+            if self.size > int(self.initial_capacity * self.load_factor):
+                self.resize()
+
+    def resize(self) -> None:
+        nodes = [x for x in self.hash_table if x is not None]
```
ArsenPidhoretskyi, line 26: YOu can use:
list(filter(lambda cell: cells it not None, self.hash_table))

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #352, opened 2023-06-08
py-dictionary-v1

muriakula:


app/main.py
```diff
@@ -1,2 +1,70 @@
...
+                    return
+        self.buckets[index].append(node)
+        self.size += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._calculate_index(key)
+        if not self.buckets[index]:
+            raise KeyError
+        for element in self.buckets[index]:
+            if key == element.key:
```
EdAlekseiev, line 38: We should check hash key too before checking the key

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+
+    def _rebuild(self) -> None:
+        self.capacity *= 2
+        temp_buckets = self.buckets
+        self.buckets = [[] for _ in range(self.capacity)]
+        for internal_list in temp_buckets:
+            if internal_list:
+                for element in internal_list:
+                    index = self._calculate_index(element)
+                    self.buckets[index].append(element)
```
olena-hul, line 61: Use `__setitem__` here

olena-hul (changes requested):

app/main.py
muriakula: I'm sorry, I don't understand.
How can I use __setitem__ in _rebuild if I already use _rebuild in __setitem__ ?

Bohdan-Salamakha (approved):

PR #354, opened 2023-06-11
Solution

yuro740:


app/main.py
```diff
@@ -1,2 +1,146 @@
...
+                break
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.load_factor >= self._load_factor_threshold:
+            self._resize_and_rehash()
+
+        for index, pair in self._probe(key):
+            if pair is DELETED:
+                continue
+            if pair is None or pair.key == key:
```
ArsenPidhoretskyi, line 59:
```suggestion
            if pair is None or pair.key == key:
```
Also compare by hash

ArsenPidhoretskyi: In getitem the same

ArsenPidhoretskyi (changes requested):

app/main.py
yuro740: No need to compare the hash.  Since the _probe function returns an element and an index that was calculated based on the hash (from key) and capacity (in function _index) .

app/main.py
```diff
@@ -1,2 +1,159 @@
...
+            if pair.key == key:
+                return pair.value
+        raise KeyError(key)
+
+    def __contains__(self, key: Hashable) -> bool:
+        try:
+            self[key]
+        except KeyError:
+            return False
+        else:
```
vsmutok, line 84: Redundant `else`

app/main.py
```diff
@@ -1,2 +1,159 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.load_factor >= self._load_factor_threshold:
+            self._resize_and_rehash()
+
+        """all the available slots in the hash table,
+        starting from the calculated by hash index"""
+        for index, pair in self._probe(key):
+            if pair is DELETED:
+                continue
+            # Don't need to compare hash.
```
vsmutok, line 61: remove this comment

vsmutok (changes requested):

app/main.py
```diff
@@ -1,2 +1,157 @@
...
+
+    def __iter__(self) -> Any:
+        yield from self.keys
+
+    def __delitem__(self, key: Hashable) -> None:
+        for index, pair in self._probe(key):
+            if pair is None:
+                raise KeyError(key)
+            if pair is DELETED:
+                continue
```
ArsenPidhoretskyi, line 47: Why if is deleted do continue?

ArsenPidhoretskyi (changes requested):
DELETED is not needed you can just set value to None. As for me, It makes logic complicated

app/main.py
yuro740: I can't set to none. In case:
When I have collision, and I put into next slot my pair, and after I delete previous. If I set to none, I can't find my pair in next slots.

Grayder0152 (changes requested):
Pls, create valid doc string

app/main.py
ArsenPidhoretskyi: But in this case better also raise exception
image

ArsenPidhoretskyi: But okay for now. The rest code is okay)

ArsenPidhoretskyi (approved):
Good job!

PR #357, opened 2023-06-18
custom Dictionary solution

sbilikepy:


app/main.py
```diff
@@ -1,2 +1,121 @@
...
+from typing import Any, Union
+
+
 class Dictionary:
-    pass
+    """
+    Dict(object) clone
+
+    some_dict = {"One": 1}          => {"Key": "value"}
+    var = Dictionary_obj("One", 1)  => [("One", 1)]
```
ArsenPidhoretskyi, line 11: Not true

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+    """
+
+    def __init__(
+            self,
+            key: Union[
+                int, float, complex, str, bool, tuple, frozenset,
+                bytes, None, Enum, date, datetime
+            ],
+            value: Any
```
ArsenPidhoretskyi, line 20: No need to accept params like key and value

app/main.py
ArsenPidhoretskyi: The original dicts does not work like this

ArsenPidhoretskyi (changes requested):
The task is not finished

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        try:
+            hash(key)
+        except TypeError:
+            raise TypeError(f"unhashable type: '{type(key).__name__}'")
+
+        for bucket in self.buckets_table:
+            if bucket and bucket[0] == key:
+                bucket[1] = value
+                return
```
ArsenPidhoretskyi, line 36: Loop works by hash_index not by forloop

app/main.py
```diff
@@ -1,2 +1,111 @@
...
+
+        new_k_v_pair = [key, value, hash(key)]
+        index = hash(key) % self.capacity
+        empty_bucket_index = next((i for i, x in enumerate(self.buckets_table)
+                                   if not x), None)
```
ArsenPidhoretskyi, line 41: Loop works by hash_index not by forloop

ArsenPidhoretskyi (changes requested):
Loop works by hash_index not by forloop
Fix in all places

sbilikepy:
Hello, I got rid of "for" where possible
in case where key already exists, I tried to change elements in hash table this way:

```
bucket = self.buckets_table[index]
if len(bucket) and hash(key) in bucket:
    bucket[1] = value
```

the consequence was that half of the tests failed 
so I changed `for `to `while`, but in fact the enumeration remained
help 😢

vsmutok (changes requested):
don't change tests

vsmutok:
https://github.com/mate-academy/py-dictionary/pull/357#issuecomment-1611690171

please duplicate this question in the Slack chat or go to QnA

app/main.py
```diff
@@ -1,2 +1,113 @@
...
+        self.old_buckets_table = copy.deepcopy(self.buckets_table)
+        self.buckets_table = [[] for bucket in range(self.capacity)]
+
+        for bucket in self.old_buckets_table:
+            if len(bucket):
+                index_by_new_hash = bucket[2] % self.capacity
+                if not len(self.buckets_table[index_by_new_hash]):
+                    self.buckets_table[index_by_new_hash] = bucket
+                else:
+                    self.buckets_table[self.buckets_table.index([])] = bucket
```
ArsenPidhoretskyi, line 25: You can just do:
self[key] = value

app/main.py
```diff
@@ -1,2 +1,113 @@
...
+        for bucket in self.buckets_table:
+            if len(bucket) and bucket[0] == key:
+                bucket.clear()
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        for bucket in self.buckets_table:
+            if len(bucket) and key == bucket[0]:
+                return bucket[1]
+
+        return default
```
ArsenPidhoretskyi, line 77: You can reuse getitem

app/main.py
```diff
@@ -1,2 +1,113 @@
...
+        for bucket in other_dict.buckets_table:
+            if len(bucket):
+                self.__setitem__(bucket[0], bucket[1])
+
+    def pop(self, key: Hashable, default: Any = None) -> Union[None, Any]:
+        value = None
+        for bucket in self.buckets_table:
+            if len(bucket) and bucket[0] == key:
+                value = bucket[1]
+                bucket.clear()
```
ArsenPidhoretskyi, line 89: You can reuse getitem + delitem

ArsenPidhoretskyi (changes requested):

app/main.py
sbilikepy: Hello! Please review this changes:
1. I removed for loops wherever possible
2. To implement the __getitem__ (and other) methods, I created a collision list, which is much smaller than the hash table itself

sbilikepy:
Hello! Please review this changes:
1. I removed for loops wherever possible
2. To implement the __getitem__ (and other) methods, I created a collision list, which is much smaller than the hash table itself

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+        if self.hash_table[index][0] == key:
+            self.hash_table[index][1] = value
+            return
+
+        """Check old collisions"""
+        for collision in self.collisions:
+            if collision[0] == key:
+                collision_item_index = collision[1]
+                self.hash_table[collision_item_index][1] = value
+                return
```
ArsenPidhoretskyi, line 51: This is a little strange. Just find the first empty cell

ArsenPidhoretskyi: The same in getitem

ArsenPidhoretskyi: And delitem

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+                self.hash_table[collision[1]].clear()
+                return
+        raise KeyError(key)
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        self.__getitem__(key)
+
+    def pop(self, key: Hashable, default: Any = None) -> Union[None, Any]:
+        value = self.__getitem__(key)
+        self.__delitem__(key)
```
ArsenPidhoretskyi, line 96: if default is provided exception must not raise

ArsenPidhoretskyi (changes requested):

app/main.py
sbilikepy: "The method def __setitem__ first tries to:

1. Find an empty bucket and fill it with the value (+return)
2. Find a bucket with the same key and overwrite its value (+return)
In this case, I handle the scenario where I try to overwrite the value of a bucket that had a collision during placement. To find the correct bucket, I access self.collisions to get its index and then change the value in self.hash_table using the index from the collisions list.

That's why I cannot search for an empty cell to write the value, as this function is **already** performed in the code blocks above, namely in the "Empty bucket" and "Reassign value" parts, which are **protected by return** statements as well.

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+
+        """Resize"""
+        self.capacity *= 2
+        self.old_hash_table = copy.deepcopy(self.hash_table)
+        self.hash_table = [[] for bucket in range(self.capacity)]
+        self.collisions.clear()
+
+        for bucket in self.old_hash_table:
+            if len(bucket):
+                self.__setitem__(key=bucket[0], value=bucket[1])
```
ArsenPidhoretskyi, line 26: Use:
self[bucket[0]] = bucket[1]

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        item = [key, value, hash(key)]
+
+        """Empty bucket"""
+        if not len(self.hash_table[index]):
+            self.hash_table[index] = item
+            self.resize()
+            return
+
+        """Reassign value"""
+        if self.hash_table[index][0] == key:
```
ArsenPidhoretskyi, line 42: Also you must check that hash the same

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+            self.hash_table[index][1] = value
+            return
+
+        """Check old collisions"""
+        for collision in self.collisions:
+            if collision[0] == key:
```
ArsenPidhoretskyi, line 48: Also you must check that hash the same

app/main.py
ArsenPidhoretskyi: But it makes code harder. You can avoid the second list for this task. But If you want you can keep it

app/main.py
ArsenPidhoretskyi: Not fixed
image

ArsenPidhoretskyi: image
And how it must works

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        self.__delitem__(key)
+
+        if default is not None:
+            return default
+
+        if value is not None:
+            return value
+
+        if default is not None:
+            return default
```
ArsenPidhoretskyi, line 105: Duplicate if statement

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+
+        raise KeyError
+
+    def update(self, other_dict: "Dictionary") -> None:
+        for bucket in other_dict.hash_table:
+            if len(bucket):
+                self.__setitem__(bucket[0], bucket[1])
```
ArsenPidhoretskyi, line 112: Use:
self[bucket[0]] = bucket[1]

ArsenPidhoretskyi (changes requested):

sbilikepy:
@ArsenPidhoretskyi 
thanks for all the feedback, I appreciate it!

vsmutok (approved):
Looks Good 👍

sbilikepy:
> Looks Good 👍

oh thank God its done 😄

PR #358, opened 2023-06-19
Solution

AntonShpakovych:


Kirontiko (approved):
GJ, but idk, is threshold != capacity * 2/3? I think the second attr in init is load_factor

AntonShpakovych:
> GJ, but idk, is threshold != capacity * 2/3? I think the second attr in init is load_factor

I just took naming from our task in algorithm topic

app/main.py
```diff
@@ -1,2 +1,68 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.threshold: float = 0.66
+        self.size: int = 0
+        self.hash_table: list = [None] * 8
```
LiudmylaKulinchenko, line 9: Annotations for the attributes are not really necessary here. Type hints are already clear from the assigned values. You can skip annotations in such cases to write less code

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+
+    def __getitem__(self, key: Any) -> Any:
```
LiudmylaKulinchenko, line 11: Are u sure key could have any type? Dictionaries in Python require their keys to be **hashable**

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+        index = self.__index_from_key(key)
+
+        while self.hash_table[index] is not None:
```
LiudmylaKulinchenko, line 14: You can skip `is not None` here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job!! 😎🔥🔥

PR #359, opened 2023-06-20
solution

oleksiikolii:


app/main.py
```diff
@@ -1,2 +1,48 @@
...
+                if self.hash_table[index][0] == key:
+                    self.hash_table[index] = (key, self.get_index(key), value)
+                    break
+                index = index + 1 if index < self.capacity - 1 else 0
+
+    def __getitem__(self, item: Any) -> Any:
+        index = self.get_index(item)
+        cnt = 0
+        while cnt < self.capacity:
+            if self.hash_table[index] is not None:
```
ant-komarov, line 29: You can use here just "if self.hash_table[index]" without "is not None"

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+                if self.hash_table[index][0] == item:
+                    return self.hash_table[index][2]
+            index = index + 1 if index < self.capacity - 1 else 0
+            cnt += 1
+        raise KeyError
+
+    def resize(self) -> None:
+        old_table = self.hash_table
+        self.hash_table = [None] * (self.capacity * 2)
+        self.capacity *= 2
```
ant-komarov, line 39: You can change these two lines by places, and you'll don't be need multiply by 2 twice

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        for item in old_table:
+            if item is not None:
```
ant-komarov, line 41: Also here, no need compare with None

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        if self.hash_table.count(None) < int(self.capacity * self.load_factor):
+            self.resize()
+        index = self.get_index(key)
+        while True:
+            if self.hash_table[index] is None:
+                self.hash_table[index] = (key, self.get_index(key), value)
+                break
+            else:
+                if self.hash_table[index][0] == key:
+                    self.hash_table[index] = (key, self.get_index(key), value)
```
ant-komarov, line 21: also here index instead hash

ant-komarov (changes requested):

ant-komarov (approved):

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
 class Dictionary:
-    pass
+
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 0.66
+        self.hash_table = [None] * self.capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
Oleg-s-s, line 11: Are you sure that key type can be `Any`?

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        index = self.get_index(key)
+        while True:
+            if not self.hash_table[index]:
+                self.hash_table[index] = (key, hash(key), value)
+                break
+            else:
+                if self.hash_table[index][0] == key:
+                    self.hash_table[index] = (key, hash(key), value)
+                    break
+                index = index + 1 if index < self.capacity - 1 else 0
```
Oleg-s-s, line 23: You can combine those part + optimize a bit

Oleg-s-s (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🍬

PR #360, opened 2023-06-20
Solution

Kirontiko:


app/main.py
```diff
@@ -1,2 +1,68 @@
...
+            if self.hash_table[index][0] == key:
+                self.hash_table[index] = None
+
+    def resize(self) -> None:
+        hash_table = self.hash_table
+        self.size = 0
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+        for item in hash_table:
+            if item is not None:
```
kovaliskoveronika, line 64: Maybe it will be better to not use 'if item is not None'. You can write 'if item' and will work correctly. The same with 'if item is None'  you can use 'if not item'

kovaliskoveronika (approved):

app/main.py
Kirontiko: oh, that's great notice. I will fix it

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+                    self.hash_table[index] = (key, hash(key), value)
+                    break
+                if index + 1 < self.capacity:
+                    index += 1
+                else:
+                    index = 0
+
+    def __getitem__(self, key: Hashable) -> Any:
+        if not isinstance(key, Hashable):
+            raise TypeError(f"unhashable type: '{type(key)}'")
```
LiudmylaKulinchenko, line 38: No need to make this check and raise an error. It will raise the same error in the next line of code, because in get_index_from_hash method you will try to get a hash from this value

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+    def resize(self) -> None:
+        hash_table = self.hash_table
+        self.size = 0
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+        for item in hash_table:
+            if item:
+                self.__setitem__(item[0], item[2])
+
+    def get_index_from_hash(self, key: Any) -> int:
```
LiudmylaKulinchenko, line 69: Key should be hashable 🤫 not any

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+                return self.hash_table[index][2]
+            index = (index + 1) % self.capacity
+        raise KeyError(f"key {key} is not found")
+
+    def clear(self) -> None:
+        self.hash_table = [None] * self.capacity
+
+    def __delitem__(self, key: Hashable) -> None:
+        if not isinstance(key, Hashable):
+            raise TypeError(f"unhashable type: '{type(key)}'")
```
LiudmylaKulinchenko, line 52: Here is the same about raising TypeError

LiudmylaKulinchenko (changes requested):

app/main.py
Kirontiko: oh, changed it everywhere but here(

app/main.py
Kirontiko: Thank you, I thought I have to raise it by myself. I'll fix it!

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        self.hash_table: list = [None] * self.capacity
+
+    def __len__(self) -> int:
+        print(self.capacity)
+        return self.size
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        threshhold = int(self.capacity * self.load_factor)
+        if self.capacity - threshhold > self.hash_table.count(None):
+            self.resize()
```
LiudmylaKulinchenko, line 18: You should make resize after you added a new value in a dict

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+
+        index = self.get_index_from_hash(key)
+        while True:
```
LiudmylaKulinchenko, line 21: Try to avoid so many conditions here
Consider to add some condition in while instead of using just while True, as you did in __getitem__ method

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,65 @@
...
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.size: int = 0
+        self.load_factor: float = 0.66
+        self.hash_table: list = [None] * self.capacity
+
+    def __len__(self) -> int:
+        print(self.capacity)
```
LiudmylaKulinchenko, line 12: Why do you need this print here?

LiudmylaKulinchenko (approved):
Well done! 🎉🎉

PR #361, opened 2023-06-20
'Solution'

ant-komarov:


app/main.py
```diff
@@ -1,2 +1,53 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length: int = 0
```
AntonShpakovych, line 6: Annotations for the attributes are not really necessary here. Type hints are already clear from the assigned values. You can skip annotations in such cases to write less code -  `LiudmylaKulinchenko`

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        self.capacity: int = 8
+        self.load_factor: float = 0.66
+        self.hash_table: list = [None] * self.capacity
+
+    def __add_to_hash_table(self, key_hash: int, key: Any, value: Any) -> None:
```
AntonShpakovych, line 11: change annotation for key in all function where u use Any - `Hashable`

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        _index = key_hash % self.capacity
+        while True:
+            if not self.hash_table[_index]:
+                self.hash_table[_index] = (key_hash, key, value)
+                self.length += 1
+                break
+            elif self.hash_table[_index][1] == key:
+                self.hash_table[_index] = (key_hash, key, value)
```
AntonShpakovych, line 19: I know it doesn't make a difference in this case, but let's imagine that some other person will use this dict. ((key, hash, value)

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+            if item:
+                self.__add_to_hash_table(*item)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.length == int(self.capacity * self.load_factor):
+            self.__resize_hash_table()
+        key_hash = hash(key)
+        self.__add_to_hash_table(key_hash, key, value)
+
+    def __getitem__(self, key: Any) -> Any:
```
AntonShpakovych, line 40: u can simplify this method

AntonShpakovych (changes requested):

AntonShpakovych (approved):
Good job

LiudmylaKulinchenko (approved):
Well done! 🎉

PR #362, opened 2023-06-20
solution

mariiahorbova:


maccarets (approved):

app/main.py
```diff
@@ -1,2 +1,76 @@
...
+                 load_factor: float = 0.75
+                 ) -> None:
+        self.initial_capacity = initial_capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.capacity = initial_capacity
+        self.table = [None] * self.capacity
+
+    def __setitem__(self,
+                    key: int | float | str | bool | tuple,
```
LiudmylaKulinchenko, line 16: There are more types that could be dict keys. Keys should be _**hashable** types_

LiudmylaKulinchenko: I'm hinting to `hashable` type annotation

app/main.py
```diff
@@ -1,2 +1,76 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self,
+                 initial_capacity: int = 8,
+                 load_factor: float = 0.75
```
LiudmylaKulinchenko, line 7: Is load factor 0,75?

LiudmylaKulinchenko: Also consider to make initial_capacity and load_factor class attributes

app/main.py
```diff
@@ -1,2 +1,76 @@
...
+        nodes.append([key, self._hash(key), value])
+        self.size += 1
+        self._check_resize()
+
+    def __getitem__(self,
+                    key: int | float | str | bool | tuple
+                    ) -> Any:
+        index = self._get_index(key)
+        nodes = self.table[index]
+        if nodes is not None:
```
LiudmylaKulinchenko, line 35: What does this condition here?

app/main.py
```diff
@@ -1,2 +1,76 @@
...
+        return self.size
+
+    def _get_index(self,
+                   key: int | float | str | bool | tuple
+                   ) -> int:
+        hash_value = hash(key)
+        return hash_value % self.capacity
+
+    @staticmethod
+    def _hash(key: Any) -> int:
```
LiudmylaKulinchenko, line 51: If you'll add Hashable in type annotations, you could remove this function. It will raise TypeError with message "unhashable type: 'list'", for example for list
Which is the same as in the dict

app/main.py
```diff
@@ -1,2 +1,76 @@
...
+
+    def _check_resize(self) -> None:
+        load_factor = self.size / self.capacity
+        if load_factor >= self.load_factor:
+            self._resize()
+
+    def _resize(self) -> None:
+        new_capacity = self.capacity * 2
+        new_table = [None] * new_capacity
+        for nodes in self.table:
```
LiudmylaKulinchenko, line 64: try to optimize to avoid nested loops and conditions

LiudmylaKulinchenko (changes requested):

Oleg-s-s (changes requested):
`is not None` `is None` can be simplified

Oleg-s-s (approved):
Good job

PR #363, opened 2023-06-20
Solution

DmytroTsehelnyk:


app/main.py
```diff
@@ -1,12 +1,15 @@
...
 
 
-@dataclass
 class Node:
-    key: Any
-    value: Any
-    next = None
+    def __init__(
+        self,
+        key: Any,
```
OlhaTryhub, line 7: It would be better if you use "Hashable" type annotation for key 👀

OlhaTryhub (changes requested):

OlhaTryhub (approved):

Altahoma (changes requested):

Altahoma (approved):

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+
+class Node:
+    def __init__(
+        self,
+        key: Hashable,
+        value: Any
+    ) -> None:
+        self.key = key
+        self.value = value
+        self.next = None
```
LiudmylaKulinchenko, line 12: You should store hash(key) value in Node, as it works in dictionaries in Python
Also u can remove next from here, store just key, hash and value

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+        if self.table[index] is None:
+            self.table[index] = Node(key, value)
+            self.table_size += 1
+        else:
+            current = self.table[index]
+            while True:
+                if current.key == key:
+                    current.value = value
+                    return
+                if current.next is None:
```
LiudmylaKulinchenko, line 37: No need to add `is None`. Fix throuhout the code

LiudmylaKulinchenko (changes requested):

app/main.py
DmytroTsehelnyk: I can't remove 'next' because of AttributeError: 'Node' object has no attribute 'next'

LiudmylaKulinchenko (approved):
Well done! 🐣🌺

PR #364, opened 2023-06-20
Solution

imelnyk007:


erikagayan (approved):

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8, load_factor: float = 0.66) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.hash_table = [None] * 8
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
LiudmylaKulinchenko, line 11: Are you sure key could be any type? What about None, list? Key should have **hashable** type

PR #365, opened 2023-06-20
solution

YKonoplyov:


app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        return len(self.hash_table) - self.hash_table.count(None)
+
+    def check_for_presence(self,
+                           key: int | float | str | tuple) -> Node | bool:
+
+        for node in self.hash_table:
+
+            if node is not None and node.node_key == key:
+                return node
+
```
IhorVoskoboinikov, line 67: I think it's better to look for the index and then check it on None, so as not to do additional iterations. Lecture example:
        key_hash = hash(key)

        index = key_hash % self.capacity

        while (self.hash_table[index] is not None
               and self.hash_table[index].key != key):

        if self.hash_table[index] is None:
            raise KeyError(f"Key {key} doesn't exist!")

        return self.hash_table[index].value

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        if self.check_for_presence(key):
+            return self.check_for_presence(key).node_value
+
+        else:
+            raise KeyError(key)
+
+    def __len__(self) -> int:
+        return len(self.hash_table) - self.hash_table.count(None)
+
```
IhorVoskoboinikov, line 59: In check list good example is:

    def __len__(self) -> int:
        return self.length

IhorVoskoboinikov (changes requested):
Good code) look at the small remarks in the comments.

IhorVoskoboinikov (approved):
As for me now your code is better)

app/main.py
```diff
@@ -1,2 +1,91 @@
+from random import randint
+from typing import Any
+from dataclasses import dataclass
+
+
+@dataclass
+class Node:
+    node_key: int | float | str | tuple
+    node_hash: int
+    node_value: Any
```
LiudmylaKulinchenko, line 10: Why do you need to specify node_ here in every attribute? I think, its not necessary, because you anyway will take this attibute from a Node type. Also you don't have the same in dictionary like dictionary_length. What do you think?

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+        self.capacity = 8
+
+    def rearrange_dict(self) -> None:
+
+        old_dict = self.hash_table
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+
+        for node in old_dict:
+            if node is not None:
```
LiudmylaKulinchenko, line 30: You don't need `is not None`. Remove throughout the code

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+                self.__setitem__(node.node_key, node.node_value)
+                self.length -= 1
+
+    def __setitem__(self, key: int | float | str | tuple, value: Any) -> None:
```
LiudmylaKulinchenko, line 34: Key could be Hashable, not just int | float | str | tuple

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🚀

PR #367, opened 2023-06-20
My solution

kovaliskoveronika:


Tarasidze (changes requested):

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+        self.hash_table = ["_" for i in range(self.capacity)]
+
+    def collision(self, node: Node, node_ind: int) -> int:
+        curr_id = node_ind
+        while True:
+            if curr_id < len(self.hash_table):
+                if self.hash_table[curr_id] == "_":
+                    return curr_id
+                if self.hash_table[curr_id].key == node.key:
+                    return curr_id
```
thorium-zero: You could, probably, combine those

thorium-zero (approved):
there are some issues I'd address, but I'm not an expert, so...

app/main.py
```diff
@@ -1,2 +1,81 @@
+import dataclasses
+from typing import Any
+
+
+@dataclasses.dataclass()
+class Node:
+    key: Any
+    value: Any
```
LiudmylaKulinchenko, line 8: You should also store hash from a key in Node

app/main.py
```diff
@@ -1,2 +1,81 @@
+import dataclasses
+from typing import Any
+
+
+@dataclasses.dataclass()
```
LiudmylaKulinchenko, line 5: You can import just dataclass to avoid importing whole dataclasses module
Also you can remove parentheses here

app/main.py
```diff
@@ -1,2 +1,81 @@
...
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.threshold = round(self.capacity * 2 / 3)
+        self.hash_table = [None] * self.capacity
+
+    def collision(self, node: Node, node_ind: int) -> int:
+        curr_id = node_ind
```
LiudmylaKulinchenko, line 19: Do not abbreviate words if it is not really necessary, it saves you only few symbols. It will make code more readable

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        while True:
+            if curr_id < len(self.hash_table):
```
LiudmylaKulinchenko, line 21: You can move this check directly in the loop

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        for node in self.hash_table:
+            if node is not None:
+                if node.key == key:
+                    return node.value
+        raise KeyError
+
+    def __len__(self) -> int:
+        len_ = 0
+        for node in self.hash_table:
+            if node is not None:
```
LiudmylaKulinchenko, line 71: You don't need `is not None`. Remove throughout the code

app/main.py
kovaliskoveronika: But I can't change current_id to 0 if I haven't found free space. Or is there some way?

app/main.py
```diff
@@ -1,2 +1,83 @@
+from dataclasses import dataclass
+from typing import Any, Hashable
+
+
+@dataclass
+class Node:
+    key: Any
```
LiudmylaKulinchenko, line 7: Key should be hashable

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+    value: Any
+    hash_: Hashable
```
LiudmylaKulinchenko, line 9: What type does hash function return?
Also consider to rename this variable, to make it more descriptive

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+        self.capacity = 8
+        self.size = 0
+        self.threshold = round(self.capacity * 2 / 3)
+        self.hash_table = [None] * self.capacity
+
+    def collision(self, node: Node, node_ind: int) -> int:
+        current_id = node_ind
+        while True:
+            if current_id < len(self.hash_table):
+                if (self.hash_table[current_id] is None
```
LiudmylaKulinchenko, line 23: No need to add `is None`

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+        self.capacity *= 2
+        self.threshold = round(self.capacity * 2 / 3)
+        old_table = self.hash_table.copy()
+        self.hash_table = [None] * self.capacity
+        self.size = 0
+        for elem in old_table:
+            if elem:
+                self.__setitem__(elem.key, elem.value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
LiudmylaKulinchenko, line 41: Key should be hashable, not any

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+        new_node = Node(key, value, hash(key))
+        self.size += 1
+        if self.size <= self.threshold:
+            ind = hash(new_node.key) % self.capacity
```
LiudmylaKulinchenko, line 45: Use hashed key from the Node, don't calculate it again

LiudmylaKulinchenko, line 45: Rename ind variable to make it more descriptive

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
Good job

PR #369, opened 2023-06-20
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1,2 +1,77 @@
...
-    pass
+    def __init__(self) -> None:
+        self.length = 0
+        self.capacity = CAPACITY
+        self.hash_table: list[None | Node] = [None] * CAPACITY
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
LiudmylaKulinchenko, line 26: Fix annotations for key. It could not be any

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+        if self.length >= int(self.capacity * LOAD_FACTOR):
+            self.resize()
```
LiudmylaKulinchenko, line 28: Why are you making resize before you add a new key-value in a dict?

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+
+        hash_table[hash_index] = Node(key, key_hash, value)
+        self.length += 1
+
+    def __getitem__(self, key: Any) -> Any:
+        hash_key = hash(key)
+        hash_index = hash_key % self.capacity
+
+        while (
+                self.hash_table[hash_index] is not None
```
LiudmylaKulinchenko, line 56: No need to use `is None`, `is not None`. Fix throughout the code

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+    def __len__(self) -> int:
+        return self.length
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.length >= int(self.capacity * LOAD_FACTOR):
+            self.resize()
+
+        self.append_node_to_hash_table(key, value, self.hash_table)
+
+    def append_node_to_hash_table(
```
LiudmylaKulinchenko, line 32: Are you sure `append` is an appropriate name for this method?

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
Good job

PR #370, opened 2023-06-20
done

maccarets:


app/main.py
```diff
@@ -1,2 +1,43 @@
...
+            self.size += 1
+            self.table[index] = (key, hash(key), value)
+        elif self.table[index] is not None:
+            self.table[index] = (key, hash(key), value)
+        if self.size > self.capacity * self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._get_index(key)
+        if self.table[index] is None:
```
LiudmylaKulinchenko, line 23: Remove `is not None` and `is None` everywhere

app/main.py
```diff
@@ -1,2 +1,43 @@
...
+        self.size = 0
+        self.table = [None] * self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._get_index(key)
+        if self.table[index] is None:
+            self.size += 1
+            self.table[index] = (key, hash(key), value)
+        elif self.table[index] is not None:
+            self.table[index] = (key, hash(key), value)
```
LiudmylaKulinchenko, line 17: Make it DRY. Move `self.table[index] = (key, hash(key), value)` out of condition

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🔥🔥

PR #371, opened 2023-06-21
'Solution'

RedCatHot:


oleksiikolii (approved):

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+            index = (index + 1) % self.capacity
+        return index
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        old_table = self.table
+        self.table = [None] * self.capacity
+        self.size = 0
+        for node in old_table:
+            if node:
```
Oleg-s-s, line 40: Somewhere you have like this

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.get_index(key)
+        if self.table[index] is None:
+            raise KeyError(key)
+        self.table[index] = None
+        self.size -= 1
+
+    def __iter__(self) -> None:
+        for node in self.table:
+            if node is not None:
```
Oleg-s-s, line 56: And somewhere like this

Oleg-s-s (changes requested):
No need to use `is not None` `is None`

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+            if node:
+                self[node[0]] = node[2]
+
+    def clear(self) -> None:
+        self.size = 0
+        self.table = [None] * self.capacity
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.get_index(key)
+        if self.table[index] is None:
```
LiudmylaKulinchenko, line 49: Fix `is None` throughout the code according to previous comments

LiudmylaKulinchenko (changes requested):

Oleg-s-s (changes requested):
Fix that everywhere(`not None`, `is None`), not only where we tagged

Oleg-s-s (approved):
:+1:

PR #372, opened 2023-06-21
Develop ot

OlhaTryhub:


app/main.py
```diff
@@ -1,2 +1,79 @@
...
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.hash_table = [None] * self.capacity
+
+    def __getitem__(self, input_key: Hashable) -> None:
+        for item in self.hash_table:
+            if item is not None:
```
LiudmylaKulinchenko, line 14: No need to add `is not None`. Fix throughout the code

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+                if item[0] == input_key:
```
LiudmylaKulinchenko, line 15: try to avoid nested conditions here

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.hash_table = [None] * self.capacity
+
+    def __getitem__(self, input_key: Hashable) -> None:
```
LiudmylaKulinchenko, line 12: You should search by hash(key) in this method

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+    def clear(self) -> None:
+        self.__init__()
+
+    def update(self, other: Dictionary) -> None:
+        for item in other.hash_table:
+            if item is not None:
+                self[item[0]] = other[item[0]]
+
+    def pop(self, key: Hashable) -> Any:
+        if self[key]:
```
LiudmylaKulinchenko, line 70: you can utilize the `key in self` expression, which is more concise and efficient

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+            for i in range(self.capacity):
+                if self.hash_table[i] is not None:
+                    if self.hash_table[i][0] == key:
```
LiudmylaKulinchenko, line 73: You can use the `and` operator to combine multiple conditions within a single if statement and avoid nested conditions

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+                if input_key == item[0]:
+                    return item[2]
+
+    def clear(self) -> None:
+        self.__init__()
+
+    def update(self, other: Dictionary) -> None:
+        for item in other.hash_table:
+            if item:
+                self[item[0]] = other[item[0]]
```
Oleg-s-s, line 69: The update() method takes either a[ dictionary](https://www.programiz.com/python-programming/dictionary) or an iterable object of key/value pairs (generally[ tuples](https://www.programiz.com/python-programming/tuple)).
If update() is called without passing parameters, the dictionary remains unchanged.

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+
+    def pop(self, key: Hashable) -> Any:
+        if key in self:
+            for i in range(self.capacity):
+                if self.hash_table[i] and self.hash_table[i][0] == key:
+                    popped_value = self.hash_table[i]
+                    self.hash_table[i] = None
+                    return popped_value
+        else:
+            raise KeyError
```
Oleg-s-s, line 79: The pop() method returns:

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        if key in self:
+            for i in range(self.capacity):
+                if self.hash_table[i] and self.hash_table[i][0] == key:
+                    popped_value = self.hash_table[i]
+                    self.hash_table[i] = None
+                    return popped_value
+        elif default:
+            return default
+        else:
```
LiudmylaKulinchenko, line 86: No need to use `else` after return

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+        self.capacity *= 2
+        old_hash_table_items = [item for item in self.hash_table
+                                if item]
+        self.hash_table = [None] * self.capacity
+        for key, _hash, value in old_hash_table_items:
+            self[key] = value
+
+    def get(self, input_key: Hashable) -> Any:
+        for item in self.hash_table:
+            if item:
```
LiudmylaKulinchenko, line 59: use `and` to avoid nested conditions

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+                if self.hash_table[i]:
+                    if self.hash_table[i][0] == key:
+                        self.hash_table[i] = None
+                        break
+        else:
+            raise KeyError
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        old_hash_table_items = [item for item in self.hash_table
```
LiudmylaKulinchenko, line 51: make proper indentation in the list comrehention

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+        for item in self.hash_table:
+            if item:
+                len_counter += 1
+        return len_counter
+
+    def __delitem__(self, key: Hashable) -> None:
+        if key in self:
+            for i in range(self.capacity):
+                if self.hash_table[i]:
+                    if self.hash_table[i][0] == key:
```
LiudmylaKulinchenko, line 43: you can avoid nested loops in this case by using the and operator

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
Good job

PR #375, opened 2023-06-21
Solution

Kasatello:


Oleg-s-s (changes requested):
No need to use `is None` `is not None`. Fix everywhere

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+
+    def resize(self) -> None:
+        old_nodes = []
+        for node in self.hash_table:
+            if node:
+                old_nodes.append(node)
+        self.capacity = self.capacity * 2
+        new_hash_table = [None] * self.capacity
+        for node in old_nodes:
+            index = self.get_index(node.key)
```
Oleg-s-s, line 50: Same as 18, 31. Try to optimize this.
Also try to optimize resize function, it can be simplified

Oleg-s-s (changes requested):

Oleg-s-s (approved):
:+1:

PR #376, opened 2023-06-21
My Dictionary 

IhorVoskoboinikov:


app/main.py
```diff
@@ -1,2 +1,79 @@
...
+INITIAL_CAPACITY = 8
+RESIZE_THRESHOLD = 2 / 3
+CAPACITY_MULTIPLIER = 2
+
+
+@dataclass
+class Node:
+    key: Hashable
+    value: Any
+    hash_data: int
```
LiudmylaKulinchenko, line 13: It's better to clarify that it is hash from key

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+
+        self.hash_table[key_index] = Node(key, value, hash_key)
+        self.length += 1
+
+    def resize_table(self) -> None:
+        self.length = 0
+        self.capacity *= CAPACITY_MULTIPLIER
+
+        new_hash_table: List[Node | None] = [None] * self.capacity
+        node_data = [node for node in self.hash_table if node is not None]
```
LiudmylaKulinchenko, line 48: No need to add `is not None`. Fix throughout the code

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+
+        hash_key = hash(key)
+        key_index = hash_key % self.capacity
+
+        while self.hash_table[key_index]:
+            if (self.hash_table[key_index].hash_data == hash_key
+                    and self.hash_table[key_index].key == key):
+                self.hash_table[key_index].value = value
+                return
+            else:
```
LiudmylaKulinchenko, line 37: You can skip `else` after `return`

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
Good job :+1:

PR #377, opened 2023-06-21
Solution

erikagayan:


app/main.py
```diff
@@ -1,2 +1,71 @@
...
+    @property
+    def current_max_size(self) -> float:
+        return RESIZE_THRESHOLD * self.capacity
+
+    def _calculate_index(self, key: Hashable) -> int:
+        index = hash(key) % self.capacity
+        while (
+                self.hash_table[index] is not None
+                and self.hash_table[index].key != key
+        ):
```
Oleg-s-s, line 30: No need to write `is not None` here
Fix everywhere

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+
+        return index
+
+    def _cyclic_increment(self, index: int) -> int:
+        return (index + 1) % self.capacity
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
```
Oleg-s-s, line 41: No need to write `is None` 
Fix everywhere

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+        if not self.hash_table[index]:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        return self.hash_table[index].value
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._calculate_index(key)
+
+        if not self.hash_table[index]:  # add new element
+            if self.size + 1 >= self.current_max_size:
```
LiudmylaKulinchenko, line 50: You can remove comments

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+
+INITIAL_CAPACITY = 8
+RESIZE_THRESHOLD = 2 / 3
+CAPACITY_MULTIPLIER = 2
+
+
+@dataclass
+class Node:
+    key: Hashable
+    value: Any
```
LiudmylaKulinchenko, line 12: You should also store hash from a key value in Node

Oleg-s-s (approved):
:+1:

PR #380, opened 2023-06-21
my_commit

Mr-Furman:


Oleg-s-s (changes requested):
Fix `is None` and `is not None` everywhere, also if you copy-paste from lecture at least delete comments)

LiudmylaKulinchenko (approved):
Good job 🐞🐣

PR #381, opened 2023-06-21
Develop

Tarasidze:


Oleg-s-s (changes requested):
No need to use `is None`, `is not None`. Fix it

LiudmylaKulinchenko (approved):
Good job 🍬🌼

PR #382, opened 2023-06-21
Solution

Po-SvinSkipodla:


Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+            return default
+        raise KeyError
+
+    def update(self, other: "Dictionary" = None) -> None:
+        if not other:
+            return
+        if isinstance(other, dict):
+            other = other.items()
+        for key, value in other:
+            self[key] = value
```
Oleg-s-s, line 99: previous comment not fixed

Oleg-s-s (changes requested):

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+                list[set],
+                list[list]
+            ] = None
+    ) -> None:
+        if not other:
+            return
+        if isinstance(other, dict):
+            other = other.items()
+        for key, value in other:
+            self[key] = value
```
Oleg-s-s, line 108: You should be attentive to the comments. what if other is type of set, list, tuple? Does it have key, value?

Po-SvinSkipodla: I don`t understand. You mean just set or list of sets? Just tuple or list of tuples?

LiudmylaKulinchenko (approved):
🌺🎈

PR #383, opened 2023-06-25
Solution

volodymyr-komarnyckyi:


Oleg-s-s (changes requested):
Fix `is None` `is not None`, etc. No need to write like that. Also if you are copying from lecture, at least delete comments :upside_down_face:

Oleg-s-s (approved):
Good job

PR #386, opened 2023-06-29
Solution

arturiermolenko:


app/main.py
```diff
@@ -1,2 +1,76 @@
...
+                self.__setitem__(node[0], node[2])
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __str__(self) -> str:
+        return str({node[0]: node[2] for node in self.hash_table if node})
+
+    def clear(self) -> None:
+        self.hash_table = [None] * self.capacity
```
YukihiroSM, line 56: length should be set to 0 as well here.

YukihiroSM (changes requested):

PR #387, opened 2023-06-29
solution

vlodek69:


app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        self.length = 0
+        self.hash_table: list = [None] * 8
+        self.hash_table_len = 8
+        self.load_factor = 2 / 3
+        self.item_id = 0
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
vasylhnatiuk, line 20:
```suggestion
    def __setitem__(self, key:  Hashable, value: Any) -> None:
```

vasylhnatiuk: change here and in all of the code

vasylhnatiuk (changes requested):

vsmutok (approved):
Looks Good 👍

PR #396, opened 2023-07-12
Solution

Quiet-Klirik:


tests/test_main.py
Polyakiv-Andrey, line 1: In this task, the test file does not need to be touched at all.

app/main.py
```diff
@@ -1,2 +1,97 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
```
Polyakiv-Andrey, line 5: also in the initializer it is necessary to use the load factor and capacity for this job

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+        self.length = 0
+        self.hash_table: list[None | tuple[Hashable, hash, Any]] = [None] * 8
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        def add_item_to_table(
```
Polyakiv-Andrey, line 10: instead of writing a function inside a magic method, make it just a class method

Polyakiv-Andrey (changes requested):

tests/test_main.py
Quiet-Klirik: checklist.md
--------------
2) If you want check your `__delitem__` method 
uncomment `test_deletion():` test in `./tests/test_main.py`

Polyakiv-Andrey (approved):
Good job!

PR #400, opened 2023-07-24
Implemented Dictionary class

lubart2517:


app/main.py
```diff
@@ -1,2 +1,145 @@
...
+        if self.hash_table[key_index]:
+            if self.hash_table[key_index][1] == key:
+                self.hash_table[key_index][2] = value
+            else:
+                added = False  # check if key was added to dict
+                for current_index in range(
+                        key_index + 1,
+                        len(self.hash_table)
+                ):
+                    if self.hash_table[current_index]:
```
pavlejviki, line 29: try to avoid nested conditions here

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+                            key,
+                            value
+                        ]
+                        added = True
+                        if not reset:
+                            self.length += 1
+                        break
+                if not added:
+                    for current_index in range(0, key_index - 1):
+                        if self.hash_table[current_index]:
```
pavlejviki, line 46: try to avoid nested conditions here

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+            if elem:
+                self.set(elem[1], elem[2], True)
+        self.cur_hash_table = []
+
+    def find_table_index(self, key: Hashable) -> int:
+        """Search key in hash table and returns its index
+        in hash table"""
+        hash_key = hash(key)
+        key_index = hash_key % len(self.hash_table)
+        if self.hash_table[key_index]:
```
pavlejviki, line 86: try to avoid nested conditions here

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+            if self.hash_table[key_index][1] == key:
+                return key_index
+        for cur_index in range(key_index + 1, len(self.hash_table)):
+            if self.hash_table[cur_index]:
```
pavlejviki, line 90: try to avoid nested conditions here

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+                if self.hash_table[cur_index][1] == key:
+                    return cur_index
+        for cur_index in range(0, key_index - 1):
+            if self.hash_table[cur_index]:
```
pavlejviki, line 94: try to avoid nested conditions here

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+    def clear(self) -> None:
+        """Clear all records in Dictionary"""
+        self.hash_table = [None] * 8
+        self.length = 0
+
+    def get(self, key: Hashable) -> Any:
+        """ Returns value of given key"""
+        return self.__getitem__(key)
+
+    def pop(self, key: Hashable) -> Any:
```
pavlejviki, line 127: pop method allows for a second, optional argument which will serve as the default value to return if the key is not found. Implement this logic please

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+        index = self.find_table_index(key)
+        self.hash_table[index] = None
+        self.length -= 1
+
+    def clear(self) -> None:
+        """Clear all records in Dictionary"""
+        self.hash_table = [None] * 8
+        self.length = 0
+
+    def get(self, key: Hashable) -> Any:
```
pavlejviki, line 123: get method allows for a second, optional argument which will serve as the default value to return if the key is not found. And get method returns None if the key is not found,  now it raises a keyerror

Implement this logic please

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+    def pop(self, key: Hashable) -> Any:
+        """ Deletes key/value pair
+        and returns value of given key"""
+        index = self.find_table_index(key)
+        elem = self.hash_table[index][2]
+        self.hash_table[index] = None
+        self.length -= 1
+        return elem
+
+    def update(self, key: Hashable, value: Any) -> None:
```
pavlejviki, line 136: in Python's built-in dict class, the update method can accept another dictionary, an iterable of key-value pairs, and it adds all the keys to the dictionary, overwriting the value of the key if it already exists.
 
 so this method can except more 1 key,value pair

app/main.py
```diff
@@ -1,2 +1,145 @@
...
+        """Reassign value of given key"""
+        index = self.find_table_index(key)
+        self.hash_table[index][2] = value
+
+    def __iter__(self) -> Hashable:
```
pavlejviki, line 141: Your current implementation of __iter__ is almost correct, but you're yielding the entire cell (which is a list) instead of just the key. In the context of a dictionary, typically you'd want to yield just the keys when iterating.

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,2 +1,140 @@
...
+    def pop(self, key: Hashable) -> Any:
+        """ Deletes key/value pair
+        and returns value of given key"""
+        index = self.find_table_index(key)
+        elem = self.hash_table[index].value
+        self.hash_table[index] = None
+        self.length -= 1
+        return elem
+
+    def update(self, *args) -> None:
```
pavlejviki, line 119: update

pavlejviki: the correct parameters for update are 

def update(self, _m=None, **kwargs):

The use of a default value for the _m parameter allows the update method to handle cases where the method is called with or without an argument. If you don't provide an argument, _m will be None, and you can handle this case in the method's implementation accordingly.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #401, opened 2023-07-26
'Solution'

ilizh:


app/main.py
```diff
@@ -1,2 +1,128 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: Any, value: Any, hash_code: int) -> None:
```
olena-hul, line 5: Uncomment test for deletion

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        self.value = value
+        self.hash_code = hash_code
+        self.next = None
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self, initial_capacity: int = 16,
+            load_factor: float = 0.75
```
olena-hul, line 15: No need to make it as the `__init__` params, these params are the details of internal implementation, user should not know about this

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+
+        for node in old_table:
+            while node:
+                self[node.key] = node.value
+                node = node.next
+
+    def _should_resize(self) -> bool:
+        return self.size >= self.capacity * self.load_factor
+
+    def _get_node(self, key: Any, hash_code: int) -> Node:
```
olena-hul, line 39: key is Hashable, not Any

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        node = self.table[index]
+
+        while node:
+            if node.hash_code == hash_code and node.key == key:
+                return node
+            node = node.next
+
+        return None
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
olena-hul, line 50: key is Hashable, not Any

olena-hul: Fix everywhere

olena-hul (changes requested):

app/main.py
olena-hul: This is not fixed. You should go to tests file and uncomment the test that checks the delete functionality for the dictionary

olena-hul (approved):

olena-hul:
Good job!

PR #404, opened 2023-07-31
Solution

anastasiia-tsurkan:


app/main.py
```diff
@@ -1,2 +1,131 @@
...
+        self.load_factor = round(self.capacity * 2 / 3)
+        self.table = [[] for i in range(self.capacity)]
+
+        for node in temp:
+            if node:
+                self.__setitem__(node[0], node[2])
+
+    def __setitem__(
+            self,
+            key: Union[int, float, str, bool, tuple, Point],
```
olena-hul, line 27:
```suggestion
            key: Hashable,
```

app/main.py
```diff
@@ -1,2 +1,131 @@
...
+        if key_index is not None:
+            self.table[key_index] = [key, hash(key), value]
+            return
+
+        for i in range(self.capacity):
+            node = self.table[index]
+
+            if node:
+                index = (index + 1) % self.capacity
+            if not node:
```
olena-hul, line 45:
```suggestion
            else:
```

app/main.py
olena-hul: Fix in all places

app/main.py
```diff
@@ -1,2 +1,131 @@
...
+    def get_index(
+            self,
+            key: Union[int, float, str, bool, tuple, Point]
+    ) -> Union[int, None]:
+        index = hash(key) % self.capacity
+
+        for _ in range(self.capacity):
+            node = self.table[index]
+
+            if node and node[0] == key:
```
olena-hul, line 59: we should check hash too, hash should be checked first

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #408, opened 2023-08-07
Develop

verbik:


app/main.py
```diff
@@ -1,2 +1,79 @@
...
+    INITIAL_CAPACITY = 8
+
+    def __init__(self) -> None:
+        self.capacity = self.INITIAL_CAPACITY
+        self.size = 0
+        self.table = [None] * self.capacity
+
+    @property
+    def size(self) -> int:
+        return self._size
```
Grayder0152, line 16: But you don`t have `_size` arg in `__init__` method

app/main.py
```diff
@@ -1,2 +1,79 @@
...
+        return self.size
+
+    def resize(self) -> None:
+        elements = [elem for elem in self.table if elem is not None]
+        print(elements)
+        self.table = [None] * self.capacity * 2
+        self.capacity *= 2
+
+        for element in elements:
+            self.__setitem__(element[0], element[2])
```
Grayder0152, line 36:
```suggestion
            self[element[0]] = element[2]
```

Grayder0152 (changes requested):

app/main.py
verbik: I have a setter for this property and it is in __init__. Can you explain in more details, please.

Grayder0152: Is it unclear, as we must have all exist attribute in __init__, so pls, rename `size` to `_size` in 11 line

ArsenPidhoretskyi (approved):
Well done!

PR #409, opened 2023-08-10
Solution

mwellick:


app/main.py
```diff
@@ -1,2 +1,70 @@
...
+
+from typing import Any, Hashable
+
+
+@dataclasses.dataclass
 class Dictionary:
-    pass
+    capacity: int = 8
+    load_factor: float = 0.75
+    hash_table_: list = None
```
Bohdan-Salamakha, line 10: use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contains string elements and can be None has typehint Optional[list[str]]
+
add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+movies.clear()
+print(len(movies))
+
+movies.__setitem__(2, "Memento")
+print(len(movies))
+
+print(movies.__getitem__(2))
+movies.__setitem__(3, "Tenet")
+print(movies.__getitem__(3))
```
EdAlekseiev, line 102: Please, delete your test code

EdAlekseiev (approved):
Good job

app/main.py
mwellick: sorry i forgot)

PR #414, opened 2023-08-15
 Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,2 +1,105 @@
...
+
+    def __len__(self) -> int:
+        return self.size
+
+    def get_index_for_key(self,
+                          key: Hashable
+                          ) -> Any:
+        index = hash(key) % self.capacity
+        while (self.hash_table[index] is not None
+               and self.hash_table[index][0] != key):
```
ArsenPidhoretskyi, line 29: Also check for hash

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+        self.hash_table = [None] * self.capacity
+
+    @property
+    def current_max_size(self) -> int:
+        return floor(self.load_factor * self.capacity)
+
+    def __len__(self) -> int:
+        return self.size
+
+    def get_index_for_key(self,
```
ArsenPidhoretskyi, line 24: Make it as private

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+                          key: Hashable
+                          ) -> Any:
+        index = hash(key) % self.capacity
+        while (self.hash_table[index] is not None
+               and self.hash_table[index][0] != key):
+            index = (index + 1) % self.capacity
+        return index
+
+    def resize(self) -> None:
```
ArsenPidhoretskyi, line 33: Make it as private function or protected

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+
+    def clear(self) -> None:
+        self.hash_table = [None] * self.capacity
+        self.size = 0
+
+    def get(self,
+            key: Hashable,
+            value: Any = None
+            ) -> Any:
+        for index in range(len(self.hash_table)):
```
ArsenPidhoretskyi, line 53: Why you always iterate from start? Get it based by hash

ArsenPidhoretskyi: Fix in all places

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+        old_table = self.hash_table
+        self.size = 0
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+        for element in old_table:
+            if element is not None:
+                key, hash_key, value = element
+                index = self.get_index_for_key(key)
+                self.size += 1
+                self.hash_table[index] = (key, hash_key, value)
```
ArsenPidhoretskyi, line 43:
```suggestion
                self[key] = value
```

ArsenPidhoretskyi (changes requested):

PR #415, opened 2023-08-17
'Solution'

niksgawrilenko:


app/main.py
```diff
@@ -1,2 +1,46 @@
...
+                    self.bucket[empty_slot][1] == hash(key)):
+                return self.bucket[empty_slot][2]
+            empty_slot = (empty_slot + 1) % self.capacity
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self.length
+
+    def resize(self) -> None:
+        bucket1 = self.bucket
```
Polyakiv-Andrey, line 39: give a meaningful name

app/main.py
```diff
@@ -1,2 +1,46 @@
...
+            if key == self.bucket[empty_slot][0] and (
+                    self.bucket[empty_slot][1] == hash(key)):
+                return self.bucket[empty_slot][2]
+            empty_slot = (empty_slot + 1) % self.capacity
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self.length
+
+    def resize(self) -> None:
```
Polyakiv-Andrey, line 38: add encapsulation for this method, it can only be used in this class

Polyakiv-Andrey (changes requested):

app/main.py
niksgawrilenko: Done

app/main.py
niksgawrilenko: Done

PR #418, opened 2023-08-23
Solution

AlexPelevin99:


app/main.py
```diff
@@ -1,2 +1,168 @@
...
+    ("two", 222),
+    ("two", 2222),
+    ("two", 22222),
+    (Point(1, 1), "A"),
+]
+dictionary = Dictionary()
+for key, value in items:
+    dictionary[key] = value
+
+print(len(dictionary), dictionary)
```
Dimosphen1, line 168: You do not need to include your test code in the PR, remove it.

app/main.py
```diff
@@ -1,2 +1,168 @@
...
+
+
 class Dictionary:
-    pass
+    __items: List[tuple | None]
+    __current_len: int = 0
+
+    def __init__(
+        self,
+        args: List[tuple[Any, Any] | list[Any, Any]] | None = None
```
Dimosphen1, line 12: You should use either objects from `typing` or built-in data types for type hinting, do not mix them.

app/main.py
```diff
@@ -1,2 +1,168 @@
...
+        current_index = self.__find_item_index_by_key(key)
+        item_value = self.__items[current_index][1]
+        self.__del_item_by_index(current_index)
+        return item_value
+
+    def update(self, other_dict: Any) -> None:
+        if isinstance(other_dict, Dictionary):
+            other_dict_items = other_dict.items()
+        else:
+            other_dict_items = other_dict
```
Dimosphen1, line 136: There is no need in this `else` statement, just rewrite the content of `other_dict` in the first if condition.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,2 +1,135 @@
...
+
+
 class Dictionary:
-    pass
+    __items: list[tuple | None]
+    __current_len: int = 0
+
+    def __init__(
+        self,
+        args: list[tuple[Any, Any] | list[Any, Any]] | None = None
```
Dimosphen1, line 10: It is acceptable to use `Any` from `typing` but remember that we can use the `any` keyword as well.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #421, opened 2023-08-24
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,2 +1,107 @@
...
+                 capacity: int = 8,
+                 load_factor: float = 2 / 3
+                 ) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.hash_table: Optional[List[Node]] = [None] * self.capacity
+
+    def __len__(self
+                ) -> int:
```
ArsenPidhoretskyi, line 25:
```suggestion
    def __len__(self) -> int:
```

ArsenPidhoretskyi: Fix in all places

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+
+    def _resize(self
+                ) -> None:
+        old_table = [node for node in self.hash_table if node]
+        self.capacity *= 2
+        self.clear()
+        for node in old_table:
+            index = self._index_for_key(node.key)
+            self.hash_table[index] = node
+            self.size += 1
```
ArsenPidhoretskyi, line 51: self[key] = value

app/main.py
```diff
@@ -1,2 +1,107 @@
...
+            value = self.hash_table[index].value
+        self.__delitem__(key)
+        return value
+
+    def update(self,
+               other_dictionary: Dictionary
+               ) -> None:
+        for node in other_dictionary.hash_table:
+            if node:
+                self.__setitem__(node.key, node.value)
```
ArsenPidhoretskyi, line 102: Use self[key] = value

ArsenPidhoretskyi (changes requested):

PR #422, opened 2023-08-24
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,2 +1,81 @@
...
+            return self.hash_table[index].value
+        raise KeyError
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._index_for_key(key)
+        self.hash_table[index] = None
+        self.size -= 1
+
+    def get(self, key: Hashable) -> Any:
+        return self[key]
```
ArsenPidhoretskyi, line 65: .get also accepts default parameter

ArsenPidhoretskyi (changes requested):

PR #423, opened 2023-08-24
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,2 +1,83 @@
...
+
+    def clear(self) -> None:
+        self.size = 0
+        self.hash_table = [None] * self.capacity
+
+    def _index_for_key(self, key: Hashable) -> int:
+        index = hash(key) % self.capacity
+        while True:
+            if (self.hash_table[index] is None
+                    or self.hash_table[index].key == key):
```
ArsenPidhoretskyi, line 32: Also better to check hash

app/main.py
```diff
@@ -1,2 +1,83 @@
...
+            return self.hash_table[index].value
+        raise KeyError
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._index_for_key(key)
+        self.hash_table[index] = None
+        self.size -= 1
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        if self[key]:
```
ArsenPidhoretskyi, line 65: In this case you will get key error

ArsenPidhoretskyi (changes requested):

PR #424, opened 2023-08-24
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,2 +1,85 @@
...
+    hash_of_node: int
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8, load_factor: float = 2 / 3) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.hash_table: Optional[List[Node]] = [None] * self.capacity
```
ArsenPidhoretskyi, line 20:
```suggestion
        self.hash_table: List[Optional[Node]] = [None] * self.capacity
```

app/main.py
```diff
@@ -1,2 +1,85 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._index_for_key(key)
+        self.hash_table[index] = None
+        self.size -= 1
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        if self[key]:
+            return self[key]
+        raise KeyError
```
ArsenPidhoretskyi, line 69: Use try-except

app/main.py
```diff
@@ -1,2 +1,85 @@
...
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
```
ArsenPidhoretskyi, line 71: Here you also need to use try-except

ArsenPidhoretskyi (changes requested):

PR #425, opened 2023-08-25
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,2 +1,87 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            value = self[key]
+            self.__delitem__(key)
+            return value
+        except KeyError:
+            return default
```
ArsenPidhoretskyi, line 78: Not always have two different cases:
d = {}
d.pop(1) # Raise KeyError
d.pop(1, None) # Will not raise

ArsenPidhoretskyi (changes requested):

PR #426, opened 2023-08-25
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,2 +1,87 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        if key in self:
+            value = self[key]
+            self.__delitem__(key)
+            return value
+        else:
+            return default
```
ArsenPidhoretskyi, line 78: Use try-except
Also do not forget about:
d = {}
d.pop(1) # Raise KeyError
d.pop(1, None) # Not raise error

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+        self.hash_table = [None] * self.capacity
+
+    def _index_for_key(self, key: Hashable) -> int:
+        index = hash(key) % self.capacity
+        while True:
+            if (self.hash_table[index] is None
+                    or (self.hash_table[index].key == key
+                        and self.hash_table[index].hash_of_node == hash(key))):
+                return index
+            index = (index + 1) % self.capacity
```
ArsenPidhoretskyi, line 36:
```
while not (self.hash_table[index] is None or 
                (self.hash_table[index].key == key and 
                    self.hash_table[index].hash_of_node == hash(key))):
return index
```
As for me return will be the same but without while True

ArsenPidhoretskyi (changes requested):

PR #428, opened 2023-08-25
raw solution, most likely will be improved, comments and critics are Welcome still

eduardhabryd:
…Welcome still

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,2 +1,146 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: str, _hash: int, value: str) -> None:
+        self._key = key
+        self._hash = _hash
+        self._value = value
+
+    @property
+    def key(self) -> str:
+        return self._key
+
+    @key.setter
+    def key(self, key: str) -> None:
+        self._key = key
+
+    @property
+    def value(self) -> str:
+        return self._value
+
+    @value.setter
+    def value(self, value: str) -> None:
+        self._value = value
+
+    @property
+    def hash(self) -> int:
+        return self._hash
+
+    @hash.setter
+    def hash(self, _hash: int) -> None:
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -1,2 +1,99 @@
...
+        node = self._hash_table[index]
+        self._hash_table[index] = None
+        return node.value
+
+    def __iter__(self) -> iter:
+        for node in self._hash_table:
+            if node is not None:
+                yield node.key
+
+    def update(self, other: dict = None, **kwargs: str) -> None:
```
Kroll410, line 90: update should be surely working with your Dict class implementation

Kroll410 (changes requested):

Kroll410:
GJ

app/main.py
eduardhabryd: > update should be surely working with your Dict class implementation

update is working fine as I tested. Could you please elaborate on your comment? Should I change logic in the update method? How?

OlegatorLE (approved):
Wow! VGJ!

app/main.py
Kroll410: if you're expecting `other` to be `dict` type, that's fine. But we're trying to implement our custom. So you should expect your Dictionary class, which doesn't have `items()` method, so it goes through iterating `key, value in items`. `__iter__` method return only the key, so as I see unpacking will fail there.

eduardhabryd: > if you're expecting `other` to be `dict` type, that's fine. But we're trying to implement our custom. So you should expect your Dictionary class, which doesn't have `items()` method, so it goes through iterating `key, value in items`. `__iter__` method return only the key, so as I see unpacking will fail there.

Thanks for the clarification! I've improved my code, please check again.

PR #429, opened 2023-08-26
Solution

AndriiIshchenko:


app/main.py
```diff
@@ -1,2 +1,115 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.length + 1 >= len(self.hash_table) * self.load_factor:
+            self.extend_hash_table()
+        self.set_node(key, value)
+
+    def __getitem__(self, key: Hashable, return_index: bool = False) -> Any:
+        index = self.hash_index(key)
+        count = 0
+        for i in range(len(self.hash_table)):
```
vasylhnatiuk, line 61: if you don't use i, better use `_`

vasylhnatiuk, line 61:
```suggestion
        for _ in range(len(self.hash_table)):
```

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+        self.value = value
+
+    def __repr__(self) -> str:
+        return f"{self.key} : {self.value}"
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.load_factor: float = 0.75
```
vasylhnatiuk, line 16: load factor and other variables that cannot be changed better write like a class constant

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -12,12 +12,14 @@ def __repr__(self) -> str:
...
+    LOAD_FACTOR = 0.75
+    INITIAL_LENGTH = 8
+
     def __init__(self) -> None:
-        self.load_factor: float = 0.75
         self.length: int = 0
-        self.initial_length: int = 8
-        self.hash_table_length: int = self.initial_length
-        self.hash_table: list[Union[Node, None]] = [None] * self.initial_length
+        self.INITIAL_LENGTH: int = 8
```
vasylhnatiuk, line 20: why do you need it here if you already have it in 16 line? delete pls

vasylhnatiuk (changes requested):

AndriiIshchenko:
sorry I copied the line and forgot to remove unnecessary one

vasylhnatiuk (approved):

PR #433, opened 2023-08-29
custom dict implemented

maxkatkalov:


OlegatorLE (approved):
GJ!

Kroll410 (changes requested):
If you don't want to create other methods - re-request review and it'll be approved

Kroll410 (approved):

PR #434, opened 2023-08-29
Solution

OlegatorLE:


Kroll410 (changes requested):
Looks good, if you don't want to implement rest methods - re-request review

innyshka (approved):
Good job!!!

PR #435, opened 2023-08-29
Solution

TarasSavchyn:


app/main.py
```diff
@@ -1,2 +1,68 @@
...
+    def __setitem__(self, key: Any, value: Any) -> None:
+        hash_value = self._hash_function(key)
+        new_node = Node(key, value)
+
+        if self.table[hash_value] is None:
+            self.table[hash_value] = new_node
+        else:
+            current_node = self.table[hash_value]
+            while current_node:
+                if current_node.key == key:
```
ArsenPidhoretskyi, line 44: Also check hash

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+        self.size += 1
+        if self.size / self.capacity >= self.load_factor:
+            self._resize(self.capacity * 2)
+
+    def __getitem__(self, key: Any) -> None:
+        hash_value = self._hash_function(key)
+        current_node = self.table[hash_value]
+
+        while current_node:
+            if current_node.key == key:
```
ArsenPidhoretskyi, line 61: Also check hash

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,2 +1,77 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: Any, value: Any) -> None:
```
vsmutok, line 5:
```suggestion
    def __init__(self, key: Hashable, value: Any) -> None:
```

vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
ArsenPidhoretskyi: You must check for the key
current_node.key == key and current_node.hash == key_hash

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #436, opened 2023-08-29
Implemented solution

RPodolets:


nyanplague (approved):
may the force be with you

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+    hash_value: int
+    value: Any
+
+    def __repr__(self) -> str:
+        return f"<Node: key={self.key} value={self.value}>"
+
+
 class Dictionary:
-    pass
+    __items: List[Optional[Node]]
```
Kroll410, line 19: why is this class attribute? So all Dictinary objects I create will share the same items..?

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+            return default
+
+    def pop(self, key: Hashable) -> Any:
+        value = self.get(key)
+        self.__delitem__(key)
+        return value
+
+    def update(self, other: Dictionary) -> None:
+        for item in other:
+            key, value = item
```
Kroll410, line 104: Did you test update?

Kroll410: Aha, I saw `__next__`, gj

Kroll410 (changes requested):
Gj, but only resolve comment

PR #442, opened 2023-08-30
Solved a task, implemented all additional functions. Didn't implement whole functions of dict

ORENYT:
I used special functions to implement default dict functional, also used additional tests for delete object.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -60,12 +60,9 @@ class Dictionary:
             return -1
 
     def update(self, second_dict: "Dictionary") -> None:
-        nodes = [
-            node
-            for node in second_dict.__dict__["_Dictionary__hashtable"]
-            if node is not None
-        ]
-        for node in nodes:
+        for node in second_dict.__dict__["_Dictionary__hashtable"]:
+            if node:
+                self.__setitem__(node.key, node.value)
             self.__setitem__(node.key, node.value)
 
     def __iter__(self) -> list:
@@ -95,7 +92,7 @@ class Dictionary:
         be multiplied automatically
         """
         length = len(self.__hashtable)
-        if length - self.__hashtable.count(None) > length * self.__load_factor:
+        if self.__len__() > length * self.__load_factor:
             self.__multiply_hashtable()
 
     def __multiply_hashtable(self) -> None:
@@ -110,18 +107,14 @@ class Dictionary:
     def __choose_cell(self, node: "Node") -> None:
         """
         Do not use this function, it chose where to mark data
-        according to it's hash
+        according to its hash. We use len + cell to go to the 0
+        if we are in top border
         """
         length = len(self.__hashtable)
         preferred_cell = node.key_hash % length
-        for i in range(preferred_cell, length):
-            if self.__hashtable[i] and self.__hashtable[i].key == node.key:
-                self.__hashtable[i] = node
-                return
-            if self.__hashtable[i] is None:
-                self.__hashtable[i] = node
-                return
-        for i in range(0, preferred_cell):
+        for i in range(preferred_cell, length + preferred_cell):
+            if i >= len(self.__hashtable):
+                i -= length
             if self.__hashtable[i] and self.__hashtable[i].key == node.key:
                 self.__hashtable[i] = node
                 return
--- a/app/main.py
+++ b/app/main.py
@@ -1,11 +1,11 @@
 import dataclasses
-from app.point import Point
+from typing import Hashable
 
 
 @dataclasses.dataclass
 class Node:
     key: object
-    hash: int
+    key_hash: int
     value: object
 
 
@@ -15,13 +15,13 @@ class Dictionary:
     def __init__(self) -> None:
         self.__hashtable = [None] * 8
 
-    def __setitem__(self, key, value) -> None:
+    def __setitem__(self, key: Hashable, value: object) -> None:
         hashed_key = hash(key)
         node = Node(key, hashed_key, value)
         self.__choose_cell(node)
         self.__check_overload()
 
-    def __getitem__(self, key) -> object:
+    def __getitem__(self, key: Hashable) -> object:
         hashed_key = hash(key)
         length = len(self.__hashtable)
         preferred_cell = hashed_key % length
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -1,2 +1,130 @@
...
+        except KeyError:
+            return -1
+
+    def update(self, second_dict: "Dictionary") -> None:
+        nodes = [
+            node
+            for node in second_dict.__dict__["_Dictionary__hashtable"]
+            if node is not None
+        ]
+        for node in nodes:
```
denys-source, line 68: I would recommend you using one loop like that:
```
for node in second_dict.__dict__["_Dictionary__hashtable"]:
    if not is not None:
        self.__setitem__(node.key, node.value)
```

app/main.py
```diff
@@ -1,2 +1,130 @@
...
+        preferred_cell = node.key_hash % length
+        for i in range(preferred_cell, length):
+            if self.__hashtable[i] and self.__hashtable[i].key == node.key:
+                self.__hashtable[i] = node
+                return
+            if self.__hashtable[i] is None:
+                self.__hashtable[i] = node
+                return
+        for i in range(0, preferred_cell):
+            if self.__hashtable[i] and self.__hashtable[i].key == node.key:
```
denys-source, line 125: I think one `if` condition with `or` operator will remove duplicate code

MarianKovalyshyn (approved):
I think you could replace these loops with one loop because they are very similar and they can confuse at first look. But everything else looks good to me.
```
for i in range(preferred_cell, length):
    if self.__hashtable[i] and self.__hashtable[i].key == key:
        return self.__hashtable[i].value
for i in range(0, preferred_cell):
    if self.__hashtable[i] and self.__hashtable[i].key == key:
        return self.__hashtable[i].value
 ```

File: app/main.py [hunk 3]
```diff
@@ -1,2 +1,130 @@
...
+        self.__delitem__(key)
+        return item
+
+    def __check_overload(self) -> None:
+        """
+        Do not use this function, it calculates when hashtable should
+        be multiplied automatically
+        """
+        length = len(self.__hashtable)
+        if length - self.__hashtable.count(None) > length * self.__load_factor:
```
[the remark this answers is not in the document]
ORENYT: PM'ed you, what do you actually think it is better to change? I use length to calculate full length of dict, but __len__ calculates non-None elements

app/main.py
```diff
@@ -1,2 +1,123 @@
+import dataclasses
+from typing import Hashable
+
+
+@dataclasses.dataclass
+class Node:
+    key: object
```
Kroll410, line 7: Hashable type

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+    key_hash: int
+    value: object
```
Kroll410, line 9: Any type

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+
+
 class Dictionary:
-    pass
+    __load_factor = 2 / 3
+
+    def __init__(self) -> None:
+        self.__hashtable = [None] * 8
+
+    def __setitem__(self, key: Hashable, value: object) -> None:
```
Kroll410, line 18: `object` is too broad and high-abstract type. Use typings

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+                self.index += 1
+            return (
+                self.__hashtable[self.index].key,
+                self.__hashtable[self.index].value,
+            )
+        except IndexError:
+            raise StopIteration
+
+    def pop(self, key: Hashable) -> object:
+        item = self.__getitem__(key)
```
Kroll410, line 85: if not found, default value should be returned
image

app/main.py
```diff
@@ -1,2 +1,123 @@
...
+
+    def update(self, second_dict: "Dictionary") -> None:
+        for node in second_dict.__dict__["_Dictionary__hashtable"]:
+            if node:
+                self.__setitem__(node.key, node.value)
+            self.__setitem__(node.key, node.value)
+
+    def __iter__(self) -> list:
+        self.index = 0
+        return self.__hashtable
```
Kroll410, line 70: `dict` when being iterated sends us the keys, if not values or items methods were specified

Kroll410 (changes requested):

app/main.py
ORENYT: Understood, accepted.

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+        for i in range(preferred_cell, length):
+            if self.__hashtable[i] and self.__hashtable[i].key == key:
+                return self.__hashtable[i].value
+        for i in range(0, preferred_cell):
+            if self.__hashtable[i] and self.__hashtable[i].key == key:
+                return self.__hashtable[i].value
+        raise KeyError("Key not found")
+
+    def __len__(self) -> int:
+        return len([node for node in self.__hashtable if node is not None])
```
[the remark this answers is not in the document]
ORENYT: Oh, sorry, fixed

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+        preferred_cell = hashed_key % length
+        for i in range(preferred_cell, length):
+            if self.__hashtable[i] and self.__hashtable[i].key == key:
+                self.__hashtable[i] = None
+                return
+        for i in range(0, preferred_cell):
+            if self.__hashtable[i] and self.__hashtable[i].key == key:
+                self.__hashtable[i] = None
+                return
+        raise KeyError("Key not found")
```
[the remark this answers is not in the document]
ORENYT: I swear I fixed it yesterday, omg. Something wrong with github i think

app/main.py
```diff
@@ -1,2 +1,127 @@
...
+                self.__hashtable[i] = None
+                self.length -= 1
+                return
+        raise KeyError("Key not found")
+
+    def get(self, key: object, default: Any = -1) -> object:
+        try:
+            return self.__getitem__(key)
+        except KeyError:
+            return default
```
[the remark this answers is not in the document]
ORENYT: Changed to None

app/main.py
```diff
@@ -1,2 +1,127 @@
...
+
+    def update(self, second_dict: "Dictionary") -> None:
```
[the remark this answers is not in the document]
ORENYT: Deleted this method

app/main.py
```diff
@@ -1,2 +1,127 @@
...
+        for node in second_dict.__dict__["_Dictionary__hashtable"]:
+            if node:
+                self.__setitem__(node.key, node.value)
+            self.__setitem__(node.key, node.value)
```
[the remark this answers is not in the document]
ORENYT: this method is removed as an optional

ORENYT (commented):
Accepted all changes

PR #443, opened 2023-08-30
completed py-dictionary

Barsh4ec:


Barsh4ec:
> Only minimal required functionality :(

I will try to do extra functionality at the weekends

PR #444, opened 2023-08-30
Solution

CaCtUs1k:


KristianYe (approved):

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        self._capacity = 8
+        self._load_factor = 2 / 3
+        self._size = 0
+        self._storage = [None] * self._capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        idx = hash(key) % self._capacity
+        while True:
+            if self._storage[idx] is None:
+                self._storage[idx] = [key, hash(key), value]
```
Kroll410, line 15: Should be tuple

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+                self._size += 1
+                break
+            if (
+                key == self._storage[idx][0]
+                and hash(key) == self._storage[idx][1]
```
Kroll410, line 20: Multiple `hash` calls

app/main.py
```diff
@@ -1,2 +1,81 @@
...
-    pass
+    def __init__(self) -> None:
+        self._capacity = 8
+        self._load_factor = 2 / 3
+        self._size = 0
+        self._storage = [None] * self._capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        idx = hash(key) % self._capacity
+        while True:
```
Kroll410, line 13: You can put it in a separate method and call in getitem setitem

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        else:
+            raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self._size
+
+    def __iter__(self) -> None:
+        for item in self._storage:
+            if item is not None:
+                yield item[2]
```
Kroll410, line 56: It's a Dict, why we return only value? This shoudl return jkeys if values() or items() not specified

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+    def clear(self) -> None:
+        self._storage = [None] * self._capacity
+
+    def get(self, key: Any, default: Any = None) -> None:
+        try:
+            return self.__getitem__(key)
+        except KeyError:
+            return default
+
+    def pop(self, key: Any) -> None:
```
Kroll410, line 76: Wrong logic
image

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+                and key == self._storage[idx][0]
+                and hash(key) == self._storage[idx][1]
+            ):
+                return self._storage[idx][2]
+            if idx == end_idx:
+                raise KeyError
+            idx = (idx + 1) % self._capacity
+
+    def __delitem__(self, key: Any) -> None:
+        for idx in range(self._capacity):
```
Kroll410, line 43: Why do we iterating through capacity range? we can simply iterate storage

Kroll410 (changes requested):

app/main.py
CaCtUs1k: I'm changing the value of one of the storage cells. But to do this without range, you need to introduce an additional variable,what is not very desirable, no?

PR #445, opened 2023-08-30
Solution

denys-source:


CaCtUs1k (approved):

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
+
 class Dictionary:
-    pass
+    LOAD_FACTOR: float = 2 / 3
+
+    def __init__(self) -> None:
+        self._capacity: int = 8
+        self._length: int = 0
+        self._Item = namedtuple("_Item", ["key", "hash", "value"])
```
Kroll410, line 12: That  works

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+        self._items = [None] * self._capacity
+        for item in old_items:
+            if item is not None:
+                self._length -= 1
+                self.__setitem__(item.key, item.value)
+
+    def _find_index(self, key: Hashable) -> int:
+        element_index: int = hash(key) % self._capacity
+        while self._items[element_index] is not None and (
+            self._items[element_index].hash != hash(key)
```
Kroll410, line 28: u're breaking transitivity logic law here, simply compare hashes

Kroll410 (changes requested):

app/main.py
denys-source: if you are talking about removing this part `or self._items[element_index].key != key`, it will break the tests

PR #468, opened 2023-09-15
Solution

jinjo0222988:


app/main.py
```diff
@@ -1,2 +1,94 @@
...
+    def __init__(self, elements: list[tuple[Any, Any]] = None) -> None:
+        self.indices = [-1 for _ in range(8)]
+        self.length = 0
+        self.entries = []
+        if elements is not None:
+            self._assign_entries(elements)
+
+    def _assign_entries(self, elements: list[tuple]) -> None:
+        for key, value in elements:
+            self.__setitem__(key, value)
```
ArsenPidhoretskyi, line 14:
```suggestion
            self[key] = value
```

app/main.py
```diff
@@ -1,2 +1,94 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, elements: list[tuple[Any, Any]] = None) -> None:
+        self.indices = [-1 for _ in range(8)]
```
ArsenPidhoretskyi, line 6: This is redundant

ArsenPidhoretskyi: You can just create entries with len 8

ArsenPidhoretskyi (changes requested):
indices is redundant part of realization

app/main.py
```diff
@@ -1,2 +1,90 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, elements: list[tuple[Any, Any]] = None) -> None:
+        self.entries = [-1 for _ in range(8)]
```
ArsenPidhoretskyi, line 6:
```suggestion
        self.entries = [None for _ in range(8)]
```

app/main.py
```diff
@@ -1,2 +1,90 @@
...
+    def _object_equal(
+            input_key: Any,
+            input_hash: int,
+            entries_key: Any,
+            entries_hash: int
+    ) -> bool:
+        return input_key == entries_key and entries_hash == input_hash
+
+    def _resize_entries(self) -> None:
+        new_entries = [-1 for _ in range(len(self.entries) * 2)]
```
ArsenPidhoretskyi, line 30:
```suggestion
        new_entries = [None for _ in range(len(self.entries) * 2)]
```

ArsenPidhoretskyi (changes requested):
-1 is magic number. Better to initialize with None

PR #469, opened 2023-09-16
Solution

Artemh59:


app/main.py
```diff
@@ -1,2 +1,50 @@
...
+        for index, (existing_key, existing_value) in enumerate(slot):
+            if existing_key == key:
+                slot[index] = (key, value)
+                break
+        else:
+            slot.append((key, value))
+            self.size += 1
+
+        if self.size >= self.capacity * 2:
+            self.resize()
```
viktoria-rybenchuk, line 26: You should optimize the __setitem__ method by checking the load factor and resizing the hash table if necessary before inserting the key-value pair

app/main.py
```diff
@@ -1,2 +1,50 @@
...
+
+    def __getitem__(self, key: Any) -> Any:
+        hash_index = self.get_hash(key)
+        slot = self.table[hash_index]
+
+        for existing_key, existing_value in slot:
+            if existing_key == key:
+                return existing_value
+
+        raise KeyError(f"Key '{key}' not found")
```
viktoria-rybenchuk, line 36: You can enhance the efficiency of your code by implementing an approach that directly calculates the index based on the key and retrieves the value from the hash table. This approach avoids unnecessary iterations and checks, making it a more efficient way to access key-value pairs

viktoria-rybenchuk (changes requested):

app/main.py
viktoria-rybenchuk: you didn't fix it. Find values associated with keys by directly using the key's hash code for quick identification and retrieval, without the need for time-consuming for loops

viktoria-rybenchuk (changes requested):

Artemh59:
I really don't know how do this without dict

Polyakiv-Andrey (changes requested):

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+                slot[index] = (hash_index, key, value)
+                key_found = True
+                break
+            index += 1
+
+        if not key_found:
+            slot.append((hash_index, key, value))
+            self._size += 1
+
+    def __getitem__(self, key: Hashable) -> Hashable:
```
viktoria-rybenchuk, line 34: You should not specify a return type of Hashable in this context because the value associated with a key in a dictionary or hash table may not necessarily be of type `Hashable`

app/main.py
```diff
@@ -1,2 +1,59 @@
...
-    pass
+    def __init__(self, capacity: int = 8) -> None:
+        self.capacity = capacity
+        self._size = 0
+        self.table = [[] for _ in range(self.capacity)]
+
+    def get_hash(self, key: Hashable) -> int:
+        return hash(key) % self.capacity
+
+    def __setitem__(self, key: Hashable, value: Hashable) -> None:
```
viktoria-rybenchuk, line 13: value associated with a key might not necessarily be hashable

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,2 +1,59 @@
...
-    pass
+    def __init__(self, capacity: int = 8) -> None:
+        self.capacity = capacity
+        self._size = 0
+        self.table = [[] for _ in range(self.capacity)]
+
+    def get_hash(self, key: Any) -> int:
+        return hash(key) % self.capacity
+
+    def __setitem__(self, key: Any, value: Hashable) -> None:
```
viktoria-rybenchuk, line 13:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

viktoria-rybenchuk: In dictionary key must indeed be hashable, but there are no specific restrictions on the hashability of the values. Values in a dictionary can be of any type, whether hashable or not

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #472, opened 2023-09-20
Solution

AllaKuksa:


app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.size = 0
+        self.table = [[] for _ in range(self.capacity)]
+
+    def _hash(self, key: Any) -> Any:
```
vsmutok, line 11:
```suggestion
    def _hash(self, key: Hashable) -> int:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
+        for sells in self.table:
+            for key, value in sells:
+                new_index = self._hash(key) % new_capacity
+                new_table[new_index].append((key, value))
+
+        self.capacity = new_capacity
+        self.table = new_table
+
+    def __setitem__(self, key: Any, value: Any) -> Any:
```
vsmutok, line 27:
```suggestion
    def __setitem__(self, key: Any, value: Hashable) -> Any:
```

vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.size = 0
+        self.table = [[] for _ in range(self.capacity)]
+
+    def _hash(self, key: Hashable) -> int:
```
ArsenPidhoretskyi, line 11: This is redundant

ArsenPidhoretskyi (changes requested):

app/main.py
AllaKuksa: What is exactly redundant?

vasylhnatiuk: this function, use only hash(key)

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #477, opened 2023-10-02
Solution

olgierrd:


app/main.py
```diff
@@ -1,2 +1,49 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8, load_factor: float = 0.66) -> None:
```
viktoria-rybenchuk, line 5:
```suggestion
    def __init__(self, capacity: int = 8, load_factor: float =  2 / 3) -> None:
```

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+                entry[1] = value
+                return
+        self.table[index].append([key, value])
+        self.size += 1
+        if self.size > self.capacity * self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: Any) -> Any:
+        index = self._hash(key)
+        if self.table[index] is not None:
```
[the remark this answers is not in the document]
olgierrd: But if I do not, I get TypeError: 'NoneType' object is not iterable

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+        for entry in self.table[index]:
+            if entry[0] == key:
+                entry[1] = value
+                return
+        self.table[index].append([key, value])
+        self.size += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._hash(key)
+        if self.table[index] is not None:
```
pavlejviki, line 26:
```suggestion
        if self.table[index]:
```

pavlejviki (changes requested):

PR #478, opened 2023-10-03
created class Dictionary

antonnech2309:


Polyakiv-Andrey (changes requested):
The dictionary key can be not only immutable types. They must be hashable. An instance of the class is mutable, but hashable and can be a dictionary key.

Polyakiv-Andrey (approved):
Good job!

PR #480, opened 2023-10-06
'Solution'

usKos:


app/main.py
```diff
@@ -1,2 +1,56 @@
...
+                if self.hash_table[hash_key][0] == key:
+                    self.hash_table[hash_key][1] = value
+                    break
+                else:
+                    hash_key += 1
+            else:
+                self.hash_table[hash_key] = [key, value]
+                break
+
+        if self.__len__() > self.capacity * self.load_factor:
```
ArsenPidhoretskyi, line 24:
```suggestion
        if len(self) > self.capacity * self.load_factor:
```

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+            self.resize()
+
+    def __getitem__(self, key: Any) -> Any:
+        hash_key = hash(key) % self.capacity
+        while True:
+            if self.hash_table[hash_key] is not None:
+                if self.hash_table[hash_key][0] == key:
+                    return self.hash_table[hash_key][1]
+                else:
+                    hash_key += 1
```
ArsenPidhoretskyi, line 34: Do not forgot to divide for
```suggestion
                    hash_key = (hash_key + 1) % self.capacity
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.hash_table: list = [None] * capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        hash_key = hash(key) % self.capacity
+
+        for i in range(len(self.hash_table)):
+            if self.hash_table[i] is not None:
+                if self.hash_table[i][0] == key:
+                    self.hash_table[i][1] = value
+                    return
```
ArsenPidhoretskyi, line 18: Why you use forloop, dict works with hash

app/main.py
```diff
@@ -1,2 +1,54 @@
...
+            hash_key = (hash_key + 1) % self.capacity
+
+        if len(self) > self.capacity * self.load_factor:
+            self.resize()
+
+    def __getitem__(self, key: Any) -> Any:
+        for i in range(len(self.hash_table)):
+            if self.hash_table[i] is not None:
+                if self.hash_table[i][0] == key:
+                    return self.hash_table[i][1]
```
ArsenPidhoretskyi, line 33: The same use while loop

ArsenPidhoretskyi (changes requested):

YukihiroSM (changes requested):
Fix tests, please.

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.hash_table: list = [None] * capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        hash_key = hash(key) % self.capacity
+
+        while True:
+            if self.hash_table[hash_key] is not None:
+                if self.hash_table[hash_key][0] == key:
```
ArsenPidhoretskyi, line 16: Also you need check that hash the same

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+            self.resize()
+
+    def __getitem__(self, key: Any) -> Any:
+        hash_key = hash(key) % self.capacity
+
+        for i in range(len(self.hash_table)):
+            h = (hash_key + i) % self.capacity
+            if self.hash_table[h] is not None:
+                if self.hash_table[h][0] == key:
+                    return self.hash_table[h][1]
```
ArsenPidhoretskyi, line 35: It must work as while loop

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+            hash_key = (hash_key + 1) % self.capacity
+
+        if len(self) > self.capacity * self.load_factor:
+            self.resize()
+
+    def __getitem__(self, key: Any) -> Any:
+        hash_key = hash(key) % self.capacity
+        i = 0
+        while i < len(self):
+            for i in range(len(self.hash_table)):
```
ArsenPidhoretskyi, line 32: Why you need loop in this case????

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+                h = (hash_key + i) % self.capacity
+                if self.hash_table[h] is not None:
+                    if self.hash_table[h][0] == key:
+                        return self.hash_table[h][1]
```
ArsenPidhoretskyi, line 36: Please just use while loop instead of forloop

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi:
Also naming like capacity2 is bad

usKos:
why does it have to be a while loop?
while loop doesn't pass the test,
 for loop is taking tests ('Solution_branch7_1')

usKos:
this for loop is pass tests

app/main.py
EdAlekseiev: Not fixed

app/main.py
EdAlekseiev: Not fixed

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #483, opened 2023-10-10
implemented py-dictionary task

HlovaOleh:


app/main.py
```diff
@@ -1,2 +1,71 @@
...
+        else:
+            self.size += 1
+            node = Node(key, hash(key), value)
+            self.table[index] = node
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self.get_index(key)
+        if len(self.table) > index:
+            if self.table[index]:
+                if self.table[index].key == key:
```
EdAlekseiev, line 43: Check hash value too

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+                _is_matching_key(self.table[index], key)):
+            index = (index + 1) % self.capacity
+        return index
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size >= self.capacity * self.LOAD_FACTOR:
+            self.resize()
+        index = self.get_index(key)
+        while self.capacity <= index:
+            self.resize()
```
ArsenPidhoretskyi, line 37: Why you need resize in this case?????

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+                self[element.key] = element.value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.get_index(key)
+        if self.table[index] and _is_matching_key(self.table[index], key):
+            self.table[index] = None
+            self.size -= 1
+
+    def get(self, item: Hashable) -> Any:
+        return self[item]
```
ArsenPidhoretskyi, line 69: Really? In this case get will raise error

ArsenPidhoretskyi (changes requested):

app/main.py
HlovaOleh: If the index exceeds the current capacity, it means that the table is full or close to full, and inserting a new element may result in poor performance or inefficient storage. So, it triggers a resize operation to increase the capacity of the table, allowing it to accommodate more elements efficiently.

app/main.py
HlovaOleh: I agree

HlovaOleh: I'll delete it

app/main.py
YukihiroSM: You are resizing in line 34, so there will not be case, when Index would be out of range. Also, it's better to perform a resize after you added element, mean in the end of this function, because size increases there.

YukihiroSM (approved):
Looks good to me! Thanks for fixing things!

PR #484, opened 2023-10-11
'Solution'

struhanets:


vasylhnatiuk (changes requested):
fix tests

vasylhnatiuk (changes requested):
The resizing logic in the resize method is incorrect. You shouldn't recreate the entire table with an increased capacity. Instead, you should create a new list with the increased capacity and rehash all the existing elements into the new table.

struhanets:
"Here's a corrected version of the resize method:"
where?)

app/main.py
```diff
@@ -1,2 +1,93 @@
...
+        self.resize(key, value)
+        if self.size == 0:
+            self.table[index_key].append((key, value, hash_key))
+            self.size += 1
+        else:
+            for elem in range(len(self.table[index_key])):
+                if self.table[index_key][elem][0] == key:
+                    self.table[index_key][elem][1] = (key, value, hash_key)
+
+            self.table[index_key + 1].append((key, value, hash_key))
```
ArsenPidhoretskyi, line 39: In this way we always do replace and add new item

app/main.py
```diff
@@ -1,2 +1,93 @@
...
+            self.size += 1
+
+    # прописуємо розширення таблиці при заповненні на 2/3
+    def resize(self, key: Any, value: Any) -> None:
+        if self.size >= self.overload:
```
ArsenPidhoretskyi, line 44: Resize must do resize, not check condition for it

app/main.py
```diff
@@ -1,2 +1,93 @@
...
+            self.capasity *= 2
+
+            self.table = [[] for _ in range(self.capasity)]
+            self.__setitem__(key, value)
```
ArsenPidhoretskyi, line 48:
```suggestion
            self[key] = value
```

app/main.py
```diff
@@ -1,2 +1,93 @@
...
+        else:
+            for elem in range(len(self.table[index_key])):
+                if self.table[index_key][elem][0] == key:
+                    self.table[index_key][elem][1] = (key, value, hash_key)
+
+            self.table[index_key + 1].append((key, value, hash_key))
+            self.size += 1
+
+    # прописуємо розширення таблиці при заповненні на 2/3
+    def resize(self, key: Any, value: Any) -> None:
```
ArsenPidhoretskyi, line 43: Why resize accept key and value???

app/main.py
```diff
@@ -1,2 +1,93 @@
...
+            self.capasity *= 2
+
+            self.table = [[] for _ in range(self.capasity)]
+            self.__setitem__(key, value)
+
+    # повернення значення при запиті по ключу, якщо такого ключа немає, викидає помилку
+    def __getitem__(self, item: Union[int, float, str, object]) -> Any:
+        index_key = self.hash_func(item)
+        for key, value, hash_key in self.table[index_key]:
+            if key == item:
```
ArsenPidhoretskyi, line 54: also check that hash the same

ArsenPidhoretskyi (changes requested):

app/main.py
struhanets: I do't understand
Where I need to add it?

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+            self.resize()
+
+    # прописуємо розширення таблиці при заповненні на 2/3
+    def resize(self) -> None:
+        self.capasity *= 2
+        new_table = [] * self.capasity
+        self.size = 0
+        for item in new_table:
+            if item and item is not None:
+                self.__setitem__(item[0], item[1])
```
ArsenPidhoretskyi, line 54:
```suggestion
                self[item[0]] = item[1]
```

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+
+        if self.size >= self.overload:
+            self.resize()
+
+    # прописуємо розширення таблиці при заповненні на 2/3
+    def resize(self) -> None:
+        self.capasity *= 2
+        new_table = [] * self.capasity
+        self.size = 0
+        for item in new_table:
```
ArsenPidhoretskyi, line 52: new_table always empty

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
fix tests

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+        if self.table[index] is None:
+            raise KeyError(f"Key '{key}' not found in the dictionary")
+        return self.table[index][1]
+
+    # довжина таблиці із заповненими елементами
+    def __len__(self) -> int:
+        return self.size
+
+    # видалення елемента по ключу
+    def __delitem__(self, key: Any) -> None:
```
vasylhnatiuk, line 68:
```suggestion
    def __delitem__(self, key: Hashable) -> None:
```

app/main.py
```diff
@@ -1,2 +1,78 @@
...
+        if self.size >= self.capasity * self.load_factor:
+            self.resize()  # зміна розміру після перегрузки таблиці
+        index = self.find_index(key)
+        if self.table[index] is None:
+            self.size += 1
+        self.table[index] = [key, value, hash(key)]
+
+    def find_index(self, key: Hashable) -> int:
+        index_key = self.hash_func(key)
+        while self.table[index_key] is not None and \
```
vasylhnatiuk, line 38: try to use () instead of \

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #485, opened 2023-10-12
solution

Shum430:


app/main.py
```diff
@@ -1,2 +1,76 @@
...
+            if node.key == key:
+                return node.value
+            node = node.next
+
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self.size
+
+    def resize(self) -> None:
```
Polyakiv-Andrey, line 62: The resize method is inside a class, and will only be used for that class, and can also be used in child classes, use protected-level encapsulation for it.

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #486, opened 2023-10-13
'Solution'

tiron-vadym:


app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        if not self.table[index]:
+            self.table[index] = []
+        for node in self.table[index]:
+            if node[0] == key:
+                node[2] = value
+                return
+        self.table[index].append([key, self._hash(key), value])
+        self.size += 1
+        if self.size > self.capacity * self.load_factor:
+            self._resize()
```
viktoria-rybenchuk, line 26: You should perform this checking before you set the item

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+    def __getitem__(self, key: Any) -> Any:
```
viktoria-rybenchuk, line 28: `key` cannot be Any. Fix in all places

app/main.py
```diff
@@ -1,2 +1,52 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+            initial_capacity: int = 16,
+            load_factor: float = 0.75
```
viktoria-rybenchuk, line 8:
```suggestion
            load_factor: float = 2/3
```

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,2 +1,52 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+            initial_capacity: int = 16,
```
vsmutok, line 7: initial_capacity must  be 8

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        for bucket in self.table:
+            if bucket:
+                for node in bucket:
+                    index = self._hash(node[0])
+                    if not new_table[index]:
+                        new_table[index] = []
+                    new_table[index].append(node)
+        self.table = new_table
+
+    def __setitem__(self, key: str, value: Any) -> None:
```
vsmutok, line 27:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

vsmutok: fix everywhere

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        index = self._hash(key)
+        if not self.table[index]:
+            self.table[index] = []
```
vsmutok, line 30: why you need it?

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        for node in self.table[index]:
+            if node[0] == key:
+                node[2] = value
+                return
+        self.table[index].append([key, self._hash(key), value])
```
vsmutok, line 35:
```suggestion
        self.table[index] = [key, self._hash(key), value]
```

vsmutok (changes requested):

app/main.py
tiron-vadym: Where can I import Hashable from?

app/main.py
tiron-vadym: For proper collision handling

app/main.py
tiron-vadym: If I replace this part, in the case of a collision, it will overwrite the existing entry in self.table[index] instead of adding a new one.

app/main.py
viktoria-rybenchuk: not fixed.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self.size
+
+    def _hash(self, key: Hashable) -> int:
+        return hash(key) % self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._hash(key)
```
pavlejviki, line 43: why did you completely remove the call of the  _resize method?

pavlejviki (changes requested):

app/main.py
viktoria-rybenchuk: By checking and resizing the table before adding a new item, you ensure there's enough space and slots available for the new key-value pair.

viktoria-rybenchuk (changes requested):
fix according to comment

tiron-vadym:
> fix according to comment

I don't understand exactly how I should change the part, can you explain more specifically, please

Polyakiv-Andrey (approved):
Good job!

PR #488, opened 2023-10-15
Solution

sachavskyi:


app/main.py
```diff
@@ -1,2 +1,84 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+        for _ in range(self.capacity):
+            if key == self.table[index][0]:
+                self.table[index] = [None]
+                self.size -= 1
+                break
+            else:
```
Dimosphen1, line 65: There is no need for the `else` statement, code will not execute after `break` anyway, so remove it.

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+        for key_value in old_table:
+            if key_value != [None]:
+                key_hash = hash(key_value[0])
+                index = key_hash % self.capacity
+                while True:
+                    if key_value[0] == self.table[index][0] or \
+                            (self.table[index] == [None]):
+                        self.table[index] = [key_value[0], key_value[1]]
+                        break
+                    else:
```
Dimosphen1, line 51: There is no need for the `else` statement, code will not execute after break anyway, so remove it.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #497, opened 2023-10-23
Solution

mykolamateichuk:


RabenkoYevhenii (approved):
great!

fsocie7y (changes requested):
use more descriptive name instead of k, v, h etc.)

PR #501, opened 2023-10-24
Solution

Katherine-Greg:


Lyutillis (approved):
In __getitem__ you could get items by index = hash(key) % capacity

app/main.py
```diff
@@ -1,2 +1,45 @@
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.load_factor = self.capacity * 0.67
+        self.table = [None] * self.capacity
+
+    def __setitem__(self, key: int, value: int) -> None:
```
fsocie7y, line 8: The correct type annotation for key is Hashable.

app/main.py
```diff
@@ -1,2 +1,45 @@
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.load_factor = self.capacity * 0.67
```
fsocie7y, line 5: I think it would be more intuitive to multiply by 2 / 3, since this is a more straightforward representation of what you want to achieve here.

app/main.py
```diff
@@ -1,2 +1,45 @@
...
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        self.load_factor = self.capacity * 0.67
+        current_table = self.table
+        self.table = [None] * self.capacity
+        self.size = 0
+        for item in current_table:
+            if item:
+                self.__setitem__(item[0], item[2])
```
fsocie7y, line 42: here u can use [ ]

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Good job!

PR #503, opened 2023-10-25
Solution

vshvanska:


app/main.py
```diff
@@ -1,2 +1,50 @@
...
+        self.hash_table[index] = (key, value)
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * 2)
+        for item in old_hash_table:
+            if item is not None:
+                key, value = item
+                self.__setitem__(key, value)
```
fsocie7y, line 44: u can use [ ] here

fsocie7y (approved):
approved) but fix it in according to my advice ^_~

app/main.py
vshvanska: Done🫡

PR #512, opened 2023-11-05
Solution

kravetb:


app/main.py
```diff
@@ -1,2 +1,84 @@
...
+
+
 class Dictionary:
-    pass
+    count_value_in_dict = 0
+
+    def __init__(self, capacity: int = 8) -> None:
+        self.capacity = capacity
+        self.key = [[] for _ in range(self.capacity)]
+        self.value = [[] for _ in range(self.capacity)]
```
ArsenPidhoretskyi, line 10: It is so bad to store this into two different lists

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+        self.count_of_jump = []
```
ArsenPidhoretskyi, line 11: ????

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+                value,
+                self.key,
+                self.value,
+                self.count_of_jump
+            )
+            self.count_value_in_dict += 1
+            if self.count_value_in_dict >= round(self.capacity * 2 / 3):
+                self.revenge()
+
+    def revenge(self) -> None:
```
ArsenPidhoretskyi, line 55: Revenge??? Do you know what this word means?

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+        self.value = [[] for _ in range(self.capacity)]
+        self.count_of_jump = []
+
+    def add_value_in_dict(
+        self,
+        key: Any,
+        value: Any,
+        ls_key: list,
+        ls_value: list,
+        ls_buf_count: list
```
ArsenPidhoretskyi, line 19: This variables totally redundant

ArsenPidhoretskyi (changes requested):
Logic is so hard......

YukihiroSM (changes requested):
Looks better now! Also, instead of `def overwrite` you should use `def resize`, because it's not overwriting anything.

YukihiroSM (approved):

PR #515, opened 2023-11-10
Solution

frezworx:


app/main.py
```diff
@@ -1,2 +1,49 @@
...
+
+    def __init__(
+            self,
+            initial_size: int = 8,
+            load_factor: float = 0.7
+    ) -> None:
+        self._initial_size = initial_size
+        self._load_factor = load_factor
+        self._array_buckets = [[] for _ in range(initial_size)]
+        # self._length =
```
pavlejviki, line 14: remove redundant commented code

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+
+    def _get_hash_code(self, key: int | float | str) -> int:
```
pavlejviki, line 16: wrong type annotation for key. it has to be Hashable. fix in all places

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+        return hash(key) % self._initial_size
+
+    def add_data(self, key: int | float | str, value: Any) -> None:
+        hash_code = self._get_hash_code(key)
+        bucket = self._array_buckets[hash_code]
+        for i, (key_, _) in enumerate(bucket):
```
pavlejviki, line 22: i is not a descriptive variable name

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+        for key_, value_ in bucket:
+            if key_ == key:
+                return value_
+        raise KeyError(f"Key {key} not found")
+
+    def __setitem__(self, key: int | float | str, value: Any) -> None:
+        self.add_data(key, value)
+
+    def __getitem__(self, key: int | float | str) -> Any:
+        return self.get_data(key)
```
pavlejviki, line 40: no need to create addiitional method get_data. you can implement its logic straight in def __getitem__.the same with def __setitem__

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+    ) -> None:
+        self._initial_size = initial_size
+        self._load_factor = load_factor
+        self._array_buckets = [[] for _ in range(initial_size)]
+        # self._length =
+
+    def _get_hash_code(self, key: int | float | str) -> int:
+        return hash(key) % self._initial_size
+
+    def add_data(self, key: int | float | str, value: Any) -> None:
```
pavlejviki, line 19: you need to resize the dictionary when its size is more than its load_factor.

pavlejviki (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #516, opened 2023-11-11
solution

Andrey1104:


app/main.py
```diff
@@ -1,2 +1,64 @@
...
+            self.size -= 1
+
+    def resize(self) -> None:
+        temp_table = self.table
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+        for element in temp_table:
+            if element:
+                self.__setitem__(element[1], element[2])
```
vsmutok, line 50:
```suggestion
                self[element[1]] = element[2]
```

vsmutok (changes requested):

app/main.py
Andrey1104: can you comment in more detail?

Andrey1104: can you comment in more detail?

vsmutok: `self[element[1]] = element[2]` is a more straightforward and concise way to set a value for a specific key. It directly accesses the dictionary element and is self-explanatory.

vsmutok (approved):
Good Job 👍

app/main.py
Andrey1104: but i can set value with setter, is it a mistake to not accept the request?

Andrey1104:
I try to use different methods so as not to forget them

Andrey1104:
соу мени нью інформейшн ін май хед, со ші експлоуд вері суун)))))

PR #517, opened 2023-11-12
Solution

Artem-user-19:


app/main.py
```diff
@@ -1,2 +1,86 @@
...
+            load_factor: float = 0.75
+    ) -> None:
+        self.capacity: int = initial_capacity
+        self.load_factor: float = load_factor
+        self.size: int = 0
+        self.table: list[Optional[Node]] = [None] * initial_capacity
+
+    def __setitem__(
+            self,
+            key: Any,
```
Dimosphen1, line 25: `key` should be not any, but hashable. Indicate this in the type annotation in all places.

app/main.py
```diff
@@ -1,2 +1,86 @@
+from typing import Optional, Any
+
+
+class Node:
+    def __init__(self, key: Any, hash_value: int, value: Any) -> None:
```
Dimosphen1, line 5: Place all arguments on a new line to match the code style for the `__init__` of the `Dictionary` class.

Dimosphen1 (changes requested):
Several minor changes were requested.

app/main.py
```diff
@@ -1,2 +1,91 @@
+from typing import Optional, Any, Hashable
+
+
+class Node:
+    def __init__(
+            self,
+            key: Any,
```
Dimosphen1, line 7: Indicate that the `key` should be `Hashable` in all places in the type annotation where you expect to obtain it, not only in the `__setitem__`.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #523, opened 2023-11-19
Solution

AnyoneClown:


app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
+    def _get_index(self, key: Hashable) -> int:
+        index = hash(key) % self.capacity
+        while (self.hash_table[index] is not None
+               and self.hash_table[index][0] != key):
+            index = (index + 1) % self.capacity
+        return index
+
+    def _resize(self) -> None:
+        self.capacity *= 3
```
pavlejviki, line 42: the more conventional approach is doubling the capacity instead of tripling it

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        self.hash_table = [None] * self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._get_index(key)
+        if self.hash_table[index] is None:
+            self.size += 1
+            self.hash_table[index] = (key, hash(key), value)
+        elif self.hash_table[index] is not None:
+            self.hash_table[index] = (key, hash(key), value)
+        if self.size > self.load_factor * self.capacity:
```
pavlejviki, line 22: you should perform this check at the beginning of the method

you ensure that the hash table is always within its load factor limits before inserting a new item

pavlejviki (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #524, opened 2023-11-21
Solution

Molodoirockstar:


app/main.py
```diff
@@ -1,2 +1,60 @@
...
+        self.capacity = 8
+        self.hash_table = [None] * self.capacity
+        self.load_factor = 2 / 3
+        self.length = 0
+
+    def __setitem__(self, key: any, value: any) -> None:
+        if self.length == round(self.capacity * self.load_factor):
+            self.resize_table()
+        index = hash(key) % self.capacity
+        for i in range(self.capacity):
```
ArsenPidhoretskyi, line 12: Better to use while loop in this case

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+            elif not self.hash_table[index]:
+                self.hash_table[index] = (key, value)
+                self.length += 1
+                break
+            else:
+                index += 1
+
+    def __getitem__(self, key: any) -> None:
+        index = hash(key) % self.capacity
+        for i in range(self.capacity):
```
ArsenPidhoretskyi, line 27: Better to use while loop in this case

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+            if index > self.capacity - 1:
+                index -= self.capacity
```
ArsenPidhoretskyi, line 29: index = (index + 1) % self.capacity

ArsenPidhoretskyi (approved):
Well done!

PR #525, opened 2023-11-23
solution

oleksashcherbakov:


Polyakiv-Andrey (changes requested):
Tests not passed!

app/main.py
```diff
@@ -1,2 +1,60 @@
+from typing import Hashable, Any
+
+
 class Dictionary:
-    pass
+    def __init__(self,
+                 capacity: int = 8,
+                 size: int = 0,
+                 load_factor: float = 2 / 3
+                 ) -> None:
```
Dimosphen1, line 9: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

app/main.py
```diff
@@ -1,2 +1,60 @@
...
+        return self.storage[index][2]
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __repr__(self) -> str:
+        return (f"capacity: {self.capacity} "
+                f"size: {self.size} "
+                f"storage: {self.storage}"
+                )
```
Dimosphen1, line 60: In case there are arguments on the same line with an opening parenthesis, your closing parenthesis should have arguments on the same line as well. Please, resolve this issue

app/main.py
```diff
@@ -1,2 +1,60 @@
...
-    pass
+    def __init__(self,
+                 capacity: int = 8,
+                 size: int = 0,
+                 load_factor: float = 2 / 3
+                 ) -> None:
+        self.capacity = capacity
+        self.size = size
+        self.load_factor = load_factor
+        self.storage = [(None, None, None)] * self.capacity
```
Dimosphen1, line 13: There should be no need to store tuple of three `None` as a default state, only `None` instead of tuple should be sufficient. Make according change here and in the `resize` method as well

Dimosphen1 (changes requested):

PR #528, opened 2023-12-02
'Solution'

SerhiiDomik:


app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        if self.size / self.capacity >= self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: int) -> int:
+        index = self._hash(key)
+        while self.hash_table[index] is not None:
+            if self.hash_table[index].key == key:
+                return self.hash_table[index].value
+            index = (index + 1) % self.capacity
+        raise KeyError
```
ArsenPidhoretskyi, line 44:
```suggestion
        raise KeyError(key)
```
Provide key

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+                return
+            index = (index + 1) % self.capacity
+
+        self.hash_table[index] = Node(key, value)
+        self.size += 1
+
+        if self.size / self.capacity >= self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: int) -> int:
```
ArsenPidhoretskyi, line 38: key is Hashable, not int

app/main.py
```diff
@@ -1,2 +1,65 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: int, value: Any) -> None:
+        self.key = key
+        self.value = value
```
ArsenPidhoretskyi, line 7: also store hash and then compare it also

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,2 +1,69 @@
...
+        self.hash_table = new_table
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._hash(key)
+        while self.hash_table[index].key != key:
+            if (self.hash_table[index].hash == hash(key)
+                    and self.hash_table[index].key == key):
+                self.hash_table[index] = None
+            index = (index + 1) % self.capacity
+        self.size -= 1
```
ArsenPidhoretskyi, line 69: Here is the logic must be another, check this test code(problem in case deletion):
```
d = Dictionary()
d[1] = 4
d[9] = 15
del d[1]
print(d[9])
```

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Great  👍

PR #530, opened 2023-12-04
Solution

aleksandra-prognimak:


app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.hash_table = [None] * self.capacity
+        self.length = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
LiudmylaKulinchenko, line 11: key is not Any, it's hashable

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        index = self.get_index(key)
+
+        if not self.hash_table[index]:
+            if self.length >= int(self.capacity * self.load_factor):
+                self.resize()
+                index = self.get_index(key)
+            self.length += 1
```
LiudmylaKulinchenko, line 18: You should add 1 to self.length before you make a check and resizing

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+            raise KeyError
+
+        return self.hash_table[index][2]
+
+    def __len__(self) -> int:
+        return self.length
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        self.length = 0
```
LiudmylaKulinchenko, line 35: Is there need to set length to zero? You add 1 to length every time you set a new value, so it will be correct

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good!

PR #540, opened 2023-12-17
'solution'

boris-devs:


app/main.py
```diff
@@ -1,2 +1,116 @@
...
+        return hash(key) % self.capacity
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __getitem__(self, item: any) -> any:
+        index = self._find_hash(item)
+        current = self.storage[index]
+        while current:
+            if current.key == item:
```
EdAlekseiev, line 26: You should compare hash of key too

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+                    return
+                if current.next is None:
+                    break
+                current = current.next
+
+            current.next = new_node
+        self.length += 1
+
+        if self.length >= self.capacity * self.load_factor:
+            self._resize()
```
EdAlekseiev, line 52: Make resize in the start of function

EdAlekseiev (changes requested):

app/main.py
EdAlekseiev: Not fixed

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #547, opened 2024-01-13
not solved py-dictionary

zdmytryszyn:


vsmutok (changes requested):
tests failed!

zdmytryszyn:
> tests failed!

I know, the task was pushed in order to be checked by one of the mentors

app/main.py
```diff
@@ -1,2 +1,143 @@
...
+        index = key_hash % self.capacity
+        result = self.find_index_of_item(index=index, item=key)
+        if result is not None:
+            return self.hash_table[result].value
+        return value
+
+    def pop(
+            self,
+            keyname: int | float | bool | str | tuple | Hashable,
+    ) -> Any:
```
pavlejviki, line 134: the standard  pop method  accepts a default value which is returned  in case the specified key is not present. the same as for get method. please implement this logic

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,2 +1,144 @@
...
+        key_hash = hash(keyname)
+        index = key_hash % self.capacity
+        result = self.find_index_of_item(index=index, item=keyname)
+        if result is not None:
+            return_value = self.hash_table[result].value
+            self.hash_table[result] = None
+            self.size_of_dict -= 1
+            return return_value
+        return defaultvalue
```
pavlejviki, line 144: If the key is not found, return the default if given; otherwise, raise a KeyError.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #555, opened 2024-01-22
Solution

govorunenko:


app/main.py
```diff
@@ -1,2 +1,119 @@
+from typing import Any, Hashable
+
+
+class Node:
+    def __init__(self, key: Hashable, value: Any,
+                 next_node: "Node" = None) -> None:
```
EdAlekseiev, line 6: If argument can be None, you must show it in type hint
```suggestion
                 next_node: "Node" | None = None) -> None:
```

app/main.py
```diff
@@ -1,2 +1,119 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._hash(key)
+        if self.table[index] is None:
+            self.table[index] = Node(key, value)
+
+        else:
+            current = self.table[index]
+            while current is not None:
+                if current.key == key:
```
EdAlekseiev, line 28: Compare hash of key too

app/main.py
```diff
@@ -1,2 +1,119 @@
...
+        self.size += 1
+
+        if self.size / self.capacity > self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._hash(key)
+        current = self.table[index]
+        while current is not None:
+            if current.key == key:
```
EdAlekseiev, line 44: Compare hash of key too

EdAlekseiev (changes requested):

app/main.py
govorunenko: Done.

app/main.py
govorunenko: Done:
if current.key == key and hash(current.key) == hash(key):

NikOneZ1 (approved):
Good Job!

PR #557, opened 2024-01-23
Solution

kabust:


app/main.py
```diff
@@ -1,2 +1,76 @@
...
+
+    def __delitem__(self, key: Any) -> None:
+        index = hash(key) % self.capacity
+        try:
+            while self.bucket[index] and key != self.bucket[index].key:
+                index = (index + 1) % self.capacity
+            self.bucket[index] = None
+            self.length -= 1
+        except AttributeError:
+            raise KeyError(f"Dictionary doesn't have a {key} key")
```
NEkropo1, line 34: Suppose you have a collision, where multiple keys hash to the same index and are placed in subsequent indices due to linear probing.
If one of these keys (not the last one in the chain) is deleted and its slot is set to None, then a lookup for a key further down the chain will fail. This happens because the search for the key will stop at the None slot, incorrectly assuming that the key is not in the dictionary.

I know it's complicated for basic implementation, still you can think about it if you have time.

NEkropo1 (approved):
Good job!

PR #558, opened 2024-01-23
Solution

ArturPoltser:


kabust (approved):

NEkropo1 (approved):
Overall great, but think about your overall implementations with None. What if we delete some values, which are parts of the collision chain? We won't get all following objects, and it's not an expected behavior.

PR #563, opened 2024-01-25
Solution

msymonovych:


app/main.py
```diff
@@ -1,2 +1,100 @@
...
+
+        step = 0
+        while step < self.length:
+            if element and element[0] == key:
+                return element[2]
+            index = (index + 1) % self.capacity
+            step += 1
+            element = self.hash_table[index]
+
+        raise KeyError(key)
```
aLEKS-e3, line 58: why do you need **step** here?
why don't you use a `while element:` statement and remove all the steps?

aLEKS-e3 (changes requested):

aLEKS-e3 (approved):

PR #568, opened 2024-01-31
'Solution'

yura-push:


app/main.py
```diff
@@ -1,2 +1,35 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.hashmap: list = [[] for _ in range(self.capacity)]
+        self.length = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        hash_key = hash(key) % self.capacity
```
Dimosphen1, line 11: There is no need to obtain a modulo from the key hash since after resizing all elements would be reordered in the hashmap and it introduces additional load to the elements adding. You should avoid reordering by storing the hash directly.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+                )
+
+        self.hashmap = new_hashmap
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        hash_key = hash(key)
+        bucket = self.hashmap[self.hash_key_index(key)]
+
+        for index, (stored_key, stored_value, stored_hash) \
+                in enumerate(bucket):
```
vsmutok, line 31: better to use `()` then `\`

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+
+    def resize(self) -> None:
+        self.capacity *= 2
+        new_hashmap = [[] for _ in range(self.capacity)]
+
+        for bucket in self.hashmap:
+            for key, value, hash_key in bucket:
+                new_hashmap[self.hash_key_index(key)].append(
+                    (key, value, hash_key)
+                )
```
vsmutok, line 22: use here `__setitem__`

vsmutok (changes requested):

PR #570, opened 2024-02-01
solution v1

motivated2:


app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        raise KeyError
+
+    def __len__(self) -> int:
+        return self.length
+
+    def resize(self) -> None:
+        new_hash_table: list = [None] * (self.initial_capacity * 2)
+        for i in range(self.initial_capacity):
+            if self.hash_table[i]:
+                new_hash_table[i] = self.hash_table[i]
```
mdubyna, line 31: after changing the size of the list, the indexes on which the elements should appear may differ from those that were previously. For rxample key == 9, hash(9) =9 , capacity == 8, index should be 9 % 8 == 1, but if capacity  == 16  index should be 9 % 16 == 9

mdubyna (approved):
As for me, the code does not quite correspond to the purpose of the task

PR #571, opened 2024-02-01
Solution

mdubyna:


oksankakot (approved):
Good job! 
I like you put static method here.
Maybe better to put backspaces between condition blocks to make it easier to read.

PR #590, opened 2024-02-04
Solution

Nick-r2c:


app/main.py
```diff
@@ -1,2 +1,48 @@
...
+            index %= self._capacity
+        return index
+
+    def _resize(self) -> None:
+        table_copy = self._table
+        self._capacity *= 2
+        self._table = [None] * self._capacity
+        self._size = 0
+        for item in table_copy:
+            if item is not None:
```
isikach, line 47: You can use here shorter form like "if not..."

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+
+        self._table[index] = (key, key_hash, value)
+
+    def __len__(self) -> int:
+        return self._size
+
+    def get_index(self, key: Hashable, key_hash: int) -> int:
+        index = key_hash % self._capacity
+
+        while self._table[index] is not None and self._table[index][0] != key:
```
isikach, line 36: "if something" equal to  "if something is not None"

isikach (approved):

PR #595, opened 2024-02-09
Done

DenZatolokin:


app/main.py
```diff
@@ -1,2 +1,110 @@
...
+            current = current.next
+        raise KeyError(key)
+
+    def get(self, key: Any) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return "The key does not exist"
+
+    def pop(self, key: Any) -> Any:
```
NikOneZ1, line 94: This method should take two parameters as a dict

NikOneZ1 (changes requested):

app/main.py
DenZatolokin: > This method should take two parameters as a dict

What should the second parameter be? The "pop" method does a "remove and return"

EdAlekseiev (approved):
Good job

PR #619, opened 2024-03-05
Solution

IhorDmytriv:


app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        while cell_in_table[hash_index] is not None:
+            if cell_in_table[hash_index][0] == key_item:
+                break
+            if hash_index < self._initial_capacity - 1:
+                hash_index += 1
+            else:
+                hash_index = 0
+        return hash_index
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
vasylhnatiuk, line 26:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+            raise TypeError(f"Unsupported type for {iterable}. "
+                            f"Expected Dictionary, list, or tuple.")
+
+    def __iter__(self) -> Any:
+        for item in self._hash_table:
+            if item is not None:
+                yield item[0]
+
+    def __next__(self) -> Any:
+        raise StopIteration
```
YukihiroSM, line 100: Why do you need this method to be like this? 
It will not allow to iterate over anything. It will just immediately raise exception. 

Also, there is a question about `__iter__` method. Should this be an iterator, or generator? For now it looks, like a generator with method next for some reason.

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+    def pop(self, key: Hashable, default: Any = None) -> list:
+        try:
+            result = self.__getitem__(key)
+        except KeyError as error:
+            if not default:
+                raise error
+            return default
+        else:
+            self.__delitem__(key)
+            return result
```
YukihiroSM, line 80: Probably the return typehint here shouldn't be list.

app/main.py
YukihiroSM: Also, the return typehint for next is wrong. In current implementation it should be None

YukihiroSM (changes requested):

EdAlekseiev (approved):

PR #627, opened 2024-03-12
'Solution'

donbas777:


app/main.py
```diff
@@ -1,2 +1,74 @@
...
+            if index == start_index:
+                raise KeyError("Key was not found")
+
+    def get(self, key: Hashable, default: Optional[Any] = None) -> Any:
+        try:
+            return self.hash_table[self.get_index(key)][2]
+        except KeyError:
+            if default is not None:
+                return default
+            raise KeyError(key)
```
Dimosphen1, line 48: `get` should not raise a `KeyError` in any case, it should return `None` when the `default` is not provided since `None` is a default option.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #636, opened 2024-03-26
Solution

Boyok1337:


nicksetrakov (approved):

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+    def __delitem__(self, key: Hashable) -> None:
+        index_key = self.hash_func(key)
+        if self.table[index_key] and self.table[index_key][0] == key:
+            del self.table[index_key]
+            self.size -= 1
+
+    def clear(self) -> None:
+        self.table = [None]
+        self.size = 0
+        self.capacity = 0
```
YuriiKindrat, line 40: it seems to me, that this method should just remove all ements from the dict
so will it work corectly with capacity 0 and table with only 1 elem ?

YuriiKindrat (changes requested):

app/main.py
Boyok1337: Not in all cases, so i change it

app/main.py
```diff
@@ -1,2 +1,58 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        index_key = self.hash_func(key)
+        if self.table[index_key] and self.table[index_key][0] == key:
+            del self.table[index_key]
+            self.size -= 1
+
+    def clear(self) -> None:
+        self.table = []
+        self.size = 0
```
YuriiKindrat, line 39: it seems to me, that u should do the same thing as in init method
make capacity equal to 8 and then make table with 8 spaces in it )

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: u didn't fix that, i mean clear method, that should start from capacity = 8, and table [None] * capacity )

app/main.py
```diff
@@ -1,2 +1,57 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self,
+                 load_factor: float = 2 / 3,
+                 size: int = 0) -> None:
```
YuriiKindrat, line 7: are u sure, that load factor & size of dict, should be configurable ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #640, opened 2024-03-30
Solution

Viesich:


app/main.py
```diff
@@ -1,2 +1,45 @@
...
+            self,
+            initial_capacity: int = 8,
+            load_factor: float = 2 / 3
+    ) -> None:
+        self._size = 0
+        self._capacity = initial_capacity
+        self._load_factor = load_factor
+        self._table = [[] for _ in range(self._capacity)]
+
+    def __setitem__(self, key: str, value: int) -> None:
```
Dimosphen1, line 12: Indicate that `key` should be `Hashable` in the type annotation and that value could be `Any`, not only `int`. Apply this change in other places as well

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+            self,
+            initial_capacity: int = 8,
+            load_factor: float = 2 / 3
+    ) -> None:
+        self._size = 0
+        self._capacity = initial_capacity
+        self._load_factor = load_factor
+        self._table = [[] for _ in range(self._capacity)]
+
+    def __setitem__(self, key: Hashable, value: any) -> None:
```
Dimosphen1, line 15: Since `Any` from `typing` is not deprecated, you could use it in the type annotation instead of `any`

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #641, opened 2024-04-01
Solution

Bujhm99:


Boyok1337 (approved):
Not bad!

app/main.py
```diff
@@ -1,2 +1,53 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, initial_capacity: int = 8,
+                 load_factor: int | float = 2 / 3,
+                 resize: int = 2) -> None:
```
YuriiKindrat, line 7: are u sure that this params should be configurable ?)

YuriiKindrat (changes requested):

app/main.py
Bujhm99: As I understand exercise, this is the main essence of the task - to create an object that is different from an already defined dictionary in terms of basic parameters. So, when adjusting the capacity, we can create an instance of the class that suits us, for example, with a minimum number of cases of hash table re-indexing. Or, by adjusting the load factor, achieve fewer cases of collision processing.

YuriiKindrat (approved):
GJ

PR #647, opened 2024-04-06
Solution

vitalinamal:


app/main.py
```diff
@@ -1,2 +1,121 @@
+from typing import Any
+
+
+class HashTable:
```
Oleksl888, line 4: This is not really a hash table - it is a node that is inside of  a hash table which is `self.table` inside your dicitionary. Also it says in task requirements 

Also this is basically a part of a Dictionary class and will not be used outside of it so it would make sense to declare it inside of class dictionary

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+        self.value = value
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.table = [None] * self.capacity
+        self.load_factor = 0.75
```
Oleksl888, line 15: if we follow dictionary original implementation its load_factor is 2/3 - can be more accurately represented by `Fraction(2, 3)` this class can be used with mathematical operations from the box

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.table}"
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
Oleksl888, line 20: you can use `Hashable` type for key

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+        if not hash(key):
```
Oleksl888, line 21: hash() from unhashable type will trigger a TypeError by itself

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+            raise TypeError("Dictionary key must be immutable")
+
+        index = self.get_index(key)
+
+        while self.table[index] is not None:
+            if self.table[index].key == key:
+                self.table[index].value = value
+                return
+            index = (index + 1) % self.capacity
```
Oleksl888, line 30: this basically is almost the same logic as in your `find_index` method. Can you refactor to follow DRY principle?

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+    def rehash(self) -> None:
+        old_table = self.table[:]
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
+
+    def __iter__(self) -> list:
```
Oleksl888, line 91: __iter__ method is expected to return an iterator but you return a list which is an iterable

Oleksl888 (changes requested):

Oleksl888 (changes requested):
please change

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+        return f"{self.table}"
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.get_index(key)
+        index = self.find_index(index, key)
+        if index[1]:
+            self.table[index[0]].value = value
+            return
+
+        self.table[index[0]] = Dictionary.Node(key, value, hash(key))
```
Oleksl888, line 28: you can access Node with `self`

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+            items = other
+
+        try:
+            for key, value in items:
+                self.__setitem__(key, value)
+        except TypeError:
+            raise ValueError("The input must be a dictionary "
+                             "or an iterable of key-value pairs")
+
+    def pop(self, key: Hashable) -> Any:
```
Oleksl888, line 107: pop method should have am optional default param that will be returned if key is not found

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+
+    def resize(self) -> None:
+        old_table = self.table[:]
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 80: you already defined __setitem__ method which will allow to use synt sugar

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+
+    def __iter__(self) -> list:
+        return (node.key for node in self.table if node is not None)
+
+    def update(self, other: Any) -> None:
```
Oleksl888, line 94: other can be a dict or an iterable. Use more appropriate typing here

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+                self[node.key] = node.value
+
+    def rehash(self) -> None:
+        old_table = self.table[:]
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 89: you could use self[node.key] = node.value here and below -  since you already defined __Setitem__ behaviour

Oleksl888 (approved):
Well done!

Oleksl888 (approved):
Well done!

app/main.py
vitalinamal: Changed in "resize" but did not see in "rehash"..... Done )

PR #648, opened 2024-04-07
Solution v0.1

panicua:


panicua:
Just deny it, I forgot that I don't need to create pull request, I also want to implement other unnescessary methods for the dict + need to ask something on QnA about it.

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+            key_hash_position = hash(key) % self._hash_size
+        except TypeError as e:
+            raise TypeError(e)
+        else:
+            return key_hash_position
+
+    def _hash_size_check_increase(self) -> None:
+        """Supporting func to check and increase hash_table size when needed"""
+
+        if self.__len__() == int(self._hash_size * self._resize_factor):
```
Oleksl888, line 91: you already defined __len__ method therefor you can use syntactic sugar here

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+    def clear(self) -> None:
+        self._hash_table = [None] * 8
+        self._hash_size = 8
+        self._length = 0
+
+    def _find_key_position(self, key: Hashable) -> int:
+        try:
+            key_hash_position = hash(key) % self._hash_size
+        except TypeError as e:
+            raise TypeError(e)
```
Oleksl888, line 84: there not much sense in catching this error and reraising without adding any extra logic to it.  If you call hash() on unacceptable type TypeError will be raised anyway

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        self._hash_size_check_increase()
+        key_hash_position = self._find_key_position(key)
+
+        while self._hash_table[key_hash_position]:
+            if self._hash_table[key_hash_position][0] == key:
+                self._hash_table[key_hash_position][1] = value
+                return print(f"value with key '{key}' were rewritten")
+            key_hash_position = (key_hash_position + 1) % self._hash_size
```
Oleksl888, line 20: You repeat this block of code on line 31, 50 so move it to a function to make your code DRY

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+
+        self._hash_table[key_hash_position] = [key, value]
```
Oleksl888, line 22: > Also, don't forget to store (key, hash, value) as a node in the hash table. A hash table may be stored as a simple list of nodes.

As per task requirements you also need to store hash as well. Would make sense to make a Node data structure and add it to your hash table

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+                self._hash_table[key_hash_position] = None
+                print(f"removed: {key} from position {key_hash_position}")
+                self._length -= 1
+                return
+            key_hash_position = (key_hash_position + 1) % self._hash_size
+
+        raise KeyError("Key is not found in the Dictionary")
+
+    def pop(self, key: Hashable) -> Any:
+        result = self.__getitem__(key)
```
Oleksl888, line 61: use syntactic sugar here since you have these methods implemented:
e.g. del self[key]

app/main.py
```diff
@@ -1,2 +1,108 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self._length: int = 0
+        self._hash_table: list = [None] * 8
+        self._hash_size: int = 8
+        self._resize_factor: float | int = 2 / 3
```
Oleksl888, line 9: better use Fraction type here

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+        for item in self._hash_table:
+            if item:
+                item_key_hash_position = hash(item[0]) % self._hash_size
+                # check so they don't overwrite each-other
+                while temp_hash_table[item_key_hash_position]:
+                    item_key_hash_position = ((item_key_hash_position + 1)
+                                              % self._hash_size)
+                temp_hash_table[item_key_hash_position] = item
+        self._hash_table = temp_hash_table
+        print(f"Dictionary size were increased "
```
Oleksl888, line 106: imagine you work with dicts and they resize like every second, you will be flooded by these messages :)
Also a lot of prints actually will consume your system resources so its ok to use them while you test things but remove it when you open a PR.

Oleksl888 (changes requested):

panicua:
I've spend 5~ more hours today just to fix/update/refactor. 
---First hard part were to refactor it correctly so I can use one method ```_while_key_hash_occupied``` for getitem/setitem magic methods.
---Second hard part were to understand why my del magic method doesn't work as it should be, appeared that problem where hiden deep in how algorythm of "linear probing" work.
 
I'll try to explain if you want to read:
Imagine we have hash table and the spot that we want to write in is already taken, so we need to jump to the next spot (in our case to right, so index + 1) untill we find an empty spot and write your Node there. But imagine now you have to delete the item, but it's real spot not where it's hash says, it's actually to the right of it on some kind of length so we need to run from that entry point where our hash says and run to the right the whole list untill we make sure that no Node with this key exists (it's the worst case scenario O(n)). But in average it's still O(1).

Oleksl888:
>But imagine now you have to delete the item, but it's real spot not where it's hash says, it's actually to the right of it on some kind of length so we need to run from that entry point where our hash says and run to the right the whole list untill we make sure that no Node with this key exists (it's the worst case scenario O(n)). But in average it's still O(1).

I didn't look into your code thoroughly yet but I think logic of traversing the hash table in case of collision should be the same no matter if you insert, retrieve or delete the item. The only place where you might get some wrong values is when you have to resize table. But again there should be no problem if you resize table, recalculate all new positions for all exisitng values and then add new node. On the opposite if you delete an item you might need to resize after you removed the element.

panicua:
> > But imagine now you have to delete the item, but it's real spot not where it's hash says, it's actually to the right of it on some kind of length so we need to run from that entry point where our hash says and run to the right the whole list untill we make sure that no Node with this key exists (it's the worst case scenario O(n)). But in average it's still O(1).
> 
> I didn't look into your code thoroughly yet but I think logic of traversing the hash table in case of collision should be the same no matter if you insert, retrieve or delete the item. The only place where you might get some wrong values is when you have to resize table. But again there should be no problem if you resize table, recalculate all new positions for all exisitng values and then add new node. On the opposite if you delete an item you might need to resize after you removed the element.


I did not implement resizign after del btw, It's gonna get too deep for this task I belive. It's not that I'm lazy, but.. :D

GPT:
However, Python does have a mechanism to resize dictionaries to a smaller size if a large number of items are deleted, but this happens gradually and through a process called "rehashing." Rehashing occurs when the dictionary's size drops below a certain threshold (usually around 1/3 of the current size), and it needs to be shrunk to save memory.

So, to answer your question, deleting all 6 items from the dictionary won't immediately reduce its hash table size. Python will keep the hash table at its current size for efficiency, and it may eventually resize it smaller through rehashing if many more items are deleted.

app/main.py
```diff
@@ -1,2 +1,119 @@
...
+
+            key_hash_position += 1
+            index_len += 1
+
+        raise KeyError("Key is not found in the Dictionary")
+
+    def pop(self, key: Hashable) -> Any:
+        result = self[key]
+        del self[key]
+        return result
```
YuriiKindrat, line 60: thats how pop method should work )
image

YuriiKindrat (changes requested):

app/main.py
panicua: > thats how pop method should work ) image

you mean that I need to return None in my case instead of key error? so like ```result = self.get(key)``` instead of ```result = self[key]```?

app/main.py
```diff
@@ -1,2 +1,132 @@
...
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        """
+        func returns/removes value/item if key exists,
+        otherwise return/raise default/KeyError
+        """
+        try:
+            result = self[key]
+            del self[key]
+        except KeyError:
+            return default
```
Oleksl888, line 69: Please check your logic - you should raise KeyError in case if no key and no default value

app/node.py
```diff
@@ -0,0 +1,9 @@
+from typing import Any, Hashable
+from dataclasses import dataclass
+
+
+@dataclass
+class Node:
```
Oleksl888, line 6: I'd say Node must be defined inside of a Dictionary class - because it is part of its implementation

tests/test_main.py
```diff
@@ -6,18 +6,18 @@
...
-#     dictionary = Dictionary()
-#     for key, value in items:
-#         dictionary[key] = value
-#     for key, value in items:
-#         assert dictionary[key] == value
-#     assert len(dictionary) == len(items)
-#     for key, value in items:
-#         del dictionary[key]
-#     print(len(dictionary))
+@pytest.mark.timeout(5)
```
Oleksl888, line 9: Please revert changes to tests

app/main.py
```diff
@@ -1,2 +1,132 @@
...
+        return "{" + ", \n".join(result) + "}"
+
+    def __delitem__(self, key: Hashable) -> None:
+        """
+        func to delete item by key if key exists, otherwise raise KeyError
+        """
+        key_hash_position = self._find_key_position(key)
+        index_len = 0
+
+        while index_len <= self._hash_size:
```
Oleksl888, line 48: So why not modify  `_while_key_hash_occupied` method and reuse it here to avoid repetition?

Oleksl888 (changes requested):

app/main.py
panicua: Because if we use this method with "del", we cannot garantee that we will look for an item(Node) even if entry point ```table[key_hash_position] == None``` it will immediately exit the loop and give wrong return. It can happen due to collision if we have 3 Nodes with starting point at table[1] for example, they will be written in table[1], table[2], table[3] due to collision and imagine we need to delete first Node at table[1], after that we'll have None, table[2], table[3], now we need to delete table[2] or table[3] but their keys will have same address to table as table[1] that is already deleted, they look at the condition in ```_while_key_hash_occupied``` and see that their entry point (table[1]) is None so they don't even check other cells in table to make sure that needed item even exists.
NVIDIA_Share_MLxlV766aa

and this condition guarantees us that even if we don't guess exact entry point for our table we'll look at every cells in table at least once (in worst case scenario, if we don't have this key in our table or if every key had same entry point in our table and you had collision every time)
NVIDIA_Share_50azuCeClz

panicua:
changes:
Changed methods: ```__setitem__```, ```__getitem__```, ```__delitem__```.
Added methods: ```_move_table_right``` - to jump + 1 right in hash table
```_find_key_in_table``` - an universal method that used in ```__getitem__```, ```__delitem__``` has differences only in return, not in condition.
Deleted method: ```_while_key_hash_occupied```.

**Comment by panicua** [comment]
changes:
Changed methods: ```__setitem__```, ```__getitem__```, ```__delitem__```.
Added methods: ```_move_table_right``` - to jump + 1 right in hash table
```_find_key_in_table``` - an universal method that used in ```__getitem__```, ```__delitem__``` has differences only in return, not in condition.
Deleted method: ```_while_key_hash_occupied```.

File: app/main.py [hunk 13]
```diff
@@ -1,2 +1,110 @@
...
+
+                temp_hash_table[item_key_hash] = item
+        self._hash_table = temp_hash_table
+
+    def _find_key_in_table(self, action: str, key: Hashable) -> Any | None:
+        hash_key = self._find_key_position(key)
+        table = self._hash_table
+        for _ in range(self._hash_size):
+            if table[hash_key] is not None and table[hash_key].key == key:
+                if action == "get":
```
Oleksl888, line 103: So here you could just return `hash_key` (btw `index` would be more appropriate name) to the function that called this method and then if its __getitem__ that you calling it from you do `table[hash_key].value`
if it's __delitem__ you do 
```
 table[hash_key] = None
                    self._length -= 1

File: app/main.py [hunk 14]
```diff
@@ -1,2 +1,110 @@
+from typing import Any, Hashable
+from fractions import Fraction
+from app.node import Node
```
Oleksl888, line 3: blank line here :)

Oleksl888 (approved):
Good job!

PR #649, opened 2024-04-08
Develop

DmytroHlazyrin:


app/main.py
```diff
@@ -1,2 +1,89 @@
...
+
+    def clear(self) -> None:
+        self._table = [None] * self._capacity
+        self._size = 0
+
+    def get(self, key: Hashable) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return None
```
YuriiKindrat, line 77: what about default argument for get method ?

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+
+    def pop(self, key: Hashable) -> Any:
+        try:
+            value: Any = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return None
```
YuriiKindrat, line 85: pop method also could get 2 args )

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+
+    def update(self, other_dict: dict) -> None:
+        for key, value in other_dict.items():
+            self[key] = value
```
YuriiKindrat, line 89: it seems to me, that this method also should work with ur custom Dict class

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            value: Any = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
```
YuriiKindrat, line 90: image
this is how pop should work )

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #650, opened 2024-04-08
custom dict

mikhailyemets:


app/main.py
```diff
@@ -1,2 +1,66 @@
...
+                if new_table[new_index] is None:
+                    new_table[new_index] = Node(node.key, node.value)
+                else:
+                    current = new_table[new_index]
+                    while current.next:
+                        current = current.next
+                    current.next = Node(node.key, node.value)
+                node = node.next
+        self.capacity = new_capacity
+        self.table = new_table
```
YuriiKindrat, line 66: could u use setitem method here ?

YuriiKindrat (changes requested):

PR #651, opened 2024-04-09
Solution

Liliyakhu:


app/main.py
```diff
@@ -1,2 +1,44 @@
...
+        self.hash_table: list = [None] * 8
+        self.capacity = 8
+        self.load_factor = 2 / 3
+        self.resize = 2
+        self.threshold = self.capacity * self.load_factor
+
+    def __len__(self) -> int:
+        return self.length
+
+    def get_ind(self, key: Hashable) -> object:
```
pavlopro, line 16: It would be more readable if you name it something like 'get_index'

pavlopro, line 16: Correct return annotation

pavlopro (changes requested):

EdAlekseiev (approved):

PR #655, opened 2024-04-11
solution

ArtemLeo:


app/main.py
```diff
@@ -1,2 +1,94 @@
...
+
+        if not self._table[index]:
+            raise KeyError(f"Key: {key} is not found!")
+
+        return self._table[index].value
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.get_index(key)
+        if not self._table[index]:
+            if self._size + 1 >= self._capacity * 2 / 3:
```
Dimosphen1, line 31: It is better to define load factor in the `__init__` to be able to change it from there, not from the code itself

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        try:
+            value: Any = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
+
+    def update(self, other_dict: dict | Dictionary) -> None:
+        for key, value in other_dict.items():
```
Dimosphen1, line 93: `items()` method seems to be not implemented, and because of that you cannot pass `Dictionary` class instance here

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #657, opened 2024-04-13
solution

nrjss67:


app/main.py
```diff
@@ -1,2 +1,113 @@
...
-    pass
+
+    def __init__(self) -> None:
+        self.hash_table = [[]] * 8
+        self.current_bucket = 8
+        self.current_size = 0
+
+    def __setitem__(self,
+                    key: int | str | tuple | float | Point,
+                    value: Any) -> None:
```
vsmutok, line 15:
```suggestion
    def __setitem__(self, key: Hashable,  value: Any) -> None:
```

vsmutok: please fix everywhere

vsmutok (changes requested):

app/main.py
```diff
@@ -1,2 +1,125 @@
...
+    def clear(self) -> None:
+        self.current_bucket = 8
+        self.current_size = 0
+        self.hash_table = [[]] * self.current_bucket
+
+    def __resize(self) -> None:
+        if inspect.stack()[1].function == "__setitem__":
+            self.current_bucket *= 2
+        if inspect.stack()[1].function == "__delitem__":
+            self.current_bucket //= 2
```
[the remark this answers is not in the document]
nrjss67: fixed
but I thought if when deleting an element the current size is less than 2/3 it is necessary to reduce the size of the dictionary, so as not to occupy extra memory

vasylhnatiuk (approved):

PR #658, opened 2024-04-14
solution

alonasorochynska:


app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        self.load_factor = load_factor
+        self.size = 0
+        self.table = [None] * capacity
+
+    def __getitem__(self, key):
+        index = self._get_index(key)
+        if self.table[index] is not None:
+            return self.table[index].value
+        else:
+            raise KeyError(f"Key '{key}' not found.")
```
YuriiKindrat, line 20: what if elements will have a collision ?

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        return hash(key) % self.capacity
+
+    def _resize(self):
+        self.capacity *= 2
+        new_table = [None] * self.capacity
+        for node in self.table:
+            if node is not None:
+                index = node.hash_value % self.capacity
+                new_table[index] = node
+        self.table = new_table
```
YuriiKindrat, line 47: u could use setitem method here

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        index = self._get_index(key)
+        hash_value = hash(key)
+        node = Node(key, hash_value, value)
+        if self.table[index] is None:
+            self.table[index] = node
+            self.size += 1
+            if self.size > self.capacity * self.load_factor:
+                self._resize()
+        else:
+            raise KeyError(f"Key '{key}' already exists.")
```
YuriiKindrat, line 32: u need to write a mechanism to handle collisions

YuriiKindrat (changes requested):
all tests should pass, fix ur flake8 errors

PR #665, opened 2024-04-20
Solution

mgoryn:


app/main.py
```diff
@@ -1,2 +1,56 @@
+class Node:
+    def __init__(self, key: int | str, value: int | str) -> None:
+        self.key = key
+        self.value = value
+        self.next = None
```
vsmutok, line 5: why you need it?

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+    def __init__(self, capacity: int = 10, load_factor: float = 0.7) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.table = [None] * capacity
+
+    def _hash(self, key: int | str) -> int:
+        return hash(key) % self.capacity
+
+    def __setitem__(self, key: int | str, value: int | str) -> None:
```
vsmutok, line 18:
```suggestion
    def __setitem__(self, key: Hashable) -> None:
```

vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
mgoryn: The Node class is utilized in the Dictionary class to represent individual elements in a hash table, each containing a key-value pair.

mgoryn: using a separate Node class allows for more flexibility and control over the data structure. It enables us to store additional information and implement additional functionalities such as extending chains or changing the order of elements. Additionally, utilizing our own Node class provides greater flexibility and control over the data structure.

app/main.py
```diff
@@ -1,2 +1,61 @@
...
+    ) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.table = [None] * capacity
+
+    def _hash(self, key: Hashable) -> int:
+        return hash(key) % self.capacity
+
+    def __setitem__(self, key: Hashable, value: Hashable) -> None:
```
vsmutok, line 23:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

app/main.py
```diff
@@ -1,2 +1,61 @@
...
+    def __init__(self, key: Hashable, value: Hashable) -> None:
+        self.key = key
+        self.value = value
+        self.next = None
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self, capacity: Hashable = 10, load_factor: Hashable = 0.7
```
vsmutok, line 13:
```suggestion
            self, capacity: int = 10, load_factor: float = 0.7
```

app/main.py
```diff
@@ -1,2 +1,61 @@
+from typing import Hashable, Any
+
+
+class Node:
+    def __init__(self, key: Hashable, value: Hashable) -> None:
```
vsmutok, line 5:
```suggestion
    def __init__(self, key: Hashable, value: Any) -> None:
```

vsmutok (changes requested):

app/main.py
mgoryn: Thank you)

EdAlekseiev (approved):

PR #667, opened 2024-05-01
Solution

LitvinOleksandr:


app/main.py
```diff
@@ -1,2 +1,72 @@
...
+                return
+            position = (position + 1) % self.capacity
+            if position > self.capacity:
+                break
+        raise KeyError(key)
+
+    def clear(self) -> None:
+        self.hash_table = [None] * self.capacity
+
+    def get(self, key: Hashable) -> Any:
```
vasylhnatiuk, line 66: get() should return None if there is not such key

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #670, opened 2024-05-05
Solution

Nyxan:


app/main.py
```diff
@@ -1,2 +1,54 @@
...
+        self.size = 0
+        self.table = [None] * capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index: int = self._hash(key) % self.capacity
+        if not self.table[index]:
+            self.table[index] = [(key, value)]
+            self.size += 1
+        else:
+            pairs = self.table[index]
```
MNDonut, line 17: it's better to make if self.table[index] else (case when not), so that it's more readable

MNDonut (approved):

MNDonut (approved):

PR #678, opened 2024-05-18
'Solution'

MarianaTykha:


app/main.py
```diff
@@ -1,2 +1,123 @@
...
+    def _resize(self) -> None:
+        old_table = copy.deepcopy(self.table)
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for member in old_table:
+            try:
+                self.__setitem__(
+                    key=member.key,
```
vsmutok, line 68: bad code style,  use `black` to improve it

vsmutok (changes requested):

vsmutok (changes requested):
Tests Failed!

app/main.py
```diff
@@ -1,2 +1,91 @@
+from typing import Hashable
+
+import copy
+
+
+class DictionaryMember:
+    def __init__(self, key: Hashable, value: any, index: int = None) -> None:
```
vsmutok, line 7:
```suggestion
    def __init__(self, key: Hashable, value: Any, index: int = None) -> None:
```
fix everywhere

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+            self._resize()
+
+    def __getitem__(self, key: int | str | tuple | float) -> any:
+        index = self._index(key)
+        current = self.table[index]
+
+        if current is None:
+            raise KeyError
+        else:
+            return current.value
```
vsmutok, line 77:
```suggestion
        if current is None:
            raise KeyError
        return current.value
```
no need to use `else` in this case

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+            self.size += 1
+            print("Data wrote")
+        else:
+            self.table[key_index].value = value
+            print("Data rewrote")
+
+        if self.size / self.capacity > self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: int | str | tuple | float) -> any:
```
vsmutok, line 70:
```suggestion
    def __getitem__(self, key: Hashable) -> Any:
```
fix everywhere

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+        old_table = copy.deepcopy(self.table)
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for member in old_table:
+            try:
+                self.__setitem__(key=member.key, value=member.value)
+            except AttributeError:
+                pass
```
vsmutok, line 54: is bad practice to use `pass` in this case 
how the user should understand if something goes wrong?

vsmutok (changes requested):

app/main.py
vasylhnatiuk: please use print() and write a message

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #682, opened 2024-05-26
Hell Yeah solution

RuslanLevchuk:


app/main.py
```diff
@@ -1,2 +1,97 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.storage_length = 8
+        self.storage = [None] * self.storage_length
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
vasylhnatiuk, line 9:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

vasylhnatiuk: key can be only Hashable

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+
+    def __iter__(self) -> tuple:
+        for item in self.storage:
+            if item:
+                yield (item[0], item[2])
+
+    def clear(self) -> None:
+        self.__init__()
+
+    def get(self, item: Any) -> Any:
```
vasylhnatiuk, line 74: get() should get additional default argument, if key doesn't exist

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+
+    def resize(self) -> None:
+        temp_storage = [i for i in self.storage if i]
+        self.storage_length *= 2
+        self.storage = [None] * self.storage_length
+        for i in temp_storage:
+            self.__setitem__(i[0], i[2])
+        del temp_storage
+
+    def __getitem__(self, item: Any) -> Any:
```
vasylhnatiuk, line 35:
```suggestion
    def __getitem__(self, key: Hashable) -> Any:
```

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+        key_hash = hash(item)
+        for item_ in self.storage:
```
vasylhnatiuk, line 37: you should find by key, not by item

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+        self.storage = [None] * self.storage_length
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        new_item = (key, hash(key), value)
+        index = self.get_index(new_item)
+        self.storage[index] = new_item
+        if len([i for i in self.storage if i]) >= 2 * self.storage_length / 3:
+            self.resize()
+
+    def get_index(self, new_item: tuple) -> int:
```
vasylhnatiuk, line 16: you can find index - index = hash(key) % self.capacity

app/main.py
vasylhnatiuk: you should get item by key

vasylhnatiuk (changes requested):

app/main.py
RuslanLevchuk: i just variable, that stores key value named as "item". The logic of code - looking by key

app/main.py
RuslanLevchuk: > you can find index - index = hash(key) % self.capacity

Yes, that's it! But I wrote method that return index in two cases:
#1: return index, if I want to chage value by existing key
`
 for item_index, item in enumerate(self.storage):
            if item:
                if item[0] == new_item[0] and item[1] == new_item[1]:
                    return item_index
`
#2 written logic, that checks is the cell empty/ If not - check next cell
`
 for item_index, item in enumerate(self.storage):
            if item:
                if item[0] == new_item[0] and item[1] == new_item[1]:
                    return item_index
`

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+        while True:
+            if not self.storage[index]:
+                return index
+            index = (index + 1) % self.storage_length
+
+    def resize(self) -> None:
+        temp_storage = [i for i in self.storage if i]
+        self.storage_length *= 2
+        self.storage = [None] * self.storage_length
+        for i in temp_storage:
```
vsmutok, line 31: `i`  not clear variable name

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #689, opened 2024-06-11
Solution

SrMusienko:


app/main.py
```diff
@@ -1,2 +1,71 @@
...
+        index = hash(find_key) % self.capacity
+        while True:
+            key_d, val = self.hash_table[index]
+            if key_d != find_key:
+                coli += 1
+                index += 1
+                if index > self.capacity:
+                    index = 0
+                if coli > self.capacity * 2:
+                    raise KeyError
```
DevSpaciX, line 47: Raise erros with error message

tests/test_main.py
```diff
@@ -161,6 +161,7 @@ def test_dictionary_add_with_mocked_hash(
         dictionary[key] = value
 
     for key, value in pairs_after_adding:
+        print(dictionary[key])
```
DevSpaciX, line 164: Is that necessery ?

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+                if index > self.capacity:
+                    index = 0
+                if coli > self.capacity * 2:
+                    raise KeyError("Error due to big count of collisions!!")
+            else:
+                return val
+
+    @staticmethod
+    def create_hash_table(size: int) -> list:
+        return [[None, None] for _ in range(size + 1)]
```
DevSpaciX, line 53: You can try this `self.hash_table: list[None | Node] = [None] * capacity`

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+
+    def resize(self) -> None:
+        temp = self.hash_table
+        self.capacity *= 2
+        self.hash_table = self.create_hash_table(self.capacity)
+        for fill_cell in temp:
+            if fill_cell != [None, None]:
+                key, val = fill_cell
+                index = hash(key) % self.capacity
+                while self.hash_table[index] != [None, None]:
```
DevSpaciX, line 67: ` while self.hash_table[index] is not None` It will be the same

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+                if index > self.capacity:
+                    index = 0
+            self.hash_table[index] = [key, value]
+            self.size -= 1
+        else:
+            while self.hash_table[index] != [None, None]:
+                index += 1
+                if index > self.capacity:
+                    index = 0
+            self.hash_table[index] = [key, value]
```
DevSpaciX, line 34: Also look , your logic is repeating in the different methods ( same while loop , almost same cheks etc ) . Try to create some func to reuse your logic

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        self.hash_table = [[None, None]] * (self.capacity + 1)
+
+    def __len__(self) -> int:
+        return self.size
+
+    def tst_ind(self, ar_ind: int) -> int:
+        ar_ind = ar_ind + 1
+        if ar_ind > self.capacity:
+            ar_ind = 0
+        return ar_ind
```
DevSpaciX, line 19: Pay more attention on naming ( functions , viriables etc ) tst_ind ?? ar_ind ?? It is not descriptive ( Fix that please

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        index = hash(find_key) % self.capacity
+        while True:
+            key_d, val = self.hash_table[index]
+            if key_d != find_key:
+                coli += 1
+                index = self.tst_ind(index)
+                if coli > self.capacity * 2:
+                    raise KeyError("Error due to big count of collisions!!")
+            else:
+                return val
```
DevSpaciX, line 49: Can we remove `else` here and leave just `return` ? Also it is better to call return at the start and rewrite your condition to make your code less nested

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+    def calculate_index(self, hash_key: int) -> int:
+        index = hash_key % self.capacity
+        return index
+
+    def resize(self) -> None:
+        temp = self.hash_table
+        self.capacity *= 2
+        self.hash_table = [[None, None]] * (self.capacity + 1)
+        for fill_cell in temp:
+            if fill_cell != [None, None]:
```
DevSpaciX, line 59: `if fill_cell:` it will be the same

DevSpaciX (approved):
Good , check comment I left

PR #690, opened 2024-06-11
'Solution'

dsokolovskiy:


app/main.py
```diff
@@ -1,2 +1,92 @@
...
+        index = hashed_key % self.__capacity
+
+        if self.table[index] is not None:
+            for i, (k, h, v) in enumerate(self.table[index]):
+                if k == key:
+                    del self.table[index][i]
+                    self.length -= 1
+                    return
+
+        raise KeyError(key)
```
DevSpaciX, line 28: Imho `key` it is not very descriptive error message : ( Fix that please

DevSpaciX (changes requested):

app/main.py
dsokolovskiy: ok, ty. I try to fix that)

DevSpaciX (approved):

PR #692, opened 2024-06-11
solution

NikitaNikolaitsev:


app/main.py
```diff
@@ -1,2 +1,57 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.hash_table = self.create_hash_table()
```
DevSpaciX, line 7: You can try this [None] * capacity

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        return int(self.capacity * (2 / 3))
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        key_hash = hash(key)
+        position = key_hash % self.capacity
+        while True:
+            if not self.hash_table[position]:
+                self.hash_table[position] = (key, value, key_hash)
+                self.size += 1
+                if self.size >= self.load_factor:
```
DevSpaciX, line 24: Also look , your logic is repeating in the different methods ( same while loop , almost same cheks etc ) . Try to create some func to reuse your logic

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        self.hash_table = self.create_hash_table()
+        self.size = 0
+        for i in old_hash_table:
+            if i:
+                key, value, key_hash = i
+                self.__setitem__(key, value)
+
+    def __len__(self) -> int:
+        return self.size
+# im tired, i need more practice
```
DevSpaciX, line 57: 🥲

DevSpaciX (changes requested):
DRY and will be perfect !

DevSpaciX (approved):

PR #696, opened 2024-06-11
py-dictionary v1

BornToLivee:
i didnt do__delitem__ method. As i understand, i need to reposition my nodes after __delitem__ method?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,2 +1,84 @@
+from dataclasses import dataclass
+from typing import Hashable, Any
+
+
+@dataclass
+class Node:
+    key: Hashable
+    hash_: int
+    value: Any
+
+
 class Dictionary:
-    pass
+    INITIAL_CAPACITY = 8
+    LOAD_FACTOR = 2 / 3
+
+    def __init__(self, capacity: int = INITIAL_CAPACITY) -> None:
+        self.capacity = capacity
+        self.length = 0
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,2 +1,84 @@
...
+        return self.length
+
+    def pop(self, key: Hashable) -> Any:
+        index = self.get_index(key)
+        value = self.hash_table[index].value
+        if self.hash_table[index] is None:
+            raise KeyError("No such key")
+        self.hash_table[index] = None
+        self.length -= 1
+        return value
```
DevSpaciX, line 63: You need to del old key and value . 
Like this 

```
value = self[key]
del self[key]
```

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+    def clear(self) -> None:
+        self.hash_table = [None] * self.capacity
+        self.length = 0
+
+    def get(self, key: Hashable) -> Any:
+        index = self.get_index(key)
+        value = self.hash_table[index].value
+        if self.hash_table[index] is None:
+            raise KeyError("No such key")
+        return value
```
DevSpaciX, line 74: Maybe just call `self[key]`? Try it in other places

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #697, opened 2024-06-11
sol1

IvanVahanov:


app/main.py
```diff
@@ -1,2 +1,97 @@
...
+        index = hash(key) % self.capacity
+
+        while self.hash_table[index] is not None:
+            if self.hash_table[index].key == key:
+                self.hash_table[index].value = value
+                return
+            index = (index + 1) % self.capacity
+
+        self.hash_table[index] = Node(key, value)
+        self.size += 1
```
DevSpaciX, line 32: Look , your logic is repeating in the different methods ( same while loop , almost same cheks etc ) . Try to create some func to reuse your logic

DevSpaciX (changes requested):
DRY

PR #698, opened 2024-06-11
Solution_V1

BileichukIvan:


app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        while True:
+            if not self.hash_table[position]:
+                self.hash_table[position] = (key, value, hash_key)
+                self.size += 1
+                if self.size >= self.load_factor:
+                    self.resize()
+                break
+            elif self.hash_table[position][0] == key:
+                self.hash_table[position] = (key, value, hash_key)
+                break
```
DevSpaciX, line 32: You can write you while loop like this :
`while not self.hash_table[position]:`
That way your code will bw less nested imho. :)

DevSpaciX (changes requested):

PR #701, opened 2024-06-11
solution

DimaYankovskiy:


app/main.py
```diff
@@ -1,2 +1,119 @@
...
+
+    def pop(self, key: Hashable, default: any = None) -> any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            if default is None:
+                raise
+            return default
```
DevSpaciX, line 100: It is better to add some error message for user

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #702, opened 2024-06-11
Develop

Unlie9:


app/main.py
```diff
@@ -1,2 +1,56 @@
...
+                self.hash_table[index] = (key, value, hash_key)
+                self.size += 1
+                if self.size == self.load_factor:
+                    self.resise_table()
+                break
+            elif self.hash_table[index][0] == key:
+                self.hash_table[index] = (key, value, hash_key)
+                break
+            index += 1
+            index %= self.capacity
```
DevSpaciX, line 29: You can make while loop loke here : `while self.hash_table[index] is not None:` and this way your code will bw less nested

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+
+    def __getitem__(self, key: int) -> None:
+        hash_key = hash(key)
+        index = hash_key % self.capacity
+
+        while True:
+            if self.hash_table[index] is None:
+                raise KeyError
```
DevSpaciX, line 37: Try to pass some error message for better user experience in the future

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #703, opened 2024-06-12
'Solution'

Irina17191:


app/main.py
```diff
@@ -1,2 +1,114 @@
...
+        if self._hash_table[index] is None:
+            raise KeyError(f"No such key: {key}")
+
+        return self._hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._get_index(key)
+
+        if self._hash_table[index] is None:
+            raise KeyError(f"No such key: {key}")
```
DevSpaciX, line 75: It will rise error is not it ? For this type of check it is better to use `get()`

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+            return self[key]
+        except KeyError:
+            return default
+
+    def __contains__(self, key: Hashable) -> bool:
+        index = self._get_index(key)
+
+        if self._hash_table[index] is None:
+            return False
+        return True
```
DevSpaciX, line 94: Write it in one line : )

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+
+            return value
+
+        if len(other) == 0:
+            raise KeyError("No such key")
+        elif len(other) > 1:
+            raise TypeError(f"pop expected at most 2 arguments, "
+                            f"got {len(args)}")
+        else:
+            return other[0]
```
DevSpaciX, line 114: Can we remove `else` here?

app/main.py
Irina17191: yes, it will, rewrote it using get method

app/main.py
Irina17191: yes, sure

app/main.py
Irina17191: done)

DevSpaciX (approved):

PR #705, opened 2024-06-12
Solution

YevaYe:


app/main.py
```diff
@@ -1,2 +1,113 @@
...
+
+            return value
+
+        if len(other) == 0:
+            raise KeyError("No such key")
+        elif len(other) > 1:
+            raise TypeError(f"pop expected at most 2 arguments, "
+                            f"got {len(args)}")
+        else:
+            return other[0]
```
DevSpaciX, line 106: Can we remove else here ?

app/main.py
```diff
@@ -1,2 +1,113 @@
...
+            return self[key]
+        except KeyError:
+            return default
+
+    def __contains__(self, key: Hashable) -> bool:
+        index = self._get_index(key)
+
+        if self._hash_table[index] is None:
+            return False
+        return True
```
DevSpaciX, line 89: Can we write it in one line ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #707, opened 2024-06-12
Solution

b3v3kt0r:


app/main.py
```diff
@@ -1,2 +1,118 @@
...
+    def _resize(self) -> None:
+        new_capacity = self.capacity * 2
+        new_hash_table = [None] * new_capacity
+        for i in range(self.capacity):
+            current = self.hash_table[i]
+            while current:
+                new_index = hash(current.key) % new_capacity
+                if new_hash_table[new_index] is None:
+                    new_hash_table[new_index] = Node(current.key,
+                                                     current.value)
```
DevSpaciX, line 48: Look , your logic is repeating in the different methods ( same while loop , almost same cheks etc ) . Try to create some func to reuse your logic

DevSpaciX (changes requested):
Dry !

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        self.capacity = new_capacity
+
+    def __getitem__(self, item: Hashable) -> Any:
+        index = self._hash(item)
+        current = self.hash_table[index]
+        while current:
+            if current.key == item:
+                return current.value
+            current = current.next_node
+        raise KeyError(item)
```
DevSpaciX, line 65: Also it is not very descriptive error message imho . Fix that pls in all places

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #713, opened 2024-06-15
Develop

Timur5050:


app/main.py
```diff
@@ -1,2 +1,110 @@
...
+            if item != -1:
+                self.__setitem__(item[0], item[2])
+
+    def __setitem__(self, key: any, value: any) -> None:
+        if self.size + 1 > self.capacity * 2 / 3:
+            self.resize_dict()
+        hash_for_key = self.hash_func(key)
+
+        if (self.dictionary[hash_for_key] == -1
+                or self.dictionary[hash_for_key] == "deleted"):
```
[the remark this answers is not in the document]
Timur5050: because there can be a situation like:
dictionary[2] = 5 
then : del dictionary[2] -> dictionary["deleted"] (it is indexed by 2 in list)
and when we need to set again : dictionary[2] = 10, we need to check also if it is not deleted one

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+                temp_counter = (temp_counter + 1) % self.capacity
+
+            if (not self.dictionary[temp_counter]
+                    or self.dictionary[temp_counter] == "deleted"):
+                self.dictionary[temp_counter] = [key, hash_for_key, value]
+                self.size += 1
+            else:
+                self.dictionary[temp_counter][2] = value
+
+    def __getitem__(self, item: any) -> any:
```
vsmutok, line 51:
```suggestion
    def __getitem__(self, item: Any) -> Any:
```

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        temp_counter = hash_for_key
+        while ((self.dictionary[temp_counter]
+                or self.dictionary[temp_counter] != "deleted")
+               and self.dictionary[temp_counter][0] != key):
+            temp_counter = (temp_counter + 1) % self.capacity
+
+        if not self.dictionary[temp_counter]:
+            raise KeyError
+
+        self.dictionary[temp_counter] = "deleted"
```
vsmutok, line 82: use `del` to delete item

vsmutok (changes requested):

app/main.py
Timur5050: can u explain what the difference?

app/main.py
Timur5050: i guess it will break evrth, because for example we have list like that: [None, [1, 1, 3], None, [3, 3, 4], [4, 4, 5], None, [6, 6, 10], None]
key - 4 - has hash - 4
and if i delete 3 indexed key, that element, that had hash 4 - will have index 3, and it will break logic

Timur5050:
No, it won't work that way, if we want to set none for the deleted element, we will need to shift the others by one position after that, and this will have a bad complexity and difficult implementation

app/main.py
```diff
@@ -1,2 +1,112 @@
...
```
Dimosphen1, line 82: Decrease the dict size after deleting an item

app/main.py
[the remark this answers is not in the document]
Dimosphen1: In a way of adding a `deleted` flag you introduce complexity because there is a need to additionally handle `deleted` elements, at the same time there is no difference of `deleted` elements compared to `None` elements. Please, redefine the logic for setting and deleting an element to omit using the `deleted` keyword.

Dimosphen1 (changes requested):

Timur5050:
u had not updated verion, now i have empty elements as None, not as -1, so in that way i need to set "deleted" to deleted element, because if i set None to deleted and None to all empty element it will not work correctly with getting and setting elements.

Timur5050:
i tried, and if to do without "deleted" keyword - i will need to shift every element after deleted one, and it was so hard to write it correctly, many conditions to check and so on. but i dont understand what problem with that "deleted" - i guess that's nice practive

Polyakiv-Andrey (approved):
Good job!

PR #717, opened 2024-07-03
Solution

b4oody:


app/main.py
```diff
@@ -1,2 +1,52 @@
+from typing import Any, Iterable, Iterator, Tuple
+
+
 class Dictionary:
-    pass
```
pavlopro, line 2: Be attentive to basic requirements for the implementation of the dictionary (initial capacity, load factor, resize ...)
Also, don't forget to store (key, hash, value) as a node in the hash table. A hash table may be stored as a simple list of nodes.

If you want check your __delitem__ method uncomment test_deletion(): test in ./tests/test_main.py

pavlopro (changes requested):

PR #719, opened 2024-07-06
Solution

Katy4S:


app/main.py
```diff
@@ -1,2 +1,101 @@
...
 class Dictionary:
-    pass
+    def __init__(self, initial_capacity: int = 8,
+                 load_factor: float = 0.75) -> None:
+        self.load_factor = load_factor
+        self.capacity = initial_capacity
+        self.size = 0
+        self.hash_table: List[List[Node]] = [[] for _ in range(self.capacity)]
+
+    def _hash(self, key: Any) -> int:
```
vsmutok, line 19:
```suggestion
    def _hash(self, key: Hashable) -> int:
```
please fix everywhere

vsmutok (changes requested):

app/main.py
```diff
@@ -1,2 +1,101 @@
+from typing import Any, List, Hashable
+
+
+class Node:
+    def __init__(self, key: Any, hash_value: int, value: Any) -> None:
```
EdAlekseiev, line 5: `key` must be `Hashable`
```suggestion
    def __init__(self, key: Hashable, hash_value: int, value: Any) -> None:
```

EdAlekseiev (changes requested):
Good job

app/main.py
Katy4S: thank you!

vasylhnatiuk (approved):

PR #725, opened 2024-07-16
Solution

Wallian169:


app/main.py
```diff
@@ -1,2 +1,75 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: Any, value: Any) -> None:
```
pavlopro, line 5: key can't be Any, it should be Hashable
fix in all places

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #727, opened 2024-07-19
implemented my custom dict

MasakDirt:


app/main.py
```diff
@@ -1,2 +1,78 @@
+from typing import Any
+from copy import deepcopy
+
+
+class Node:
+    def __init__(self, key: Any, value: Any) -> None:
```
oliasymak24, line 6: The correct type annotation for key is Hashable, check in all places

app/main.py
oliasymak24, line 1: Will be good to implement some __delitem__ method

oliasymak24: Did you run tests?
Run it and check everything once again

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,75 @@
+from typing import Any, Hashable
+from copy import deepcopy
+
+
+class Node:
```
oliasymak24, line 5: Will be better to put Node class inside in Dictionary class

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+    def __init__(self, key: Hashable, value: Any) -> None:
+        self.key = key
+        self.value = value
```
oliasymak24, line 8: Hash must be a part of Node class too

app/point.py
```diff
@@ -19,6 +19,9 @@ def __hash__(self) -> int:
         # in the hashtable as well
         return hash((self.x, self.y))
 
+    def __repr__(self) -> str:
+        return f"Point({self.x}, {self.y})"
+
```
oliasymak24, line 24: To don't need to change Point class

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+            self.key = key
+            self.value = value
+            self.my_hash = my_hash
+
+        def __repr__(self) -> str:
+            return f"Node(key={self.key}, value={self.value})"
+
+    def __init__(self) -> None:
+        self.__capacity = 8
+        self.__load_factor = 2 / 3
```
Oleksl888, line 17: you can accurately present 2/3 with Fraction type

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+        self.__size = 0
+        self.__arr = [0 for _ in range(self.__capacity)]
```
Oleksl888, line 19: why not call this structure a hash_table? also it is better to fill it up with None values. Also it will be beneficial to annotate that hash table is a list of Nodes or None values

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.__ready_to_resize():
```
Oleksl888, line 22: No need to put a simple expression in a function and not reuse it

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+            self.__resize()
+
+        index = self.__find_index(key)
+        if self.__arr[index] == 0:
+            self.__size += 1
+        self.__add_node(index, key, value)
```
Oleksl888, line 28: same here - it is a simple operation, you can set value here directly instead of creating function you never reuse

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+
+        return self.__arr[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.__find_index(key)
+        if self.__arr[index] == 0:
+            raise KeyError(f"Value for key '{key}' not found")
+
+        self.__arr[index] = 0
+        self.__size -= 1
```
Oleksl888, line 43: you need to recalculate remaining positions after deleting an element because indexes may be wrong for elements that had collisions

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+
+        def __repr__(self) -> str:
+            return f"Node(key={self.key}, value={self.value})"
+
+    def __init__(self) -> None:
+        self.__capacity = 8
+        self.__load_factor = 0.667
+        self.__size = 0
+        self.__hash_table: list[None | Dictionary.Node] = [
+            None for _ in range(self.__capacity)]
```
Oleksl888, line 20: you could do it more simple with `[Node] * self._capacity`

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.__find_index(key)
+        if self.__hash_table[index] is None:
+            raise KeyError(f"Value for key '{key}' not found")
+
+        self.__hash_table[index] = None
+        self.__size -= 1
+
+        for values in self.__hash_table:
```
Oleksl888, line 47: For node in hash_table would be a better name

Oleksl888: it is actually not a good design to iterate through a list and change it along the way - in this case is ok, it passes the tests but beware and avoid doing this

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+    def __find_index(self, key: Hashable) -> int:
+        index = hash(key) % self.__capacity
+        while (self.__hash_table[index] is not None
+               and self.__hash_table[index].key != key):
+            index = (index + 1) % self.__capacity
+
+        return index
+
+    def clear(self) -> None:
+        self.__init__()
```
Oleksl888, line 66: init should be called only once on object creation, it is not recommended to call it multiple times within the class

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+    def __resize(self) -> None:
+        self.__capacity = self.__capacity * 2
+        self.__size = 0
+        copy_arr: list = deepcopy(self.__hash_table)
+        self.__hash_table: list[None | Dictionary.Node] = [
+            None for _ in range(self.__capacity)]
+
+        for values in copy_arr:
+            if values is not None:
+                self.__setitem__(values.key, values.value)
```
Oleksl888, line 77: you can use syntactic sugar interface because you defined magic methods already

Oleksl888 (approved):
Approved but please check last comments

PR #728, opened 2024-07-25
Implement Dictionary class

anton-bilous:


tests/test_main.py
```diff
@@ -6,18 +6,18 @@
...
+    items = [(f"Element {i}", i) for i in range(1000)]
+    dictionary = Dictionary()
+    for key, value in items:
+        dictionary[key] = value
+    for key, value in items:
+        assert dictionary[key] == value
+    assert len(dictionary) == len(items)
+    for key, value in items:
+        del dictionary[key]
+    print(len(dictionary))
```
Oleksl888, line 20: No need to push changes in this file to remote

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+from typing import Any, Iterable, Hashable
+
+
+_not_provided = object()
+
+
 class Dictionary:
-    pass
+    DEFAULT_SIZE = 8
+    THRESHOLD = 2 / 3
```
Oleksl888, line 10: you can use Fraction here

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+
+    def __init__(self, items: Iterable = (), **kwargs) -> None:
+        self.length = 0
+        self.hash_table = [None] * self.DEFAULT_SIZE
```
Oleksl888, line 14: Would be nice to annotate hash table that it is a list of Nodes or None's

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+
+        for key, value in items:
+            self[key] = value
+        self.update(kwargs)
+
+    def __getitem__(self, key: Hashable) -> Any:
+        _, _, value = self.hash_table[self._get_with_hash(key, hash(key))]
```
Oleksl888, line 21: We actually looking for a different solution for hash table - you can create a Node class to store necessary data - it will structure your code better

   

>  Also, don't forget to store (key, hash, value) as a node in the hash table. A hash table may be stored as a simple list of nodes.

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        self.hash_table = [None] * len(old_hash_table) * 2
+
+        for saved_key, saved_hash, saved_value in filter(None, old_hash_table):
+            self._set_with_hash(saved_key, saved_hash, saved_value)
+
+    def _get_with_hash(self, key: Hashable, hash_: int) -> int:
+        hash_table_len = len(self.hash_table)
+        first_index = index = hash_ % hash_table_len
+
+        while True:
```
Oleksl888, line 49: you can separate logic of finding index in a separate method and then reuse it in both set and get methods

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        self.__init__()
+
+    def items(self) -> Iterable[tuple[Hashable, Any]]:
+        for row in self.hash_table:
+            if not row:
+                continue
+            key, _, value = row
+            yield key, value
+
+    def update(self, other: dict | Dictionary) -> None:
```
Oleksl888, line 90: update method also accepts an Iterable with key-value pairs

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        for key, value in other.items():
+            self[key] = value
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888, line 94: annotate that default is an optional parameter

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = _not_provided) -> Any:
```
Oleksl888, line 100: why would you put default value as `object()`? Signature of this method should be same as in `get` method no?

app/main.py
```diff
@@ -1,2 +1,128 @@
...
+            value = self[key]
+        except KeyError:
+            if default is _not_provided:
+                raise
+            return default
+        del self[key]
+        return value
+
+
+class DictionaryIterator:
```
Oleksl888, line 111: This class is part of internal Dict implementation so it would make sense to define it inside of Dictionary.
Also your next method is not very optimal - it will attempt to run a while loop on every call. You can make a better iterator with generators here :wink:

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,136 @@
...
+        else:
+            self.hash_table[index] = self.__class__.Node(
+                hash_, self.hash_table[index].key, value
+            )
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __iter__(self) -> DictionaryIterator:
+        return self.__class__.DictionaryIterator(self)
```
Oleksl888, line 103: You could make it even more simple with generator expression which is part of your DictionaryIterator implementaion :
```
def __iter__(self):
    return (node for node in self.hash_table if node is not None)

Oleksl888 (approved):
Well done!

PR #729, opened 2024-07-28
Implemented solution

dimak20:


app/main.py
```diff
@@ -1,2 +1,52 @@
...
+from copy import deepcopy
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+    ) -> None:
+        self._capacity = 8
+        self._load_factor = 2 / 3
```
Oleksl888, line 10: you could use `Fraction` here

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        self._threshold = self._capacity * self._load_factor
+        self._hash_table = [[] for _ in range(self._capacity)]
```
Oleksl888, line 12: So this suggests that you create a separate data structure `Node` that has these attributes `key, hash, value`. If you do so your code will be better structured

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        self._size = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self._size > self._threshold:
+            self._get_resized()
+        try:
+            calc_hash = hash(key)
+        except TypeError:
```
Oleksl888, line 20: whats the point of catching error and reraising it?

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+    def _get_resized(self) -> None:
+        self._capacity = self._capacity * 2
+        self._threshold = self._capacity * self._load_factor
+        copy_table = deepcopy(self._hash_table)
+        self._hash_table = [[] for i in range(self._capacity)]
+        self._size = 0
+        for ext_cell in copy_table:
+            for key, value, _ in ext_cell:
+                self.__setitem__(key, value)
```
Oleksl888, line 52: Once you defined `__seitem__` above you can use syntactic sugar in other methods

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+    def __init__(
+            self,
+    ) -> None:
+        self._capacity = 8
+        self._load_factor = 2 / 3
+        self._threshold = self._capacity * self._load_factor
+        self._hash_table = [[] for _ in range(self._capacity)]
+        self._size = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
Oleksl888, line 15: You can use `Hashable` here to be more precise

Oleksl888 (changes requested):

app/main.py
dimak20: But there is a custom dictionary and I did not wanna create Nodes deliberately. And It suggests that I need to change all logic 'cause It won't be this method of cell storing. Do I really need to change all dictionary?

oliasymak24: Usually better to ask your question in a chat because mentors don't check github 
And yes, better to implement Node class

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #732, opened 2024-08-03
Solution

SkepskyiDanylo:
I'm going to implement other extra tasks, but for now i have no time, so i want to finish at least this

### DIFF
--- a/app/point.py
+++ b/app/point.py
@@ -17,7 +17,7 @@ class Point:
         # which is NOT a best practice, but you will be able to predict
         # a hash value by coordinates of the point and its index
         # in the hashtable as well
-        return hash((self.x, self.y))
+        return hash(self.x + self.y)
 
     @property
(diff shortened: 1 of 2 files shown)


app/point.py
```diff
@@ -17,7 +17,7 @@ def __hash__(self) -> int:
         # which is NOT a best practice, but you will be able to predict
         # a hash value by coordinates of the point and its index
         # in the hashtable as well
-        return hash((self.x, self.y))
+        return hash(self.x + self.y)
```
vsmutok, line 20: Please donʼt change this file

vsmutok (changes requested):

app/point.py
SkepskyiDanylo: my bad, forgot to reset it

vsmutok (approved):
Great!

PR #733, opened 2024-08-05
Solution

OleksiukStepan:


app/main.py
```diff
@@ -1,2 +1,103 @@
+from __future__ import annotations
+from typing import Any, Iterator
+
+
+class Node:
```
Oleksl888, line 5: This class better be defined inside of a Dictionary since it is a part of its implementation and will not be used outside of it

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+    def __init__(self, key: Any, key_hash: int, value: Any) -> None:
+        self.key = key
+        self.key_hash = key_hash
+        self.value = value
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8) -> None:
```
Oleksl888, line 13: capacity shouldnt be parametrized - it's part of dict internal structure

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        self.capacity = capacity
+        self.load_factor = 2 / 3
```
Oleksl888, line 15: Fraction type would be perfect here

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        self.threshold = self.capacity * self.load_factor
+        self.nodes = [None] * self.capacity
```
Oleksl888, line 17: Would make sense to call it `self.hash_table` + to add type annotation that it is a list of nodes

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        self.size = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.size >= self.threshold:
+            self.resize()
+
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+
+        while self.nodes[index]:
```
Oleksl888, line 27: This block of code with slight variations repeats in other methods - make it DRY

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+            if count_loops >= self.capacity:
+                raise KeyError(f"Key {key} is not in the dict")
+
+        self.nodes[index] = None
+        self.size -= 1
+
+        if self.size < self.threshold:
+            self.resize()
+
+    def get(self, key: Any) -> Any:
```
Oleksl888, line 73: get method should also have a default value

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        try:
+            value = self.__getitem__(key)
```
Oleksl888, line 75: you can use syntactic sugar since `__getitem__` method has been defined

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        except KeyError:
+            return None
+
+        return value
+
+    def pop(self, key: Any) -> Any:
```
Oleksl888, line 81: pop method should also have a default value

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        value = self.__getitem__(key)
+        self.__delitem__(key)
+        return value
+
+    def update(self, other: Dictionary) -> None:
```
Oleksl888, line 86: other can be not only a dict but also an iterable

tests/test_main.py
```diff
@@ -6,18 +6,19 @@
...
-#     dictionary = Dictionary()
-#     for key, value in items:
-#         dictionary[key] = value
-#     for key, value in items:
-#         assert dictionary[key] == value
-#     assert len(dictionary) == len(items)
-#     for key, value in items:
-#         del dictionary[key]
-#     print(len(dictionary))
+@pytest.mark.timeout(5)
```
Oleksl888, line 9: no need to push changes to the tests to remote

app/main.py
```diff
@@ -1,2 +1,103 @@
...
+        for node in other.nodes:
+            if node:
+                self.__setitem__(node.key, node.value)
+
+    def __iter__(self) -> Iterator[Any]:
+        return iter([node.key for node in self.nodes if node])
```
Oleksl888, line 92: you can simplify this by using generator expression

Oleksl888 (changes requested):

tests/test_main.py
OleksiukStepan: I implemented the method for deleting a node, so I assumed it needed to be uncommented if I completed it.

app/main.py
```diff
@@ -1,2 +1,112 @@
+from __future__ import annotations
+
+from fractions import Fraction
+from typing import Any, Iterator
+
+INITIAL_CAPACITY = 8
```
Oleksl888, line 6: these should be class variables  inside of Dictionary since it is part of it implementation

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+LOAD_FACTOR = Fraction(2, 3)
+CAPACITY_MULTIPLIER = 2
+
+
 class Dictionary:
-    pass
+    class Node:
+        def __init__(self, key: Any, value: Any) -> None:
+            self.key = key
```
Oleksl888, line 14: what about hash value?

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+            self.value = value
+
+    def __init__(self) -> None:
+        self.capacity = INITIAL_CAPACITY
+        self.hash_table = [None] * self.capacity
```
Oleksl888, line 19: annotate types inside of hash table

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        if self.hash_table[index] is None:
+            if self.size + 1 >= self.threshold:
+                self.resize()
+                index = self._calculate_index(key)
+            self.size += 1
+
+        self.hash_table[index] = Dictionary.Node(key, value)
+
+    def __getitem__(self, key: Any) -> Any:
+        index = self._calculate_index(key)
```
Oleksl888, line 49: You must check that key and hash value of a Node equals to one that is being looked, right now you only check if key is the same

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        return self.size
+
+    def clear(self) -> None:
+        self.__init__()
+
+    def __delitem__(self, key: Any) -> None:
+        index = self._calculate_index(key)
+        loop_counter = 0
+
+        while self.hash_table[index] is None:
```
Oleksl888, line 66: delitem shares same logic as getitem why would you do this lookup? 
Also you don't need to resize table down after deleting an element

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        self.hash_table[index] = None
+        self.size -= 1
+
+        if (
+                self.size < self.threshold // 2
+                and self.capacity > INITIAL_CAPACITY
+        ):
+            self.resize()
+
+    def get(self, key: Any, default: Any = None) -> Any:
```
Oleksl888, line 81: anotate that default is optional param, key must be hashable

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Any, default: Any = None) -> Any:
```
Oleksl888, line 87: same as above + pop method works a bit different:
if key is not in dictionary and default is not set - error will be triggered,

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        value = self.get(key, default)
+        del self[key]
+        return value
+
+    def update(self, other: Iterator[Any]) -> None:
```
Oleksl888, line 92: node can be another Dictionary or Iterable with key-value pairs

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        for node in other:
+            if node:
+                self.__setitem__(node.key, node.value)
+
+    def __iter__(self) -> Iterator[Any]:
+        return (node if node else None for node in self.hash_table)
```
Oleksl888, line 98: why would you return None values? you just need to return keys from  Nodes

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+
+    def pop(self, key: Any, default: Any = None) -> Any:
+        value = self.get(key, default)
+        del self[key]
+        return value
+
+    def update(self, other: Iterator[Any]) -> None:
+        for node in other:
+            if node:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 95: you can use syntact sugar interface since you defined magic methdos already

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+        self.__init__()
+
+    def __delitem__(self, key: Any) -> None:
+        index = self._calculate_index(key, hash(key))
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
```
Oleksl888, line 72: So here we could fix the problem with wrong indexes caused by collisions by calling resize() method after deletion. You could change signature of resize method so it accepts optional parameter. And then call it with multiplier of 2 in setitem method and without params in delitem
```
 def resize(self, mulitplier=1) -> None:
        old_dict = self.hash_table
        self.capacity *= mulitplier

Oleksl888 (approved):
Approved

PR #734, opened 2024-08-06
Solution

DSahalatyi:


app/main.py
```diff
@@ -1,2 +1,242 @@
+from __future__ import annotations
+
+import copy
+import dataclasses
+from typing import Any, Callable, Iterable, Hashable
+
+
+@dataclasses.dataclass
+class Node:
```
Oleksl888, line 9: I would define this class inside of a Dictionary since it is a part of its implementation and cannot be used outside it

app/main.py
```diff
@@ -1,2 +1,242 @@
...
 class Dictionary:
-    pass
+
+    def __init__(self) -> None:
+        # added mainly to keep track of input order, iteration
+        self._keys = []
+        self._values = []
+
+        self._initial_capacity = 8
+        self._load_factor = 2 / 3
```
Oleksl888, line 23: You can use `Fraction` type here

app/main.py
```diff
@@ -1,2 +1,242 @@
...
+
+        self._capacity = self._initial_capacity
+        self._hash_table = [[] for _ in range(self._capacity)]
```
Oleksl888, line 26: Why not simplify this by filling hash_table with None values? It will be more efficient then creating one thousand empty lists. Also can you annotate hash table so it actually shows that it is a list of Nodes

app/main.py
```diff
@@ -1,2 +1,242 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+
+        if self.get_load():
+            self._data_handler(key, value)
+        else:
+            self.resize()
+            self.__setitem__(key, value)
+
+    def __len__(self) -> int:
+        return sum(1 for item in self._hash_table if item)
```
Oleksl888, line 48: Why not have an extra attribute `size` that you will use to keep track of occupied cells in hash table? Then your `len` operation will take O(1) instead of O(n)

app/main.py
```diff
@@ -1,2 +1,242 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        if self.__getitem__(key):
+            self._data_handler(key, delete=True)
+
+    def keys(self) -> str:
+        class DictKeys:
```
Oleksl888, line 55: keys and values methods overcomplicate your code

app/main.py
```diff
@@ -1,2 +1,242 @@
...
+            key = self._keys[self.current_key]
+            self.current_key += 1
+            return key
+        raise StopIteration
+
+    def update(self, other: Dictionary) -> None:
+        for item in other:
+            self.__setitem__(item, other[item])
+
+    def _data_handler(
```
Oleksl888, line 111: Whatever is below this line is an example of so called spaghetti code :spaghetti:  You need to rethink the entire concept. 
Here's the algorytm you should follow with your set item method for instance:
1. Get hash value of key
2. Check if table needs to be resized and resize if necessary
3. Create a node with all necessary data
4. Find correct index for table cell, write data, increase counter of occupied table cells.
5. Code will be  pretty much the same for `__getitem__` method.
6. It should take you up to 30  lines of code to implement this logic including methods to resize table and traverse table to find right index.

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+@dataclass
+class Node:
+    key: Hashable
+    k_hash: int
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = CAPACITY) -> None:
```
Oleksl888, line 21: capacity shouldnt be parametrized - it's part of dict internal structure

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+            if node is not None:
+                self.__setitem__(node.key, node.value)
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.get_index(key)
+
+        if self.hashtable[index] is None:
+            if self.size + 1 >= self.resize_threshold:
+                self.resize()
+                return self.__setitem__(key, value)
```
Oleksl888, line 64: you can avoid making a recursive call here - check if table needs resizing, resize if necessary, find index and set value

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+            self.size += 1
+            self.keys.append(key)
```
Oleksl888, line 66: it is an overkill to keep keys in a separate list - you already have hash_table for that

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+
+        if self.hashtable[index] is None:
+            raise KeyError(f"{key}")
+
+        self.hashtable[index] = None
+        self.keys.remove(key)
+        self.size -= 1
+
+    def clear(self) -> None:
+        for i in range(len(self.hashtable)):
```
Oleksl888, line 84: this operation takes too long, should be closer to constant complexity + you don't change the size and table capacity

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+            raise KeyError(f"{key}")
+
+        self.hashtable[index] = None
+        self.keys.remove(key)
+        self.size -= 1
+
+        return node.value
+
+    def update(self, other: Dictionary) -> None:
+        if isinstance(other, Dictionary):
```
Oleksl888, line 113: other can also be an iterable of key-value pairs

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+            for node in other:
+                if node is not None:
+                    self.__setitem__(node.key, node.value)
```
Oleksl888, line 116: you can use synt sugar since you already defined magic methods

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+            raise TypeError(
+                f"{other} is {type(other).__name__}, not a Dictionary"
+            )
+
+    def __iter__(self) -> Dictionary:
+        self.current_key = 0
+        return self
+
+    def __next__(self) -> Node:
+        while self.current_key < len(self.keys):
```
Oleksl888, line 127: this is not very optimal since you will be running a while loop on every `next` call. try with generator expressions

app/main.py
```diff
@@ -1,2 +1,133 @@
...
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
+        index = self.get_index(key)
+
+        if self.hashtable[index] is None:
+            return default
+
+        return self.hashtable[index]
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888, line 97: `Dictionary.pop(key, None)` should return None instead of triggering KeyError

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,133 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from fractions import Fraction
+from typing import Any, Hashable
+
+
+CAPACITY = 8
```
Oleksl888, line 8: These constants as well as Node class a part of dict internal implementation so it would make sense to package everything inside of Dictionary class

oliasymak24 (changes requested):

app/main.py
Oleksl888: This hasn't been done.

app/main.py
```diff
@@ -1,2 +1,142 @@
...
+
+        self.hashtable[index] = None
+        self.keys.remove(key)
+        self.size -= 1
+
+    def clear(self) -> None:
+        self.__init__()
+        self.keys = []
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888, line 91: Need to annotate that default is Optional

app/main.py
```diff
@@ -1,2 +1,142 @@
...
+        index = self.get_index(key)
```
Oleksl888, line 92: why not simply get it with brackets in try block?

app/main.py
```diff
@@ -1,2 +1,142 @@
...
+
+        if self.hashtable[index] is None:
+            return default
+
+        return self.hashtable[index]
+
+    _NOT_PROVIDED = object()
+
+    def pop(self, key: Hashable, default: Any = _NOT_PROVIDED) -> Any:
+        index = self.get_index(key)
```
Oleksl888, line 102: here you can also use sugar interface to get item by key in try block and then catch/reraise exceptions. Plus mark default as optional here too

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,144 @@
...
+                    raise ValueError(
+                        "Iterable must contain pairs of (key, value)"
+                    )
+        else:
+            raise TypeError(
+                f"Expected a dict or iterable of pairs, "
+                f"got {type(other).__name__}"
+            )
+
+    def __iter__(self) -> Dictionary:
```
Oleksl888, line 139: This will return a Generator

app/main.py
```diff
@@ -1,2 +1,144 @@
...
+        self.__init__()
+        self.keys = []
+
+    def get(self, key: Hashable, default: Optional[Any] = None) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    _NOT_PROVIDED = object()
```
Oleksl888, line 97: this should be moved above, so you have all constants in one place

app/main.py
```diff
@@ -1,2 +1,144 @@
...
+            self.capacity = Dictionary.INITIAL_CAPACITY
+        self.hashtable: list[Node | None] = [None] * self.capacity
+        self.size = 0
+        self.keys = []
+
+    def get_index(self, key: Hashable) -> int:
+        index = hash(key) % self.capacity
+
+        while (
+            self.hashtable[index] is not None
```
Oleksl888, line 33: I guess you also need to check if hash value is the same not only if keys match

Oleksl888 (approved):
Approved but check last comments

PR #735, opened 2024-08-08
Dictionary class implemented

danilsiv:


app/main.py
```diff
@@ -1,2 +1,81 @@
...
+                return el[2]
+        raise KeyError(key)
+
+    def get(self, key: Any, default_value: Any = "some value") -> Any:
+        try:
+            self.__getitem__(key)
+        except KeyError:
+            return default_value
+
+    def pop(self, key: Any) -> Any:
```
LLkaia, line 58: Method pop should remove item after it was retrieved

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        value = self.__getitem__(key)
+        return [key, value]
+
+    def update(self, other: list[tuple] | list[list]) -> None:
```
LLkaia, line 62: II think, another dict should be an input for this function

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+    def __len__(self) -> int:
+        return self.current_size
+
+    def clear(self) -> None:
+        self.hash_table = [[] for _ in range(len(self.hash_table))]
+
+    def __delitem__(self, key: Any) -> None:
+        for el in self.hash_table:
+            if el and el[0] == key and el[1] == hash(key):
+                return el[2]
```
LLkaia, line 49: This is not removing item

LLkaia (changes requested):
Also try to simplify setitem method.

app/main.py
danilsiv: but we can not use the dict in this task

LLkaia: yep, you are right. from my point of view, it should be another instance of your custom class `Dictionary`

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.hash_table = [None for _ in range(8)]
+        self.current_size = 0
+        self.capacity = len(self.hash_table)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
vasylhnatiuk, line 20:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

app/main.py
```diff
@@ -1,2 +1,89 @@
...
+            index = (index + 1) % self.capacity
+
+    def __resize_hash_table(self) -> None:
+        old_hash_table = [node for node in self.hash_table if node is not None]
+        self.hash_table = [None for _ in range(self.capacity * 2)]
+        self.capacity = len(self.hash_table)
+        self.current_size = 0
+        self.update(old_hash_table)
+
+    def __getitem__(self, key: Any) -> Any:
```
vasylhnatiuk, line 46:
```suggestion
    def __getitem__(self, key: Hashable) -> Any:
```

vasylhnatiuk: use everywhere for key

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #736, opened 2024-08-08
solution

BastovOleksandr:


app/main.py
```diff
@@ -1,2 +1,116 @@
+from typing import Hashable, Any, Iterator
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.__iter = 0
+        self.__capacity = 8
+        self.__threshold = 2 / 3
```
Oleksl888, line 8: Fraction type would be perfect to use here

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+        self.__table = [None] * self.__capacity
+        self.__length = 0
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if not isinstance(key, Hashable):
+            raise TypeError(f"unhashable type: {type(key)}")
+
+        index = hash(key) % self.__capacity
+        node = (hash(key), key, value)
```
Oleksl888, line 17: you should declare node as  a separate class

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+
+        if self.__iter >= self.__capacity:
+            raise StopIteration
+
+        result = self.__table[self.__iter][2]
+        self.__iter += 1
+        return result
+
+    def __find_entry(self, key: Hashable, erase: bool) -> Any:
+        if not isinstance(key, Hashable):
```
Oleksl888, line 72: no need to check it here really - `hash()` will throw a typeerror if key is not hashable

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+            raise TypeError(f"unhashable type: {type(key)}")
+
+        index = hash(key) % self.__capacity
+
+        if not self.__table[index]:
```
Oleksl888, line 77: what if we have collision and index is actually moved to the next cell? Perhaps this line of code should be placed later?

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+                if index == self.__capacity:
+                    index = 0
+
+                if not self.__table[index]:
+                    break
+
+        self.__table[index] = node
+        self.__length += 1
+
+        if self.__length > round(self.__capacity * self.__threshold):
```
Oleksl888, line 37: would make sense to check if table needs resizing before you add a new element

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+    def __double_capacity(self) -> None:
+        self.__iter = 0
+        self.__capacity *= 2
+        self.__length = 0
+        old_table = self.__table
+        self.__table = [None] * self.__capacity
+
+        for node in old_table:
+            if node is not None:
+                self.__setitem__(node[1], node[2])
```
Oleksl888, line 103: you can use syntactic sugar here already since you defined setiitem method above

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+
+    def update(self, source: dict) -> None:
+        """
+        Here we could use the "Iterable" type annotation to
+        make the method more flexible in terms of the arguments it
+        processes. Naturally, with a more complex implementation,
+        but I already want to move on to the next task :)
+        """
+        for key, value in source.items():
```
Oleksl888, line 112: you implementation should support update from tuple as well, why leave task half way done ?

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+                temp = self.__table[index]
+                if erase:
+                    self.__iter = 0
+                    self.__table[index] = None
+                return temp[2]
+            index += 1
+            if index == self.__capacity:
+                index = 0
+
+    def pop(self, key: Hashable) -> Any:
```
Oleksl888, line 91: pop method must also have default value

app/main.py
```diff
@@ -1,2 +1,116 @@
...
+            for entry in self.__table
+            if entry
+        ])
+
+    def __iter__(self) -> Iterator:
+        self.__iter = 0
+        return self
+
+    def __next__(self) -> Any:
+        while self.__iter < self.__capacity and not self.__table[self.__iter]:
```
Oleksl888, line 61: this is not very efficient - it means that every time you call next() you have to iterate through all table cell indexes. Also  `__iter` is not a very good name for a counter variable

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,134 @@
...
+            self.__counter += 1
+            next(self)
+            return
+
+        return self.__table[self.__counter - 1].value
+
+    def __find_entry(self, key: Hashable, erase: bool) -> Any:
+        index = hash(key) % self.__capacity
+
+        while True:
```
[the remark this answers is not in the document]
Oleksl888: this method would start a while loop and finds free index or finds index that contains a node with key and hash that we are looking for

app/main.py
```diff
@@ -1,2 +1,110 @@
...
+        self.__capacity = 8
+        self.__threshold = Fraction(2, 3)
+        self.__table = [None] * self.__capacity
+        self.__length = 0
+        self.__deleted = object()
+
+    def __find_index(self, key: Hashable) -> int:
+        index = hash(key) % self.__capacity
+        while True:
+            if not self.__table[index]:
```
Oleksl888, line 23: I think you could make it more simple without returning -1 Something like this:
```
        while self.__table[index] is not None:
            if  self.__table[index].key == key and  self.__table[index].hash_ == hash_value:
                return index
            index = (index + 1) % self.__capacity
        return index

File: app/main.py [hunk 12]
```diff
@@ -1,2 +1,110 @@
...
+
+    def __iter__(self) -> Iterator:
+        for node in self.__table:
+            if node and node != self.__deleted:
+                yield node.key, node.value
+
+    def pop(self, key: Hashable, default: Optional[Any] = None) -> Any:
+        index = self.__find_index(key)
+        if index == -1:
+            if default is None:
```
Oleksl888, line 70: Dictionary.pop(key, None) should return None instead of triggering KeyError

Oleksl888 (approved):
Well done! But check the last comments

PR #737, opened 2024-08-10
write my own dictionary

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1,2 +1,151 @@
+from typing import Hashable, Any
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+            keys: list[Hashable] | None = None,
```
Oleksl888, line 7: I think it would make more sense to pass an iterable here or **kwargs

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+            values: list[Any] | None = None
+    ) -> None:
+        self.__capacity = 8
+        self.__size = 0
+        self.__threshold = int((2 / 3) * self.__capacity)
```
Oleksl888, line 12: You can use Fraction type to correctly present 2/3

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+        self.__table = [(None,)] * self.__capacity
```
Oleksl888, line 13: Hash table should be a list of nodes. It would make sense to implement a separate Node class as per task requirements. It will help to better structure your implementation

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+                self.__setitem__(key, value)
+
+    def __repr__(self) -> str:
+        items = ", ".join(f"{key}: {value}" for key, value in self.items())
+        return f"{{{items}}}"
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        key_hash, index = self.__get_key_hash_and_index(key)
+
+        while self.__table[index][0] is not None:
```
Oleksl888, line 26: This block of code with slight variations repeats in other methods - make it DRY

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+                self.__table[index] = (key, key_hash, value)
+                return
+            index += 1
+            if index == self.__capacity:
+                index = 0
+
+        self.__table[index] = (key, key_hash, value)
+        self.__size += 1
+
+        if self.__size > self.__threshold:
```
Oleksl888, line 37: Actually resizing takes place before a new element is added - of course if it will reach the threshold with the new element

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+                index = 0
+
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return self.__size
+
+    def __iter__(self) -> "Dictionary":
+        self.__current_index = 0
+        self.__keys = [element[0] for element in self.__table if element[0]]
```
Oleksl888, line 79: you actually have same expression in `keys` method.
Also I think here using generator as iterator would be more simple here

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+            raise StopIteration
+        result_to_return = self.__keys[self.__current_index]
+        self.__current_index += 1
+        return result_to_return
+
+    def clear(self) -> None:
+        self.__table = [(None,)] * self.__capacity
+        self.__size = 0
+
+    def get(self, key: Hashable, value: Any = None) -> Any | None:
```
Oleksl888, line 93: check annotation - you need to mark that `value` is Optional and return value should be just `Any`

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+        try:
+            return self.__getitem__(key)
+        except KeyError:
+            return value
+
+    def pop(self, key: Hashable, defaultvalue: Any = None) -> Any:
+        try:
+            value_to_return = self.__getitem__(key)
```
Oleksl888, line 101: You can use syntact sugar since you already defined magic methods in your class

app/main.py
```diff
@@ -1,2 +1,151 @@
...
+            self.__delitem__(key)
+            return value_to_return
+        except KeyError:
+            if defaultvalue:
+                return defaultvalue
+            raise
+
+    def update(self, keys: list[Hashable], values: list[Any]) -> None:
```
Oleksl888, line 109: actually update method expects another dictionary or an iterable with key value pairs

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,159 @@
...
+
+    def __init__(
+            self,
+            initial_data: Optional[Iterable[tuple[Hashable, Any]]] = None,
+            **kwargs
+    ) -> None:
+        self.__capacity = 8
+        self.__size = 0
+        self.__threshold = int(Fraction(2, 3) * self.__capacity)
+        self.__table = [Dictionary.__Node()] * self.__capacity
```
Oleksl888, line 26: would be more optimal to fill out empty hash table with None values instead of creating `capacity` times instances of Node

app/main.py
```diff
@@ -1,2 +1,159 @@
+from fractions import Fraction
+from typing import Any, Hashable, Iterable, Optional
+
+
 class Dictionary:
-    pass
+    class __Node:
+        def __init__(
+                self,
+                key: Optional[Hashable] = None,
```
Oleksl888, line 9: I think these params shouldn't be optional, if you refactor hash table to hold None values instead of empty Nodes it will be possible

app/main.py
```diff
@@ -1,2 +1,159 @@
...
+        items = ", ".join(f"{key}: {value}" for key, value in self.items())
+        return f"{{{items}}}"
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.__calculate_index(key)
+
+        if self.__table[index].key is None:
+            if self.__size + 1 > self.__threshold:
+                self.__resize()
+                return self.__setitem__(key, value)
```
Oleksl888, line 43: You can refactor this so you don't have to use recursion here. You need to use recursion with caution as it may increase algorithmic complexity dramatically.
So here you can check if table needs resizing, resize if necessary, then calculate index and set node.

app/main.py
```diff
@@ -1,2 +1,159 @@
...
+        return self.__table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.__calculate_index(key)
+
+        if self.__table[index].key is None:
+            raise KeyError(key)
+
+        self.__table[index] = Dictionary.__Node()
+        self.__table[index].deleted = True
```
Oleksl888, line 63: you don't really need a deleted attr for Node. If can just set that location in hash table to None. Again this will be possible if you change your implementation for hash table to hold None values

app/main.py
```diff
@@ -1,2 +1,159 @@
...
+        self.__size -= 1
+
+    def __len__(self) -> int:
+        return self.__size
+
+    def __iter__(self) -> "Dictionary":
+        self.__current_index = 0
+        self.__keys = (node.key for node in self.__table if node.key)
```
Oleksl888, line 71: you can simply return this generator expression and it will be most efficient iterator. Then you will not need to store __keys and implement __next__ method

app/main.py
```diff
@@ -1,2 +1,159 @@
...
+            **kwargs
+    ) -> None:
+        if data:
+            for key, value in data:
+                self[key] = value
+
+        for key, value in kwargs.items():
+            self[key] = value
+
+    def items(self) -> set[tuple[Hashable, Any]]:
```
Oleksl888, line 112: why would you return it in a set?

Oleksl888 (changes requested):

app/main.py
vladislav-tsybuliak1: changed to list

app/main.py
vladislav-tsybuliak1: I need it because:
1) if there is a colision with _x_ index, I assign the _key, value_ to _x+1_ index
2) then if I delete the key with _x_ - I will not be able to access _x+1_
If I simply set the slot to None, I will lose track of the keys that were stored at the next positions when colision occured.

app/main.py
vladislav-tsybuliak1: Done

oliasymak24 (changes requested):
Did you implement separate Node class as Oleksandr was mentioned?

vladislav-tsybuliak1:
> Did you implement separate Node class as Oleksandr was mentioned?
Yes, it is within Dictionary class

app/main.py
Oleksl888: Yeah I see. You can avoid all this if you make changes to your calculate index function. Here how this could look like.
Imagine that your hash table is filled out with None values.
```
If you look to set a Node:  if table[index] is None -> free cell - you can use it to set a new NOde. If not check cell index+1 -> repeat until you find one you need
If you looking for a value by index: if table[index] is occupied -> check if node that is there has the same key and hash that you looking for and return that index if matching. If not check cell index+1 -> repeat until you find one you need

Comment (Oleksl888) [reply app/main.py:26]: You basically have the same logic in your calculate index method, just need to change  a little bit to make it work

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: app/main.py (hunk 10 above)
Comment (vladislav-tsybuliak1) [reply app/main.py:26]: @Oleksl888 
How then can I implement the __del__ method rightly? And deal with the getting colissioned values?
I understand what you want me to do, but this doesn't solve the issue.

If I have Nones in my hash table instead of Nodes then, for example:
1) if I ad an element: my_dict[0] = "a" --> this will assign to my hash_table[0] a Node object.
2) Then I want to add one more element: my_dict[8] = "b" --> a colission will occure. Because it will want to assign to my hash_table[0] a new Node. But in my calculate function I look to the next index. And find that it is None and assign to it: hash_table[1] a Node object.
3) In this case I have no problem to access my_dict[8], because I look to my_dict[0], I see it is not empty and keys are different, then I look to my_dict[1].

But here is the problem:
4) I delete my_dict[0] (it was with value "a"). I set None to hash_table[0]
5) Then if I want to access my_dict[8] -- I have no access. Because I look to hash_table[0] and see that it is None. I return KeyError
But actually there is this key in my dictionary...
What if I add my_dict[16], my_dict[24] and so on? 

I need somehow then to flag my deleted indexes. If the index was deleted and before colission happened -- I can move further and check other indexes in order to get the value.

File: tests/test_main.py [hunk 16]
```diff
@@ -6,18 +6,18 @@
...
-#     dictionary = Dictionary()
-#     for key, value in items:
-#         dictionary[key] = value
-#     for key, value in items:
-#         assert dictionary[key] == value
-#     assert len(dictionary) == len(items)
-#     for key, value in items:
-#         del dictionary[key]
-#     print(len(dictionary))
+@pytest.mark.timeout(5)
```
Oleksl888, line 9: please revert these changes

app/main.py
Oleksl888: Ok you can keep it as is because your implementation actually works. However it is not most efficient one - you fill out your hash table with lots of garbage values which takes extra memory.

Oleksl888: I assure you that it is possible to do without marking deleted indexes - basically the fact that a table has None at that index you pointing to means that its either free memory you can use in case if you want to set value or if it's not None you check if key and hash value is correct. Then if something doesn't add up you move to one cell on the right..

Oleksl888 (changes requested):
revert changes to tests and I will approve

app/main.py
vladislav-tsybuliak1: Thank you!

Oleksl888 (approved):
Approved :+1:

PR #738, opened 2024-08-12
first_try

yashkunn:


app/main.py
```diff
@@ -1,2 +1,94 @@
...
+                del self._buckets[index][i]
+                self._size -= 1
+                return value
+        if default is not None:
+            return default
+        raise KeyError(key)
+
+    def update(self, other: dict) -> None:
+        for key, value in other.items():
+            self[key] = value
```
oliasymak24, line 89: If update() is called without passing parameters, the dictionary remains unchanged.

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+
+    def get(self, key: Any, default: Any = None) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Any, default: Any = None) -> Any:
+        index = self._get_index(key)
+        for i, node in enumerate(self._buckets[index]):
```
oliasymak24, line 77: The convention for unused variables in for loops is to call them _.

app/main.py
```diff
@@ -1,2 +1,94 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: Any, hash_key: int, value: Any) -> None:
```
oliasymak24, line 5: The correct type annotation for key is Hashable, check everywhere

app/main.py
```diff
@@ -1,2 +1,94 @@
...
+        index = self._get_index(key)
+        for i, node in enumerate(self._buckets[index]):
+            if node.key == key:
+                value = node.value
+                del self._buckets[index][i]
+                self._size -= 1
+                return value
+        if default is not None:
+            return default
+        raise KeyError(key)
```
oliasymak24, line 85: Add some message here

oliasymak24 (changes requested):

app/main.py
yashkunn: I don't understand what you want

app/main.py
```diff
@@ -1,2 +1,96 @@
+from typing import Any, Hashable
+
+
+class Node:
```
Oleksl888, line 4: Move this class inside of Dictionary since it is part of it implementation

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+        self.key = key
+        self.hash_key = hash_key
+        self.value = value
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+            initial_capacity: int = 8,
```
Oleksl888, line 14: init should take these params - they are part of internal implementation and shouldnt be set from outside

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+            load_factor: float = 0.75
```
Oleksl888, line 15: use Fraction to correctly represent 2/3

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+    ) -> None:
+        self._capacity = initial_capacity
+        self._load_factor = load_factor
+        self._size = 0
+        self._buckets = [[] for _ in range(self._capacity)]
```
Oleksl888, line 20: your hash table is actually a list of Nodes - mark this with annotation
Also filling up hash table with empty lists is not very efficient - better to fill out with None values

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self._size / self._capacity >= self._load_factor:
+            self._resize()
+        index = self._get_index(key)
+        for node in self._buckets[index]:
+            if node.key == key:
+                node.value = value
+                return
+        self._buckets[index].append(Node(key, hash(key), value))
```
Oleksl888, line 33: you should implement your table differently - your node should be assigned directly at that index, not added to the list

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+        self._size += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._get_index(key)
+        for node in self._buckets[index]:
```
Oleksl888, line 38: This concept of keeping lists inside a list is not optimal - this is O(n) complexity and we are aiming for O(1)

Oleksl888: You should change the entire concept and implement mechanism for handling collisions

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._get_index(key)
+        for i, node in enumerate(self._buckets[index]):
+            if node.key == key:
+                del self._buckets[index][i]
+                self._size -= 1
+                return
+        raise KeyError(f"KeyError: '{key}' not found in Dictionary")
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888, line 69: annotate default as optional

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888, line 75: annotate default as optional

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+        index = self._get_index(key)
+        for i, node in enumerate(self._buckets[index]):
+            if node.key == key:
```
Oleksl888, line 78: you can use syntact sugar here since you already defined magic methods

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+                value = node.value
+                del self._buckets[index][i]
+                self._size -= 1
+                return value
+        if default is not None:
+            return default
+        raise KeyError(f"KeyError: '{key}' not found in Dictionary")
+
+    def update(self, other: dict) -> None:
```
Oleksl888, line 87: other can be a Dictionary or Iterable with key-value pairs

app/main.py
```diff
@@ -1,2 +1,96 @@
...
+        if other is None:
+            return
+        for key, value in other.items():
+            self[key] = value
+
+    def __iter__(self) -> Any:
```
Oleksl888, line 93: Check return type here

tests/test_main.py
```diff
@@ -6,18 +6,18 @@
...
-#     for key, value in items:
-#         dictionary[key] = value
-#     for key, value in items:
-#         assert dictionary[key] == value
-#     assert len(dictionary) == len(items)
-#     for key, value in items:
-#         del dictionary[key]
-#     print(len(dictionary))
+@pytest.mark.timeout(5)
+def test_deletion():
```
Oleksl888, line 10: revert changes to this file

Oleksl888 (changes requested):

app/main.py
Oleksl888: why do you think it is necessary to pass capacity on init?

app/main.py
```diff
@@ -1,2 +1,106 @@
...
+    class _Node:
+        def __init__(self, key: Hashable, value: Any) -> None:
+            self.key = key
+            self.hash_key = hash(key)
+            self.value = value
+
+    def __init__(
+            self,
+            initial_capacity: int = 8,
+            load_factor: float = 2 / 3
```
Oleksl888, line 14: you can use Fraction here to accurately present 2/3

app/main.py
```diff
@@ -1,2 +1,106 @@
...
+            List)[Optional[Dictionary._Node]] = [None] * self._capacity
+
+    def _get_index(self, key: Hashable) -> int:
+        return hash(key) % self._capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self._size / self._capacity >= self._load_factor:
+            self._resize()
+        index = self._get_index(key)
+        while self._buckets[index] is not None:
```
Oleksl888, line 29: this while loop repeats in all your get/set/delete methods with slight variations. Maybe place it in separate method and reuse it throughout your code?

app/main.py
```diff
@@ -1,2 +1,106 @@
...
+
+    def __len__(self) -> int:
+        return self._size
+
+    def _resize(self) -> None:
+        new_capacity = self._capacity * 2
+        new_buckets: List[Optional[Dictionary._Node]] = [None] * new_capacity
+        for node in self._buckets:
+            if node is not None:
+                index = node.hash_key % new_capacity
```
Oleksl888, line 53: you can basically reuse setitem method here to rewrite nodes from old table to new one

app/main.py
```diff
@@ -1,2 +1,106 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._get_index(key)
+        while self._buckets[index] is not None:
+            if self._buckets[index].key == key:
+                self._buckets[index] = None
+                self._size -= 1
+                return
+            index = (index + 1) % self._capacity
+        raise KeyError(f'KeyError: "{key}" not found in Dictionary')
```
Oleksl888, line 72: double check if this passes the tests. ( it is commented atm)
One of the ways for this method to work correctly is to recalculate indexes for remaining elements after deletion

app/main.py
```diff
@@ -1,2 +1,106 @@
...
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Optional[Any] = None) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
```
Oleksl888, line 85: Dictionary.pop(key, None) should return None instead of triggering KeyError

Oleksl888 (changes requested):

app/main.py
Oleksl888: try to pass object() as a default value and check against that instead of None

Oleksl888 (approved):
Approved but please check last comment

PR #740, opened 2024-08-12
Solution Politov

Alexandr-Politov:


app/main.py
```diff
@@ -1,2 +1,48 @@
+from typing import Any, Hashable
+
+
+class Node:
```
Oleksl888, line 4: This class better be defined inside of a Dictionary since it is a part of its implementation and will not be used outside of it. Also you should store hash value in it as well.

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+    def __init__(self, key: Hashable, value: Any) -> None:
+        self.key = key
+        self.value = value
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8) -> None:
```
Oleksl888, line 11: capacity shouldnt be parametrized - it's part of dict internal structure

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+        self.__init__(self.capacity * 2)
+        for item in old_hash_table:
+            if item is not None:
+                self.__setitem__(item.key, item.value)
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            if self.size + 1 >= self.capacity * 2 / 3:
```
Oleksl888, line 34: better to use Fraction type here

app/main.py
```diff
@@ -1,2 +1,48 @@
...
+                self.resize()
+                return self.__setitem__(key, value)
```
Oleksl888, line 36: can you rewrite this without this recursive call?  you can simply check if table needs resizing, resize if necessary, then calculate index and set node..

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,42 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
```
Oleksl888, line 2: Why did you delete Node class?

app/main.py
Oleksl888: hasn't been changed

app/main.py
Oleksl888: not done, if you don't need a recursive call why would you use it? recursion can increase complexity of code, need to use it carefully

app/main.py
Oleksl888: not done

app/main.py
```diff
@@ -1,2 +1,42 @@
...
+               and self.hash_table[index][1] != key):
+            index = (index + 1) % self.capacity
+        return index
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.__init__(self.capacity * 2)
+        for item in old_hash_table:
+            if item is not None:
+                self.__setitem__(item[1], item[-1])
```
Oleksl888, line 22: you can use synt sugar interface since magic methods are defined already

Oleksl888 (changes requested):

app/main.py
Oleksl888: Ok so why do you think we should pass capacity as a parameter?

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+
+
 class Dictionary:
-    pass
+
+    class Node:
+
+        def __init__(self, key: Hashable, value: Any) -> None:
+            self.key = key
+            self.value = value
```
Oleksl888, line 11: hash value should be stored here as well

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+               and self.hash_table[index].key != key):
+            index = (index + 1) % self.capacity
+        return index
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.__init__(self.capacity * 2)
+        for item in old_hash_table:
+            if item is not None:
+                self.__setitem__(item.key, item.value)
```
Oleksl888, line 30: you can use [] here since magic method is defined

Oleksl888 (changes requested):

Oleksl888 (approved):
Approved, but please change what we discussed today:
`def __init__(self, capacity: int = 8) -> None:` 
Dict should be a black box, its internal strucutre should not be affected by parameters that come from outside

PR #741, opened 2024-08-12
Solution 1.0

Narberal90:


oliasymak24 (changes requested):

tests/test_main.py
```diff
@@ -6,18 +6,18 @@
...
-#     dictionary = Dictionary()
-#     for key, value in items:
-#         dictionary[key] = value
-#     for key, value in items:
-#         assert dictionary[key] == value
-#     assert len(dictionary) == len(items)
-#     for key, value in items:
-#         del dictionary[key]
-#     print(len(dictionary))
+@pytest.mark.timeout(5)
```
Oleksl888, line 9: where you supposed to make changes to this file?

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+            node_prev = node
+            node = node.next
+
+        raise KeyError(f"Key {key} not found")
+
+    def clear(self) -> None:
+        self.hash_table = {}
+
+
+class Node:
```
Oleksl888, line 82: This class better be defined inside of a Dictionary since it is a part of its implementation and will not be used outside of it

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+    def __init__(self, key: Hashable, value: Any) -> None:
```
Oleksl888, line 83: Node should also store hash value of a key

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+        self.key = key
+        self.value = value
+        self.next = None
```
Oleksl888, line 86: storing `next` node would make sense in a linked list but we don't use it here. You can access next node just by index of a hash table

app/main.py
```diff
@@ -1,2 +1,86 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 0
+        self.hash_table: list = [None] * 8
+        self.load_factor = 2 / 3
```
Oleksl888, line 8: Fraction type would be perfect here

app/main.py
```diff
@@ -1,2 +1,86 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 0
```
Oleksl888, line 6: would make sense to have a capacity attribute here

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+
+    def _reorganize(self) -> None:
+        old_table = self.hash_table
+        new_capacity = len(old_table) * 2
+        self.hash_table = [None] * new_capacity
+        self.length = 0
+
+        for node in old_table:
+            while node:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 18: You can use syntactic sugar since you already defined `__setitem__` in class

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+        if self.length >= int(self.load_factor * len(self.hash_table)):
+            self._reorganize()
+
+        new_node = Node(key, value)
+        key_hash = hash(key)
+        index = key_hash % len(self.hash_table)
+
+        if self.hash_table[index]:
+            node = self.hash_table[index]
+            while node:
```
Oleksl888, line 31: This block of code repeats in few methods - make it DRY

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+            self.hash_table[index] = new_node
+
+        self.length += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
+        key_hash = hash(key)
+        index = key_hash % len(self.hash_table)
+        node = self.hash_table[index]
+        if node is None:
+            raise KeyError("Unknown key")
```
Oleksl888, line 49: Refactor so you don't have to raise error twice

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+                    node_prev.next = node.next
+                self.length -= 1
+                return
+            node_prev = node
+            node = node.next
+
+        raise KeyError(f"Key {key} not found")
+
+    def clear(self) -> None:
+        self.hash_table = {}
```
Oleksl888, line 79: your hash table is a list of nodes, why would you change data type here?

Oleksl888 (changes requested):

tests/test_main.py
Narberal90: this is a block of code from the tests, I don't have to change it. I already had a conversation with Lyudmila in the qna chat

Narberal90: in the condition for the task, it says that if I want to test the delete method, I have to unlock a certain code, so I just unlocked it

app/main.py
Narberal90: wanted to do this in the beginning, but I heard that inner classes are not common...but Of course I will move it

app/main.py
Narberal90: Why?

Oleksl888: Says in task requirements

tests/test_main.py
Oleksl888: I don't think you have to push your local changes for testing to remote

oliasymak24 (approved):

PR #742, opened 2024-08-12
Solution

panichqa:


app/main.py
```diff
@@ -1,2 +1,49 @@
+class Node:
+    def __init__(self, key: object, value: object) -> None:
```
Oleksl888, line 2: This class better be defined inside of a Dictionary since it is a part of its implementation and will not be used outside of it.

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+        self.key = key
+        self.value = value
+        self.hash = hash(key)
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 16, load_factor: float = 0.75) -> None:
```
Oleksl888, line 9: These attrs shouldnt be parametrized - it's part of dict internal structure. Also load factor must be 2/3 - use Fraction for  that

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+        self.capacity = capacity
+        self.size = 0
+        self.load_factor = load_factor
+        self.table = [[] for _ in range(capacity)]
```
Oleksl888, line 13: Better fill out initial table with None values - no need to create 1 million empty lists. Also annotate self.table that it can be a list of Nodes or None values

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+        self.table = [[] for _ in range(self.capacity)]
+        self.size = 0
+        for bucket in old_table:
+            for node in bucket:
+                self[node.key] = node.value
+
+    def _hash_index(self, key: object) -> int:
+        return hash(key) % self.capacity
+
+    def __setitem__(self, key: object, value: object) -> None:
```
Oleksl888, line 27: Key must be a hashable object, value can be anything - fix this annotation in all methods where applicable

app/main.py
```diff
@@ -1,2 +1,49 @@
...
+        if self.size / self.capacity > self.load_factor:
+            self._resize()
+
+        index = self._hash_index(key)
+        bucket = self.table[index]
+        for node in bucket:
```
Oleksl888, line 33: this pattern repeats in both your methods, make it DRY. Also perhaps while loop would be more appropriate here

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,67 @@
+from typing import Any, Hashable, List, Optional
+from fractions import Fraction
+
+
+class Node:
```
Oleksl888, line 5: Move this class inside of Dictionary - since it is part of it internal implementation

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        self.table: List[Optional[List[Node]]] = [None] * self.capacity
+
+    def _resize(self) -> None:
+        old_table = self.table
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+        for bucket in old_table:
+            if bucket:
+                for node in bucket:
```
Oleksl888, line 26: if you do this implementation correctly you will not need to have nested for loops

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+
+        index = self._hash_index(key)
+        if self.table[index] is None:
+            self.table[index] = []
+
+        node = self._find_node(self.table[index], key)
+        if node:
+            node.value = value
+        else:
+            self.table[index].append(Node(key, value))
```
Oleksl888, line 55: you are doing a list that has a list of Nodes - this solution is less effective in terms of complexity. Looking for element will take O(n) and we are looking for the best performance.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+
+    def _hash_index(self, key: Hashable) -> int:
+        return hash(key) % self.capacity
+
+    def _find_node(
+        self,
+        bucket: Optional[List[Node]],
+        key: Hashable
+    ) -> Optional[Node]:
+        if bucket:
```
Oleksl888, line 37: this method should look like so:
- you iterate through hash table in a while loop
- for each element you check hash and key
- if it's correct then you return the Node and exit loop (or maybe finding index would be more correct so you can reuse this method for both get and set)
- if not found - continue going thorugh the table

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+
+    def _resize(self) -> None:
+        old_table = self.table
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+        for bucket in old_table:
+            if bucket:
+                for node in bucket:
+                    self.__setitem__(node.key, node.value)
```
Oleksl888, line 27: you can use syntactic sugar interface because you defined magic methods already

app/main.py
Oleksl888: so you need to have only one list of Nodes. each node must have unique index, collisions must be taken care of and table needs to be resized with recalculation of indexes - review learning materials about how it needs to be done

Oleksl888 (changes requested):

PR #743, opened 2024-08-13
Solution

mihavryliuk:


app/main.py
```diff
@@ -1,2 +1,27 @@
...
+        index = self._hash(key)
+        for kv in self.table[index]:
+            if kv[0] == key:
+                kv[1] = value
+                return
+        self.table[index].append([key, value])
+
+    def __getitem__(self, key: Any) -> Any:
+        index = self._hash(key)
+        for kv in self.table[index]:
```
oliasymak24, line 21: What kv variable name means?

app/main.py
oliasymak24, line 1: Will good to add stuff for printing dictionary in a good way

oliasymak24: Also for deleting items

oliasymak24: Will be better to create separate class Node

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,51 @@
+from dataclasses import dataclass
+from typing import Any, Hashable, Optional
+
+
 class Dictionary:
-    pass
+    @dataclass
+    class Node:
+        key: Hashable
```
Oleksl888, line 8: Hash should be also included here as per task requirements - we would use it later to check if hashes are matching

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+
+    def resize(self) -> None:
+        old_hash_table = self.table
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```
Oleksl888: you can use synt sugar with brackets here since magic method is defined

app/main.py
```diff
@@ -1,2 +1,51 @@
...
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        index = self._find_index(key)
+        if self.table[index] is None:
+            if self.size + 1 >= self.capacity * (2 / 3):
+                self.resize()
+                index = self._find_index(key)
```
Oleksl888: can you avoid a recursive call here?

Oleksl888 (changes requested):

app/main.py
Oleksl888: You might not need to check hash in your implementation but I would still recommend to store the hash in a Node to add an extra check that both key and hash are matching when we are going through the indexes of hash table. Just in case, to be 100% sure that we get correct value.

app/main.py
```diff
@@ -1,2 +1,53 @@
...
+
+
 class Dictionary:
-    pass
+    @dataclass
+    class Node:
+        key: Hashable
+        value: Any
+
+    def __init__(self, capacity: int = 8) -> None:
```
Oleksl888, line 11: Parametrizing internal structure of a dict is not a very good design choice. If you allow devs to play with dict capacity it will behave differently in different parts of program. And also in real dictionary init method takes key-value pairs that will be stored in dictionary on creation

Oleksl888 (approved):
Approved but please check last comments

PR #744, opened 2024-08-13
'first_implementation'

ABaidin:


app/main.py
```diff
@@ -1,2 +1,44 @@
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.load_factor = 2 / 3
```
Oleksl888, line 4: load factor is a constant - you can set it as a class attr + you can use Fraction type to store value

app/main.py
```diff
@@ -1,2 +1,44 @@
...
+        self.size = 0
+        self.hash_table = [[] for _ in range(self.capacity)]
```
Oleksl888, line 6: We actually looking for a different solution for hash table.

Create class Node as per requirements and make a list of Nodes instead of list of lists. Also It is better to fill table with None values on init. It will be more efficient then creating a thousand empty lists.

app/main.py
```diff
@@ -1,2 +1,44 @@
...
+        old_hash_table = self.hash_table
+        self.capacity *= 2
+        self.hash_table = [[] for _ in range(self.capacity)]
+        self.size = 0
+
+        for slot in old_hash_table:
+            if slot:
+                self[slot[0]] = slot[1]
+
+    def _probe(self, key: any, setting: bool = True) -> int:
```
Oleksl888, line 18: think of a better name - such as probe_for_index. Also check type annotation for key - it should be a hashable value

app/main.py
```diff
@@ -1,2 +1,44 @@
...
+        index = hash(key) % self.capacity
+
+        while True:
+            if not self.hash_table[index]:
+                return index if setting else -1
+            if self.hash_table[index][0] == key:
+                return index
+            index = (index + 1) % self.capacity
+
+    def __setitem__(self, key: any, value: any) -> None:
```
Oleksl888, line 28: use correct type annotations everywhere

app/main.py
```diff
@@ -1,2 +1,44 @@
...
+        if self.size > self.capacity * self.load_factor:
+            self._resize()
+
+        index = self._probe(key)
+        if not self.hash_table[index]:
+            self.size += 1
```
Oleksl888, line 34: Overall logic in methods is correct - you just need to refactor it so you use `Nodes` to better structure your code

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,57 @@
+from fractions import Fraction
+from typing import Hashable, Any
+
+
+class Node:
```
Also you should include hash_value as per task requirements

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        self.hash_table: list[Node | None] = [None] * self.capacity
+
+    def _resize(self) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * 2)
+
+        for slot in old_hash_table:
+            if slot:
+                self.__setitem__(slot.key, slot.value)
```

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+
+    def _calculate_index(self, key: Hashable, setting: bool = True) -> int:
```
Oleksl888: why do you need this setting parameter if you never use it in code?

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        index = hash(key) % self.capacity
+
+        while True:
+            if not self.hash_table[index]:
+                return index if setting else -1
```
Oleksl888: it is possible to make this method work without returning -1

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+    def __init__(self, key: Hashable, value: Any) -> None:
+        self.key = key
+        self.value = value
+
+
 class Dictionary:
-    pass
+    LOAD_FACTOR = Fraction(2, 3)
+
+    def __init__(self, capacity: int = 8) -> None:
```

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        index = self._calculate_index(key, False)
+        if index == -1:
+            raise KeyError(f"Key {key} not found")
+        return self.hash_table[index].value
+
+    def __len__(self) -> int:
+        return self.size
+
+    def clear(self) -> None:
+        self.__init__(self.capacity)
```
Oleksl888: Init should be called only once on object creation, it is not recommended to call it multiple times within the class

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job!

PR #745, opened 2024-08-13
implement py-dictionary(first try)

Maksym-Turenko:


app/main.py
```diff
@@ -1,2 +1,56 @@
+from collections.abc import Hashable
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+            initial_capacity: int = 8,
```
Oleksl888, line 8: These attrs shouldnt be parametrized - it's part of dict internal structure. Also use Fraction for load factor and use correct value - 2/3

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+            load_factor: float = 0.75
+    ) -> None:
+        self.load_factor = load_factor
+        self.capacity = initial_capacity
+        self.size = 0
+        self.table = [[] for _ in range(self.capacity)]
```
Oleksl888, line 14: From requirements:

>     Also, don't forget to store (key, hash, value) as a node in the hash table. A hash table may be stored as a simple list of nodes.

Create class Node as per requirements and make a list of Nodes instead of list of lists. Also It is better to fill table with None values on init. It will be more efficient then creating a thousand empty lists.

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+    def _resize(self) -> None:
+        old_table = self.table
+        self.capacity *= 2
+        self.table = [[] for _ in range(self.capacity)]
+        self.size = 0
+
+        for entry in old_table:
+            if entry:
+                for k, v, h in entry:
+                    self.__setitem__(k, v)
```
Oleksl888, line 28: You can use syntactic sugar here since set method is defined in the class

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        self.size += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._hash(key)
+        if self.table[index] is None:
+            raise KeyError(f"Key '{key}' not found")
+
+        for k, v, h in self.table[index]:
+            if k == key:
+                return v
```
Oleksl888, line 51: try not to repeat yourself - you can refactor so KeyError is raised  only once

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size >= self.capacity * self.load_factor:
+            self._resize()
+
+        index = self._hash(key)
+
+        for i, (k, v, h) in enumerate(self.table[index]):
+            if k == key:
+                self.table[index][i] = (key, value, h)
+                return
```
Oleksl888, line 39: Overall logic look correct you just need to refactor code so you use `Nodes` instead of lists to better structure your code

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,70 @@
+from collections.abc import Hashable
+from typing import Any
+
+
+class Node:
```

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+        self.key = key
+        self.value = value
+        self.hash = hash(key)
+        self.next = None
+
+
 class Dictionary:
-    pass
+    INITIAL_CAPACITY = 8
+    LOAD_FACTOR = 2 / 3
```
Oleksl888: Let's present this with Fraction type

app/main.py
```diff
@@ -1,2 +1,70 @@
+from collections.abc import Hashable
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: Hashable, value: Any) -> None:
+        self.key = key
+        self.value = value
+        self.hash = hash(key)
+        self.next = None
```
Oleksl888: You don't really need to store `next` Node. This would make sense in a linked list, but here you can simply check if next index is free or not in case of collision

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+
+    def _resize(self) -> None:
+        old_table = self.table
+        self.capacity *= 2
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_table:
+            while node:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 33: you can use syntactic sugar interface because you defined magic methods already

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+            self._resize()
+
+        index = self._hash(key)
+        node = self.table[index]
+
+        if node is None:
+            self.table[index] = Node(key, value)
+            self.size += 1
+        else:
+            while node:
```
Oleksl888: this block should be placed in a separate method and reused in both get/set methods

Oleksl888 (changes requested):

Oleksl888 (changes requested):
Please change

app/main.py
```diff
@@ -1,2 +1,59 @@
...
+        self.table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
+
+    def _insert(self, key: Hashable, value: Any) -> None:
+        index = self._hash(key)
+        while self.table[index] is not None:
```
Oleksl888, line 37: This block of code repeats also in __getitem__ - make it DRY:
```
while self.table[index] is not None:
            if self.table[index].key == key:
                self.table[index].value = value
                return

Oleksl888 (approved):
Approved but please fix the last comment

PR #746, opened 2024-08-14
solving

Poznanskyi:


app/main.py
```diff
@@ -1,2 +1,84 @@
...
+        if self.hash_table[index] is None:
+            raise KeyError("Cannot find")
+
+        return self.hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if self._calculate_index[index] is None:
+            raise KeyError("Cannot find")
```
oliasymak24, line 69: Cannot find what?

app/main.py
```diff
@@ -1,2 +1,84 @@
...
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * CAPACITY_MULTIPLIER)
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
+
+    def __setitem__(self, key: float, value: float) -> None:
```
oliasymak24, line 46: Bad annotation for key, check everywhere

oliasymak24 (changes requested):
Try to run tests, it is failed, check  it

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+
+    class Node:
+
+        def __init__(self, key: Hashable, value: Any, hash_key: int) -> None:
+            self.key = key
+            self.value = value
+            self.hash_key = hash_key
+
+    def __init__(self) -> None:
+        self.load_factor = 0.667
```
Oleksl888: fraction type would be perfect here to present 2/3

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        self.number_of_elements = 0
+        self.capacity = 8
+        self.size = 0
+        self.hash_table: list[tuple | None] = [None] * self.capacity
```
Oleksl888: check annotation - hash table is not a list of tuples

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+    def resize(self) -> None:
+        self.capacity = self.capacity * 2
+        self.size = 0
+        copy_arr: list = deepcopy(self.hash_table)
+        self.hash_table: list[None | Dictionary.Node] = [
+            None for _ in range(self.capacity)]
+
+        for values in copy_arr:
+            if values is not None:
+                self.__setitem__(values.key, values.value)
```

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+    def __len__(self) -> int:
+        return self.size
+
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def resize(self) -> None:
+        self.capacity = self.capacity * 2
+        self.size = 0
+        copy_arr: list = deepcopy(self.hash_table)
```
Oleksl888: you can create an empty hash table here:
`new_table: list[Dictionary.Node | None] = [None] * self.capacity`
then reset it with values from old table in a loop and then simply reassign self.hash_table to new_table

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def resize(self) -> None:
+        self.capacity = self.capacity * 2
+        self.size = 0
+        new_table: list[None | Dictionary.Node] = [None] * self.capacity
+
+        new_table: list[None | Dictionary.Node] = [None] * self.capacity
```
Oleksl888: why this is duplicated?

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+
+        old_table = self.hash_table
+        self.hash_table = new_table
+
+        for bucket in old_table:
+            if bucket is not None:
```
Oleksl888: here can be simplified - you don't need nested for loop here and many if conditions. Just check if bucket is a Node and set it if yes, else do nothing

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+
+    class Node:
+
+        def __init__(self, key: Hashable, value: Any, hash_key: int) -> None:
+            self.key = key
+            self.value = value
+            self.hash_key = hash_key
+
+    def __init__(self) -> None:
+        self.load_factor = 2 / 3
```
Oleksl888: still would make sense to use Fraction here to be more precise

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+            self.key = key
+            self.value = value
+            self.hash_key = hash_key
+
+    def __init__(self) -> None:
+        self.load_factor = Fraction(2, 3)
+        self.number_of_elements = 0
+        self.capacity = 8
+        self.size = 0
+        self.hash_table = [None] * self.capacity
```
Oleksl888, line 19: `self.hash_table: list[Dictionary.Node | None] = [None] * self.capacity` - adding annotation here will make type hints available in the code below

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+    class Node:
+
+        def __init__(self, key: Hashable, value: Any, hash_key: int) -> None:
+            self.key = key
+            self.value = value
+            self.hash_key = hash_key
+
+    def __init__(self) -> None:
+        self.load_factor = Fraction(2, 3)
+        self.number_of_elements = 0
```
Oleksl888, line 16: This is not used anywhere

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+    def resize(self) -> None:
+        self.capacity = self.capacity * 2
+        self.size = 0
+        new_table: list[None | Dictionary.Node] = [None] * self.capacity
+
+        old_table = self.hash_table
+        self.hash_table = new_table
+
+        for bucket in old_table:
+            if isinstance(bucket, list):
```
Oleksl888, line 59: Should be more simple - something like this:
```
  for node in self.hash_table:
      if isinstance(node, Dictionary.Node):
          index = self.find_index(node.key)
          new_table[index] = node

Oleksl888 (approved):
Approved - but please check last comments

PR #747, opened 2024-08-14
Solution

pashawarganov:


app/main.py
```diff
@@ -1,2 +1,97 @@
...
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * CAPACITY_MULTIPLIER)
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
oliasymak24, line 47: The correct type annotation for key is Hashable.

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def clear(self) -> None:
+        self.__init__()
+
+    def pop(self, key: Any) -> Any:
+        value = self[key]
+        self.__delitem__(key)
+        return value
```
oliasymak24, line 93: The pop() method returns:

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+
+    def update(self, **kwargs) -> None:
```
oliasymak24, line 95: Check this method

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        return self.hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
```
oliasymak24, line 70: Will be better to say that cannot find a key

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,118 @@
+from __future__ import annotations
+from dataclasses import dataclass
+from typing import Hashable, Any, Iterable
+
+INITIAL_CAPACITY = 8
+RESIZE_THRESHOLD = 2 / 3
+CAPACITY_MULTIPLIER = 2
+
+
+@dataclass
```
Also you should include hash_value as per task requirements

app/main.py
```diff
@@ -1,2 +1,118 @@
+from __future__ import annotations
+from dataclasses import dataclass
+from typing import Hashable, Any, Iterable
+
+INITIAL_CAPACITY = 8
```
Oleksl888: These constants should be defined as class attrs since it is a part of Dict implementation and will not be used outside of it

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+RESIZE_THRESHOLD = 2 / 3
```
Oleksl888: Fraction would be perfect for representing 2/3

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+@dataclass
+class Node:
+    key: Hashable
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = INITIAL_CAPACITY) -> None:
```

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        return index
+
+    @property
+    def current_max_size(self) -> int:
+        return round(self.capacity * RESIZE_THRESHOLD)
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * CAPACITY_MULTIPLIER)
```
Oleksl888: Init should be called only once on object creation, it is not recommended to call it multiple times within the class

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            if self.size + 1 >= self.current_max_size:
+                self.resize()
+                return self.__setitem__(key, value)
```

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        return self.hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
```
Oleksl888, line 81: you need to recalculate remaining positions after deleting an element because indexes may be wrong for elements that had collisions

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+    def get(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888: annotate default as optional parameter

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+    def __len__(self) -> int:
+        return self.size
+
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def clear(self) -> None:
+        self.__init__()
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888: annotate default as optional parameter

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        try:
+            value = self[key]
+        except KeyError as e:
+            if default:
+                return default
+            raise e
```
Oleksl888: Dictionary.pop(key, None) should return None if key is not in dict. KeyError must be raised only when default value is not passed

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        self.__delitem__(key)
+        return value
+
+    def update(self, other: Dictionary | Iterable = None, **kwargs) -> None:
+        if not other:
+            for key, value in kwargs.items():
```
Oleksl888, line 114: you don;t need to set anything when nothing is passed to this method - just return nothing in this case

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+                self.__setitem__(key, value)
+        elif isinstance(other, Dictionary):
+            for node in other.hash_table:
+                if node is not None:
+                    self.__setitem__(node.key, node.value)
```
Oleksl888: use sugar interface here

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        elif isinstance(other, dict):
+            for key, value in other.items():
+                self.__setitem__(key, value)
+        elif isinstance(other, Iterable):
```
Oleksl888, line 123: so we can get an iterable of key-value pairs, then we can simply run it in a for-loop and unpack them, no need to wrap it in try except - let the error happen if wrong data has been passed:
```
            for key, value in other:
                self[key] = value

Oleksl888 (changes requested):

app/main.py
pashawarganov: I changed it but what if we have wrong Iterable object. Or it`s ok if python raise 'ValueError: not enough values to unpack (expected 2, got 1)'?

Oleksl888: Yeah in this case is ok, python will not allow unpacking operation or will not be able to iterate over wrong data type.
You could add custom exception of course but this method has already too much going on

Oleksl888 (approved):
Well done sir! :clap:

PR #749, opened 2024-08-14
solution

HalynaRohatska:


app/main.py
```diff
@@ -1,2 +1,97 @@
...
+        return str(self.hash_table)
+
+    def clear(self) -> None:
+        self.__init__(INITIAL_CAPACITY)
+
+    def pop(self, key: Hashable) -> Any:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
```
oliasymak24, line 92: Will be better to say that cannot find a key

app/main.py
```diff
@@ -1,2 +1,97 @@
...
+    def __len__(self) -> int:
+        return self.size
+
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def clear(self) -> None:
+        self.__init__(INITIAL_CAPACITY)
+
+    def pop(self, key: Hashable) -> Any:
```
oliasymak24, line 88: The pop() method returns:

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,99 @@
+from dataclasses import dataclass
+from typing import Hashable, Any
+
+
+INITIAL_CAPACITY = 8
+RESIZE_THRESHOLD = 2 / 3
+CAPACITY_MULTIPLIER = 2
```
Oleksl888, line 7: You use this varibable only once. Do you really need it?
Also these constants belong to Dictionary implementation so they should be defined as class attrs

app/main.py
```diff
@@ -1,2 +1,99 @@
+from dataclasses import dataclass
+from typing import Hashable, Any
+
+
+INITIAL_CAPACITY = 8
+RESIZE_THRESHOLD = 2 / 3
```
Oleksl888, line 6: Fraction type is perfect for this scenario

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+
+@dataclass
+class Node:
+    key: Hashable
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = INITIAL_CAPACITY) -> None:
```
Oleksl888, line 17: capacity shouldnt be parametrized - it's part of dict internal structure

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+
+        while (
+                self.hash_table[index] is not None
+                and self.hash_table[index].key != key
+        ):
+            index = (index + 1) % self.capacity
+        return index
+
+    @property
+    def current_max_size(self) -> float:
```
Oleksl888, line 33: how is it float?

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
+
+    def get(self, key: Hashable, default: Any = None) -> Hashable:
```
Oleksl888, line 73: you should annotate default as optional parameter. Also check return type and correct

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+    def __len__(self) -> int:
+        return self.size
+
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def clear(self) -> None:
+        self.__init__(INITIAL_CAPACITY)
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
```
Oleksl888, line 88: you should annotate default as optional parameter

app/main.py
```diff
@@ -1,2 +1,99 @@
...
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            if default is not None:
+                return default
+            raise KeyError(f"Cannot find a key: {key}")
+
+        removed = self.hash_table[index].value
```
Oleksl888, line 96: why not reuse getitem and delitem methods here? of course using syntactic sugar interface

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+
+    @dataclass
+    class Node:
+        key: Hashable
+        value: Any
+        hash_value: int
+
+    def __init__(
+            self,
+            capacity: int = _INITIAL_CAPACITY
```
Oleksl888, line 19: I still don't agree that we should pass capacity as a parameter. Dict can take an iterable on init that will set values on creation, but not parameters that will affect its internal organization

Oleksl888, line 19: I still don't agree that we should pass capacity as a parameter. Dict can take an iterable on init that will set values on creation, but not parameters that will affect its internal organization

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.capacity *= self.capacity_multyplier
+        self.size = 0
+        self.hash_table = [None] * self.capacity
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 49: you can use syntactic sugar because you defined magic methods already

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        hash_value = hash(key)
+        index = self._calculate_index(key, hash_value)
+
+        if self.hash_table[index] is None:
+            if self.size + 1 >= self.current_max_size:
+                self.resize()
+                return self.__setitem__(key, value)
```
Oleksl888, line 58: refactor this so you don't have to make a recursive call. 
Check if table needs resizing first, resize if necessary, then calculate index and set node.

Oleksl888 (changes requested):

app/main.py
Oleksl888: I still think you can do it without recursion, you can even make it more simple with less if statements. Recursion can be dangerous and it makes your code harder to debug

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        hash_value = hash(key)
+        index = self._calculate_index(key, hash_value)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
```
Oleksl888, line 80: you need to recalculate remaining positions after deleting an element because indexes may be wrong for elements that had collisions

app/main.py
```diff
@@ -1,2 +1,105 @@
...
+    def clear(self) -> None:
+        self.__init__()
+
+    def pop(self, key: Hashable, default: Optional[Any] = None) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            if default is not None:
```
Oleksl888, line 103: Dictionary(key, None) should return None if key is not in dict. KeyError must be raised only when default value is not passed

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+        for node in old_hash_table:
+            if node is not None:
+                self[node.key] = node.value
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        hash_value = hash(key)
+        index = self._calculate_index(key, hash_value)
+
+        while self.hash_table[index] is not None:
```
Oleksl888, line 52: why not reuse your def _calculate_index method here?

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        return self.hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        hash_value = hash(key)
+        index = self._calculate_index(key, hash_value)
+
+        while True:
```
Oleksl888, line 77: here also - calculate index should be used to keep it DRY

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+            if self.hash_table[index] is None:
+                raise KeyError(f"Cannot find value for key: {key}")
+            elif self.hash_table[index].key == key:
+                break
+            index = (index + 1) % self.current_max_size
+
+        self.hash_table[index] = None
+        self.size -= 1
+
+        next_index = (index + 1) % self.current_max_size
```
Oleksl888, line 87: You can fix it by calling resize - but it shouldnt actually resize the table - it must recalculate remaining indexes for elements without using the multiplier. So you could refactor your resize method so it accepts a multiplier value and uses different one when called on delete and when called on set methods

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+            return default
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def clear(self) -> None:
+        self.__init__()
```
Oleksl888, line 108: If you call init it will reduce your list capacity to original - 8 cells which is not very accurate. __init__ is usually called once on object creation, and it is not recommended to call it many times

app/main.py
```diff
@@ -1,2 +1,118 @@
...
+
+    def pop(self, key: Hashable, default: Optional[Any] = None) -> Any:
```
Oleksl888, line 110: try to pass `object()` as a default value and check against that.

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+        hash_value = hash(key)
+        index = self._calculate_index(key, hash_value)
+
+        if self.hash_table[index] is not None:
+            self.hash_table[index].value = value
+        else:
+            self.hash_table[index] = Dictionary.Node(key, value, hash_value)
+            self.size += 1
+
+        if self.size >= self.current_max_size:
```
Oleksl888, line 61: In dictionary this actually is being checked before you add new element

app/main.py
```diff
@@ -1,2 +1,112 @@
...
+
+    def __str__(self) -> str:
+        return str(self.hash_table)
+
+    def clear(self) -> None:
+        self.hash_table = [None] * self.capacity
+        self.size = 0
+
+    def pop(self, key: Hashable, default: Optional[Any] = object()) -> Any:
+        hash_value = hash(key)
```
Oleksl888, line 102: you could reuse previously defined mehods here:
```
    def pop(self, key: Hashable, default: Optional[Any] = object) -> Any:
        try:
            value = self[key]
        except KeyError as e:
            if default is object:
                raise e
            return default
        del key
        return value

Oleksl888 (approved):
well done!

PR #750, opened 2024-08-14
Solution

Bogdan-Rylov:


oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,148 @@
...
+
+        for node in previous_hash_table:
+            if node:
+                self[node.key] = node.value
+
+    def clear(self) -> None:
+        self._size = 0
+        self._hash_table = [None] * self._capacity
+
+    def get(self, key: Hashable, default_value: Any = None) -> Any:
```
Oleksl888, line 110: annotate that default_value is optional

app/main.py
```diff
@@ -1,2 +1,148 @@
...
+        ]
+
+    def items(self) -> list[tuple[Hashable, Any]]:
+        return [
+            (item.key, item.value)
+            for item in self._hash_table
+            if isinstance(item, Dictionary.Node)
+        ]
+
+    def update(self, other: Dictionary = None) -> None:
```
Oleksl888, line 145: other can be also an iterable of key-value pairs

app/main.py
```diff
@@ -1,2 +1,148 @@
...
+        self._size = 0
+        self._hash_table = [None] * self._capacity
+
+    def get(self, key: Hashable, default_value: Any = None) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            return default_value
+
+    def pop(self, key: Hashable, default_value: Any = None) -> Any:
```
Oleksl888, line 116: pop method works a little different it will trigger a keyerror if key is not in dict and default values is not set

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,148 @@
...
 class Dictionary:
-    pass
+    _DEFAULT_INITIAL_CAPACITY = 8
+    _DEFAULT_LOAD_FACTOR = 2 / 3
+    _DEFAULT_CAPACITY_MULTIPLIER = 2
+
+    @dataclass
+    class Node:
+        key: Hashable
+        value: Any
```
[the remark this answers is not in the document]
Bogdan-Rylov: In my implementation, the `Node.hash` isn't really used throughout the code. The hash is always calculated inside the `_calculate_index` method depending on the `key`, and it is unknown if this is a key of existing Node with calculated hash or a new one. I decided that it's redundant to store it as an object attribute. But ok, I added it, so the requirements of the condition are met.

app/main.py
```diff
@@ -1,2 +1,148 @@
...
+
+    def __init__(
+            self,
+            initial_capacity: int = _DEFAULT_INITIAL_CAPACITY,
```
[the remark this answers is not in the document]
Bogdan-Rylov: Actually, I think it makes sense since the name of the task is "custom dict" (what if I need a dictionary with a load factor 1/2 instead of the 2/3, and next time I'll need it with capacity multiplier 3 to satisfy my custom requirements?).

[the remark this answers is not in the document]
Oleksl888: This task is more about to learn about dict internal structure, there are no strict guidelines about what should be in `__init__` method. What I'm trying to say is that parametrizing such an important data structure will make it unstable and unpredictable.

app/main.py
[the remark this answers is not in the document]
Oleksl888: I would store the hash in a Node to add an extra check that both key and hash are matching when we are going through the indexes of hash table. Just in case, to be 100% sure that we get correct value. It is true though that you might not need to check hash in your implementation

app/main.py
```diff
@@ -1,2 +1,163 @@
...
+            self._hash_table[index].value = value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if not self._hash_table[index]:
+            raise KeyError(f"'{key}'")
+
+        self._hash_table[index] = None
+        self._size -= 1
```
Oleksl888, line 69: In case if there were collisions some indexes lead to incorrect Nodes after a Node is deleted. It can be fixed by recalculating indexes for remaining elements of hash table after deletion. 
You can uncomment a test case for delitem and check if your implementation passes

app/main.py
```diff
@@ -1,2 +1,163 @@
...
+
+    def __iter__(self) -> Dictionary:
+        def generator() -> Hashable:
```
Oleksl888: If you use generators this expression will return a generator-iterator already with build in next method, Stopiteration etc.
```
            for item in self._hash_table:
                if item:
                    yield item.key

Comment (Oleksl888) [reply app/main.py]: or you can code it as generator expression like so:
`(item.key for item in self._hash_table if item)`

File: app/main.py [hunk 8]
```diff
@@ -1,2 +1,163 @@
...
+    def __len__(self) -> int:
+        return self._size
+
+    def __str__(self) -> str:
+        return str(self._hash_table)
+
+    def _resize_hash_table(self) -> None:
+        previous_hash_table = self._hash_table
+
+        self.__init__(
```
Oleksl888: __init__  is being called once on object creation. Reusing it multpile times is generally considered an antipattern

app/main.py
```diff
@@ -1,2 +1,163 @@
...
+
+    def get(self, key: Hashable, default_value: Optional[Any] = None) -> Any:
+        try:
+            return self[key]
+        except KeyError:
+            if not default_value:
+                raise
+            return default_value
+
+    def pop(self, key: Hashable, default_value: Optional[Any] = None) -> Any:
```
Oleksl888: pop method should have a bit different behaviour:
If you use Dictionary.pop(key, None) it should not trigger error but return None (in case when key is not found of course).
Try using object() as default value and check for that in your except clause

app/main.py
```diff
@@ -1,2 +1,163 @@
...
+        ]
+
+    def update(
+            self,
+            other: Dictionary | Iterable[Hashable, Any] = None
+    ) -> None:
+        if isinstance(other, Dictionary):
+            for key, value in other.items():
+                self[key] = value
+        elif isinstance(other, Iterable):
```
Oleksl888: Here you can omit this check. Try to unpack values like you did below and python will throw a type or attribute error by itself if there is a wrong type

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,167 @@
...
+
+    _sentinel = object()
+
+    @dataclass
+    class Node:
+        key: Hashable
+        key_hash: int
+        value: Any
+
+    def __init__(self, ) -> None:
```
Oleksl888: trailing comma here is unnecessary

app/main.py
```diff
@@ -1,2 +1,167 @@
...
+
+        if not self._hash_table[index]:
+            raise KeyError(f"'{key}'")
+
+        self._hash_table[index] = None
+        self._size -= 1
+
+        next_index = (index + 1) % self._capacity
+        while self._hash_table[next_index] is not None:
+            item = self._hash_table[next_index]
```
Oleksl888, line 73: making nested while loops is not very optimal here - you could reuse part of your functionality in  _resize_hash_table to make it work. Basically if you parametrize multiplier in resize method and call it with multiplier of 1 from here it will reset all the indexes correctly

Oleksl888 (approved):
Approved! But please check the last comment

PR #752, opened 2024-08-16
Solution

AndriiHamasa:


app/main.py
```diff
@@ -1,2 +1,82 @@
...
+        return self.hash_table[index].value
+
+    def __len__(self) -> int:
+        return self.length
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"No such element with key: '{key}'")
```
oliasymak24, line 73: Will be better to say that cannot find a key

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+        self.hash_table = [None] * self.capacity
+
+    def calculate_index(self, key: Hashable) -> int:
+        index = hash(key) % self.capacity
+
+        while (
+            self.hash_table[index] is not None
+            and self.hash_table[index].key != key
+        ):
+
```
oliasymak24, line 28: No need for this empty line

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,82 @@
+from dataclasses import dataclass
+from typing import Hashable, Any
+
+CAPACITY = 8
```
Oleksl888, line 4: These constants as well as Node class a part of dict internal implementation so it would make sense to package everything inside of Dictionary class

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+INCREASE_KOEF = 2
```
Oleksl888, line 5: think of a better name - capacity multiplier for example

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+TO_INCREASE = 2 / 3
```
Oleksl888, line 6: load factor would be a better name + you can use Fraction type here

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+
+
+@dataclass()
```
Oleksl888, line 9: why you have parenthesis here?

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+class Node:
+    key: Hashable
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = CAPACITY) -> None:
```
Oleksl888, line 16: capacity shouldnt be parametrized - it's part of dict internal structure

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+        self.capacity = capacity
+        self.length = 0
+        self.hash_table = [None] * self.capacity
```
Oleksl888, line 19: would make sense to annotate hash table as a list of Nodes

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+
+            index = (index + 1) % self.capacity
+
+        return index
+
+    def rehash(self, key: Hashable, value: Any) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * 2)
+        self.__setitem__(key, value)
```
Oleksl888, line 37: you can use syntactic sugar since magic methods are already defined

app/point.py
```diff
@@ -19,6 +19,9 @@ def __hash__(self) -> int:
         # in the hashtable as well
         return hash((self.x, self.y))
 
+    def __repr__(self) -> str:
```
Oleksl888, line 22: no need to push these changes to remote

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+
+        for obj in old_hash_table:
+            if obj is not None:
+                self.__setitem__(obj.key, obj.value)
+
+    @property
+    def max_size(self) -> int:
+        return round(self.capacity * TO_INCREASE)
```
Oleksl888, line 45: why do you use round function here?

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #753, opened 2024-08-16
Solution

mnezbrytska:


app/main.py
```diff
@@ -1,2 +1,102 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: Any, value: Any, next_node: Any = None) -> None:
```
Oleksl888, line 5: This class better be defined inside of a Dictionary since it is a part of its implementation and will not be used outside of it. Also use correct type annotations.
I see that you are implementing a link list to avoid collisions but this will not be the most effective approach in terms of performance - insertion will be fast but retrieval of data will be O(n) and we are aiming for O(1). Try to follow the example in the case study

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+        self.key = key
+        self.value = value
+        self.next = next_node
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self, initial_capacity: int = 8,
```
Oleksl888, line 13: These attrs shouldnt be parametrized - it's part of dict internal structure. Also use Fraction for load factor

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+            load_factor: float = 2 / 3
+    ) -> None:
+        self.capacity = initial_capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.table = [None] * self.capacity
```
Oleksl888, line 19: add annotation to table that it is a list of Nodes or None values

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        index = self._hash_function(key) % self.capacity
+        node = self.table[index]
+
+        while node:
```
Oleksl888, line 25: You can put index lookup in a separate method and reuse it in this and other methods to keep it DRY

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+            if node.key == key:
+                node.value = value
+                return
+            node = node.next
+
+        new_node = Node(key, value, self.table[index])
+        self.table[index] = new_node
+        self.size += 1
+
+        if self.size >= self.capacity * self.load_factor:
```
Oleksl888, line 35: Would be more correct to resize before inserting an element

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+            prev_node = node
+            node = node.next
+
+        raise KeyError(f"Key {key} not found")
+
+    def __len__(self) -> int:
+        return self.size
+
+    def _hash_function(self, key: Any) -> int:
+        return hash(key)
```
Oleksl888, line 71: No need to make an extra method for it since you only call `hash()` and nothing else

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+        for node in old_table:
+            while node:
+                self[node.key] = node.value
+                node = node.next
+
+    def clear(self) -> None:
+        self.table = [None] * self.capacity
+        self.size = 0
+
+    def get(self, key: Any, default: Any = None) -> Any:
```
Oleksl888, line 90: annotate that default is optional
key should be of Hashable type - check everywhere and correct

app/main.py
```diff
@@ -1,2 +1,102 @@
...
+        except KeyError:
+            return default
+
+    def pop(self, key: Any, default: Any = None) -> Any:
+        try:
+            value = self[key]
+            del self[key]
+            return value
+        except KeyError:
+            return default
```
Oleksl888, line 102: you should raise an error in case if key is not present and default is not set

app/point.py
```diff
@@ -12,12 +12,7 @@ def __eq__(self, other: Any) -> bool:
...
     def __hash__(self) -> int:
-        # Change the implementation of the hash to debug your code.
-        # For example, you can return self.x + self.y as a hash
-        # which is NOT a best practice, but you will be able to predict
-        # a hash value by coordinates of the point and its index
-        # in the hashtable as well
-        return hash((self.x, self.y))
+        return int(self.x + self.y)
 
     @property
```
Oleksl888, line 17: revert these changes on remote

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+            self._resize()
+
+        hash_value: int = hash(key)
+        index: int = self._index(hash_value)
+
+        if not self.table[index]:
+            self.table[index] = self.Node(key, hash_value, value)
+        else:
+            node: Optional[Dictionary.Node] = self.table[index]
+            while node:
```
Oleksl888, line 31: you could make implementation without `next` Nodes which would simplify the whole thing.
So you could have one method to calculate index - it would go through the elements of a hash table and get the index of one that is not None and has matching hash and key in case if you want to retrieve. OR would set an element when it finds a suitable empty cell or update an element if key and hash is matching.. We could discuss more on QNA if you want or double check the implementation that Yuri was showing on the lecture

Oleksl888 (approved):
Approved - it works, but you can make more efficient implemention

PR #754, opened 2024-08-16
task done

Sergi0bbb:


app/main.py
```diff
@@ -1,2 +1,52 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8) -> None:
+        self.capacity = capacity
+        self.size = 0
+        self.hash_table = [[] for _ in range(self.capacity)]
```
Oleksl888, line 8: Better fill out initial table with None values - no need to create 1 million empty lists.

app/main.py
```diff
@@ -1,2 +1,52 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8) -> None:
```
Oleksl888, line 5: capacity shouldnt be parametrized - it's part of dict internal structure.

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        self.capacity = capacity
+        self.size = 0
+        self.hash_table = [[] for _ in range(self.capacity)]
+
```
Oleksl888, line 9: > Also, don't forget to store (key, hash, value) as a node in the hash table. A hash table may be stored as a simple list of nodes.

You should create a Node to better structure your code

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+    def calculate_hash(self, key: Any) -> int:
+        return hash(key) % self.capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        index = self.calculate_hash(key)
+        container = self.hash_table[index]
```
Oleksl888, line 15: I don't see table resizing implemented and proper handling of collisions

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+        for index, (key_in_container, _) in enumerate(container):
+            if key_in_container == key:
+                container[index] = (key, value)
+                return
+
+        container.append((key, value))
```
Oleksl888, line 22: You should change the entire concept of this. Storing values in a list of lists makes your code very slow and we are trying to achieve maximum performance. Try to follow the implementation explained in course materials.

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,70 @@
...
-    pass
+    class Node:
+        def __init__(self, key: Hashable, value: Any) -> None:
+            self.key = key
+            self.value = value
+
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.load_factor_threshold = 2 / 3
```
Oleksl888: This can be presented by Fraction type

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+        self.hash_table: list[tuple | None] = [None] * self.capacity
```
Oleksl888: it is a list of Nodes not tuples

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.calculate_hash(key)
+
+        if self.hash_table[index] is None:
+            if self.size + 1 >= self.current_max_size:
+                self.resize()
+                return self.__setitem__(key, value)
```

app/main.py
```diff
@@ -1,2 +1,70 @@
...
+        self.hash_table = [None] * self.capacity
+        self.size = 0
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.calculate_hash(key)
+        if self.hash_table[index] is None:
+            raise KeyError(f"No value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
```
Oleksl888, line 71: you need to recalculate remaining positions after deleting an element because indexes may be wrong for elements that had collisions

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+
+        for values in self.hash_table:
```
Oleksl888, line 73: I guess this will pass the tests but it is actually not a good practice to iterate through a list and change values in the same list. Also you may use brackets inteface when setting the node here as well

app/main.py
```diff
@@ -1,2 +1,75 @@
+from fractions import Fraction
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    class Node:
+        def __init__(self, key: Hashable, value: Any) -> None:
+            self.key = key
+            self.value = value
```
Oleksl888, line 9: would make sense to keep hash from key here as well and check if we have a match in later functions

Oleksl888 (approved):
Well done! But check the last comments

PR #755, opened 2024-08-18
Task Solution

achalaidiuk:


app/main.py
oliasymak24, line 1: Implement separate Node class

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,77 @@
+from typing import Hashable, Any
+
+DEFAULT_CAPACITY = 8
```
Oleksl888, line 3: theis should be class variables inside of Dictionary since it is part of it implementation

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+
+
+class Node:
```
Oleksl888, line 6: Hash should be also included here + would make sense to declare this class inside of Dictionary since it is part of it implementation

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+    def __init__(self, key: Hashable, value: Any) -> None:
+        self.key = key
+        self.value = value
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = DEFAULT_CAPACITY
+        self.hash_table = [None] * self.capacity
```
Oleksl888, line 15: annotate types inside of hash table

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+        return self.number_of_elements
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.number_of_elements >= self.capacity * 0.75:
+            self.resize()
+
+        hash_key = hash(key)
+        index_to_past = self.find_available_case(key, hash_key)
+
+        while (
```
Oleksl888, line 28: this block of code repeats in both set and get method + you have find_available_case method. Separate logic of finding correct index in a separate method and reuse it to keep your code DRY.

app/main.py
```diff
@@ -1,2 +1,77 @@
...
+        self.number_of_elements = 0
+
+        for node in old_cases:
+            if node:
+                self[node.key] = node.value
+
+    def find_available_case(self, key: Hashable, hash_key: int) -> int:
+        index_available_case = self.get_index_by_hash(hash_key)
+        while (
+                self.hash_table[index_available_case] is not None
```
Oleksl888, line 67: You must check that key and hash value of a Node equals to one that is being looked, right now you only check if key is the same

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #756, opened 2024-08-18
Solution

dpiuro:


app/main.py
```diff
@@ -1,2 +1,129 @@
...
+
+# resize
+    for i in range(5):
+        custom_dict[i] = i
+    print(len(custom_dict))
+    print("capacity:", custom_dict.capacity)
+
+    custom_dict[5] = 5
+    print(len(custom_dict))
+    print("capacity:", custom_dict.capacity)
```
oliasymak24, line 129: Allllll these stuff you need to delete...

app/main.py
```diff
@@ -1,2 +1,129 @@
+from dataclasses import dataclass
+from typing import Hashable, Any
+
+from app.point import Point
```
oliasymak24, line 4: You don't need this import here

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,98 @@
+from dataclasses import dataclass
+from typing import Hashable, Any
+
+
+INITIAL_CAPACITY = 8
```
Oleksl888, line 5: These constants and Node class should be defined  inside of a Dictionary since it is a part of its implementation and cannot be used outside it

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+RESIZE_THRESHOLD = 0.62
+CAPACITY_MULTIPLIER = 2
+
+
+@dataclass
+class Node:
+    key: Hashable
```
Oleksl888, line 12: Node should also store a hash_value

app/main.py
```diff
@@ -1,2 +1,98 @@
+from dataclasses import dataclass
+from typing import Hashable, Any
+
+
+INITIAL_CAPACITY = 8
+RESIZE_THRESHOLD = 0.62
```
Oleksl888, line 6: You can accurately present 2/3 with Fraction type

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+class Node:
+    key: Hashable
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(
+            self,
+            capacity:
```
Oleksl888, line 19: You shouldn't pass these params on init - they are part of dict internal structure

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+
+    def _resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.capacity *= CAPACITY_MULTIPLIER
+        self.hash_table = [None] * self.capacity
+        self.size = 0
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 46: you can use syntactic sugar since this method is already defined

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+
+    def __delitem__(
+            self,
+            key: Hashable
+    ) -> None:
+        index = self._calculate_index(key)
+        if self.hash_table[index] is None:
+            raise KeyError(f"Key {key} not found.")
+        self.hash_table[index] = None
+        self.size -= 1
```
Oleksl888, line 78: you need to recalculate remaining positions after deleting an element because indexes may be wrong for elements that had collisions

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+
+    def get(
+            self,
+            key: Hashable,
+            default: Any = None
```
Oleksl888, line 83: annotate default as optional param

app/main.py
```diff
@@ -1,2 +1,98 @@
...
+    ) -> Any:
+        try:
+            return self.__getitem__(key)
```
Oleksl888, line 86: use sugar interface instead of calling magic direcly

app/point.py
```diff
@@ -12,12 +12,7 @@ def __eq__(self, other: Any) -> bool:
         return self.x == other.x and self.y == other.y
 
     def __hash__(self) -> int:
```
Oleksl888, line 14: revert these changes

app/dict_test.py
```diff
@@ -0,0 +1,13 @@
+from main import Dictionary
+
+
+if __name__ == "__main__":
```
Oleksl888, line 4: remove unnecesary files

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #758, opened 2024-08-18
Solution

Ivan-Shakhman:
Solution (i hope.....)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from fractions import Fraction
-from typing import Hashable, Any, Iterator
+from typing import Hashable, Any, Iterator, Optional
 
 
 class Dictionary:
@@ -80,13 +80,15 @@ class Dictionary:
     def clear(self) -> None:
         self.__init__()
 
-    def pop(self, key: Hashable) -> Any | None:
+    def pop(self, key: Hashable, default: Optional[Any]) -> Any | None:
         temp = None
         index = self._calculate_index(key)
         if self.hash_table[index]:
             temp = self.hash_table[index].value
             self.hash_table[index] = None
-        return temp
+        if temp:
+            return temp
+        return default if default else temp
 
     def keys(self) -> list[Hashable] | None:
         keys = []
@@ -104,12 +106,15 @@ class Dictionary:
 
     def items(self) -> list[tuple[Hashable, Any]] | None:
         items_ = []
-        for key, value in self.values(), self.items():
-            if key and value:
+        for key, value in self.keys(), self.values():
+            if key:
                 items_.append((key, value))
         return items_ if items_ else None
 
-    def update(self, other: object) -> None:
+    def update(self, other: Optional[object]) -> None:
+        if not other:
+            return
+
         if isinstance(other, Dictionary) or type(other) is dict:
             for key, value in other.items():
                 self.hash_table[key] = value
--- a/app/main.py
+++ b/app/main.py
@@ -63,7 +63,7 @@ class Dictionary:
         index = self._calculate_index(key)
 
         if self.hash_table[index] is None:
-            raise KeyError(f'Cannot find value for key: {key}')
+            raise KeyError(f"Cannot find value for key: {key}")
 
         self.hash_table[index] = None
         self.size -= 1
@@ -111,4 +111,4 @@ class Dictionary:
             for key, value in other.items():
                 self.hash_table[key] = value
         else:
-            raise TypeError("Update data must be 'Dictionary'!")
+            raise TypeError("Update data must be Dictionary!")
--- a/app/main.py
+++ b/app/main.py
@@ -81,7 +81,6 @@ class Dictionary:
             print(f"Error! Value with key {key} does not exist")
         return None
 
-
     def clear(self) -> None:
         self.__init__()
 
--- a/app/main.py
+++ b/app/main.py
@@ -1,20 +1,19 @@
-import math
 from dataclasses import dataclass
+from fractions import Fraction
 from typing import Hashable, Any, Iterator
 
 
-@dataclass
-class Node:
-    key: Hashable
-    obj_hash: hash
-    value: Any
-
-
 class Dictionary:
-    def __init__(self, capacity: int = 8) -> None:
-        self.capacity = capacity
+    @dataclass
+    class Node:
+        key: Hashable
+        obj_hash: hash
+        value: Any
+
+    def __init__(self) -> None:
+        self.capacity = 8
         self.size = 0
-        self.hash_table: list[Node | None] = [None] * self.capacity
+        self.hash_table: list[Dictionary.Node | None] = [None] * self.capacity
 
     def _calculate_index(self, key: Hashable) -> int:
         index = hash(key) % self.capacity
@@ -28,28 +27,25 @@ class Dictionary:
         return index
 
     @property
-    def current_max_size(self) -> int:
-        return math.floor(self.capacity * 2 / 3)
+    def current_max_size(self) -> Fraction:
+        return self.capacity * Fraction(2 , 3)
 
     def _resize(self) -> None:
         old_hash_table = self.hash_table
-
-        self.__init__(self.capacity * 2)
-
+        self.capacity *= 2
+        self.size = 0
+        self.hash_table = [None] * self.capacity
         for node in old_hash_table:
             if node:
-                self.__setitem__(node.key, node.value)
+                self[node.key] = node.value
 
     def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size > self.current_max_size:
+            self._resize()
         index = self._calculate_index(key)
-
-        if self.hash_table[index] is None:
-            if self.size + 1 >= self.current_max_size:
-                self._resize()
-                return self.__setitem__(key, value)
+        if not self.hash_table[index]:
             self.size += 1
-
-        self.hash_table[index] = Node(key, hash(key), value)
+        self.hash_table[index] = Dictionary.Node(key, hash(key), value)
 
     def __getitem__(self, item: Any) -> Any:
         index = self._calculate_index(item)
@@ -88,28 +84,36 @@ class Dictionary:
         temp = None
         index = self._calculate_index(key)
         if self.hash_table[index]:
-            temp = self.hash_table[index]
+            temp = self.hash_table[index].value
             self.hash_table[index] = None
         return temp
 
     def keys(self) -> list[Hashable] | None:
         keys = []
         for node_key in self.hash_table:
-            keys.append(node_key.key)
+            if node_key:
+                keys.append(node_key.key)
         return keys if keys else None
 
-    def values(self) -> Any:
+    def values(self) -> list[Any] | None:
         values = []
         for node_value in self.hash_table:
-            values.append(node_value.value)
+            if node_value:
+                values.append(node_value.value)
         return values if values else None
 
-    def items(self) -> tuple[Hashable, Any] | tuple[None, None]:
-        return self.keys(), self.values()
+    def items(self) -> list[tuple[Hashable, Any]] | None:
+        items_ = []
+        for key, value in self.values(), self.items():
+            if key and value:
+                items_.append((key, value))
+        return items_ if items_ else None
 
     def update(self, other: object) -> None:
-        if isinstance(other, Dictionary):
+        if isinstance(other, Dictionary) or type(other) is dict:
             for key, value in other.items():
                 self.hash_table[key] = value
+        elif type(other) is Dictionary.Node:
+            self.hash_table[other.key] = other.value
         else:
             raise TypeError("Update data must be Dictionary!")
--- a/app/main.py
+++ b/app/main.py
@@ -30,9 +30,9 @@ class Dictionary:
     def current_max_size(self) -> Fraction:
         return self.capacity * Fraction(2 , 3)
 
-    def _resize(self) -> None:
+    def _resize(self, multiplier: str) -> None:
         old_hash_table = self.hash_table
-        self.capacity *= 2
+        self.capacity //= 2 if multiplier == "resize down" else 0.5
         self.size = 0
         self.hash_table = [None] * self.capacity
         for node in old_hash_table:
@@ -41,13 +41,13 @@ class Dictionary:
 
     def __setitem__(self, key: Hashable, value: Any) -> None:
         if self.size > self.current_max_size:
-            self._resize()
+            self._resize("resize up")
         index = self._calculate_index(key)
         if not self.hash_table[index]:
             self.size += 1
         self.hash_table[index] = Dictionary.Node(key, hash(key), value)
 
-    def __getitem__(self, item: Any) -> Any:
+    def __getitem__(self, item: Hashable) -> Any:
         index = self._calculate_index(item)
 
         if self.hash_table[index] is None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,2 +1,114 @@
...
+                return self.__setitem__(key, value)
+            self.size += 1
+
+        self.hash_table[index] = Node(key, hash(key), value)
+
+    def __getitem__(self, item: Any) -> Any:
+        index = self._calculate_index(item)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find object with key -> {item}")
```
oliasymak24, line 58: Which object you try to find?

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+        self.hash_table[index] = None
+        self.size -= 1
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __iter__(self) -> Iterator:
+        return iter(self.hash_table)
+
+    def get(self, key: Hashable) -> Hashable:
```
oliasymak24, line 77: https://www.programiz.com/python-programming/methods/dictionary/get
Read here about get method, you need to return some default value or None if you don't have a key

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+        try:
+            return self[key]
+        except KeyError:
+            print(f"Error! Value with key {key} does not exist")
+
+    def clear(self) -> None:
+        self.__init__(8)
+
+    def pop(self, key: Hashable) -> Any | None:
```
oliasymak24, line 86: https://www.programiz.com/python-programming/methods/dictionary/pop
The same here)

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+    def values(self) -> Any:
+        values = []
+        for node_value in self.hash_table:
+            values.append(node_value.value)
+        return values
+
+    def items(self) -> tuple:
+        return self.keys(), self.values()
+
+    def update(self, other: object) -> None:
```
oliasymak24, line 109: Add Better annotation
If update() is called without passing parameters, the dictionary remains unchanged.

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+        if self.hash_table[index]:
+            temp = self.hash_table[index]
+            self.hash_table[index] = None
+        return temp
+
+    def keys(self) -> list[Hashable]:
+        keys = []
+        for node_key in self.hash_table:
+            keys.append(node_key.key)
+        return keys
```
oliasymak24, line 98: What will happen if your dict is empty?
check values function too

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,2 +1,115 @@
+import math
+from dataclasses import dataclass
+from typing import Hashable, Any, Iterator
+
+
+@dataclass
+class Node:
```

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+    key: Hashable
+    obj_hash: hash
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8) -> None:
```

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+            self.hash_table[index]
+            and self.hash_table[index].key != key
+        ):
+            index = (index + 1) % self.capacity
+
+        return index
+
+    @property
+    def current_max_size(self) -> int:
+        return math.floor(self.capacity * 2 / 3)
```
Oleksl888: you can most accurately present 2/3 with Fraction type

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+
+    def _resize(self) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * 2)
+
+        for node in old_hash_table:
+            if node:
+                self.__setitem__(node.key, node.value)
```

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            if self.size + 1 >= self.current_max_size:
+                self._resize()
+                return self.__setitem__(key, value)
```

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+        return self.hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
```
Oleksl888, line 68: you need to recalculate remaining positions after deleting an element because indexes may be wrong for elements that had collisions

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __iter__(self) -> Iterator:
+        return iter(self.hash_table)
```
Oleksl888: this will not be the most efficient iterator since it will not skip None values

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+
+    def get(self, key: Hashable) -> Hashable | None:
```
Oleksl888: get method should have a default value parameter and it doesn't raise exception

app/main.py
Oleksl888: pop method should be work the same as it works in oython dict

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+        index = self._calculate_index(key)
+        if self.hash_table[index]:
+            temp = self.hash_table[index]
+            self.hash_table[index] = None
+        return temp
+
+    def keys(self) -> list[Hashable] | None:
+        keys = []
+        for node_key in self.hash_table:
+            keys.append(node_key.key)
```
Oleksl888: this will fail when you will try to take .key attribute from None

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+        return keys if keys else None
+
+    def values(self) -> Any:
+        values = []
+        for node_value in self.hash_table:
+            values.append(node_value.value)
```
Oleksl888: same - need to filter results better

app/main.py
```diff
@@ -1,2 +1,115 @@
...
+        return values if values else None
+
+    def items(self) -> tuple[Hashable, Any] | tuple[None, None]:
+        return self.keys(), self.values()
```
Oleksl888: this will return one tuple with two huge lists, but instead it should be a list of matching key-value tuples

app/main.py
Oleksl888: other can also be an Iterabale with key-value pairs

Oleksl888 (changes requested):

app/main.py
Oleksl888: This will not pass the optional test on delete item - after deletion you need to recount indexes and reset all remaining elements in hash table to be sure data is correct

app/main.py
Oleksl888: try to use generator expression or a generator method here

app/main.py
Oleksl888: not changed

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+
+    def get(self, key: Hashable) -> Hashable | None:
+        try:
+            return self[key]
+        except KeyError:
+            print(f"Error! Value with key {key} does not exist")
+        return None
+
+    def clear(self) -> None:
+        self.__init__()
```
Oleksl888: init should be called once on object creation. It is a bad practice to reuse it multiple times

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+
+    def pop(self, key: Hashable, default: Optional[Any]) -> Any | None:
+        temp = None
+        index = self._calculate_index(key)
+        if self.hash_table[index]:
```
Oleksl888: you can just `try` to get an element using brackets interface and then do something else in except clause.
Also keep in mind that pop method should have this behaviour:
Dictionary.pop(key, None) should return None instead of triggering KeyError

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+            temp = self.hash_table[index].value
+            self.hash_table[index] = None
+        if temp:
+            return temp
+        return default if default else temp
+
+    def keys(self) -> list[Hashable] | None:
+        keys = []
+        for node_key in self.hash_table:
```
Oleksl888: you can just make a list comprehension here

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+            if node_key:
+                keys.append(node_key.key)
+        return keys if keys else None
+
+    def values(self) -> list[Any] | None:
+        values = []
+        for node_value in self.hash_table:
```
Oleksl888: same - list comprehension would optimize it

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+            if node_value:
+                values.append(node_value.value)
+        return values if values else None
+
+    def items(self) -> list[tuple[Hashable, Any]] | None:
+        items_ = []
+        for key, value in self.keys(), self.values():
```
Oleksl888: i guess `in zip(self.keys, self.values)` would be more correct

app/main.py
```diff
@@ -1,2 +1,124 @@
...
+            if key:
+                items_.append((key, value))
+        return items_ if items_ else None
+
+    def update(self, other: Optional[object]) -> None:
```
Oleksl888: other should be a Dictionary or an Iterable of key-value pairs

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,132 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        if self.size > self.current_max_size:
+            self._resize()
+        index = self._calculate_index(key)
+        if not self.hash_table[index]:
+            self.size += 1
+        self.hash_table[index] = Dictionary.Node(key, hash(key), value)
+
+    def __getitem__(self, item: Any) -> Any:
```
Oleksl888: here we expect a Hashable type

app/main.py
```diff
@@ -1,2 +1,132 @@
...
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if not self.hash_table[index]:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
+        if self.size <= self.current_max_size // 2:
```
Oleksl888, line 69: so you don't need to actually change capacity of table when element is deleted. You can make a small change to resize method os it can accept multiplier parameter. And so when it's actual resize you pass there a multiplier of 2 and when it is resize due to element deletion then run it with multiplier of 1

app/main.py
```diff
@@ -1,2 +1,132 @@
...
+            self.capacity //= 4
+            self._resize()
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __iter__(self) -> Iterator:
+        generate = iter(self.hash_table)
```
Oleksl888: `return (node.key for node in self.hash_table if node is not None)` thats it

app/main.py
```diff
@@ -1,2 +1,132 @@
...
+        yield next(generate)
+
+    def get(self, key: Hashable) -> Hashable | None:
```
Oleksl888: are you planning to add default parameter here? if you give up simply remove this method :)

app/main.py
```diff
@@ -1,2 +1,132 @@
...
+        return items_ if items_ else None
+
+    def update(self, other: Optional[Iterable]) -> None:
+        if not other:
+            return
+
+        if isinstance(other, Dictionary) or type(other) is dict:
+            for key, value in other.items():
+                self.hash_table[key] = value
+        elif all(
```
Oleksl888: i think here you can just do without condition `for key, value in other` (considering it is an iterable of tulpes). And if theres wrong types or something let python throw error by itself

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,129 @@
...
+        ):
+            index = (index + 1) % self.capacity
+
+        return index
+
+    @property
+    def current_max_size(self) -> Fraction:
+        return self.capacity * Fraction(2 , 3)
+
+    def _resize(self, multiplier: str) -> None:
```
Oleksl888: i'd rather made a flag resize_up: bool = True so that if you call it without parameters in setitem method it would be True and in method with delitem you would pass False explicitly - that's better design then passing strings

app/main.py
```diff
@@ -1,2 +1,129 @@
...
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if not self.hash_table[index]:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
+        if self.size <= self.current_max_size // 2:
+            self.capacity //= 4
```
Oleksl888: why would change the capacity here and then again change capacity in resize method?

app/main.py
```diff
@@ -1,2 +1,129 @@
...
+            self._resize("resize down")
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __iter__(self) -> Iterator:
+        return (node.key for node in self.hash_table if node is not None)
+
+    def get(self, key: Hashable = None) -> Hashable | None:
```
Oleksl888: there is still no default parameter here method signature shold be `def get(self, key: Hashable, default: Optional[Any] = None)`

app/main.py
```diff
@@ -1,2 +1,129 @@
...
+        except KeyError:
+            print(f"Error! Value with key {key} does not exist")
+        return None
+
+    def clear(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.hash_table: list[Dictionary.Node | None] = [None] * 8
+
+    def pop(self, key: Hashable, default: Optional[Any]) -> Any | None:
```
Oleksl888: this does not behave as expected try to set default as object() and check against that to raise an error

app/main.py
```diff
@@ -1,2 +1,129 @@
...
+
+    def update(self, other: Optional[Iterable]) -> None:
+        if not other:
+            return
+
+        if isinstance(other, Dictionary) or type(other) is dict:
+            for key, value in other.items():
+                self.hash_table[key] = value
+        else:
+            for i in range(len(other)):
```
Oleksl888: here you could use unpacking like so:
`for key, value in other...`

just for the future `in range(len)` is a nooby habit (sorry 😉 ), use enumerate instead - it is more efficient and your code will earn more respect :)

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,138 @@
...
+
+    def update(self, other: Optional[Iterable]) -> None:
+        if not other:
+            return
+
+        if isinstance(other, Dictionary) or type(other) is dict:
+            for key, value in other.items():
+                self.hash_table[key] = value
+        else:
+            for index, value in enumerate(other):
```
Oleksl888, line 137: you could just do for key, value in other ...
the comment about enumerate is valid but it doesn't mean you have to use it here - it's unnecessary right now

app/main.py
```diff
@@ -1,2 +1,138 @@
...
+
+    def clear(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.hash_table: list[Dictionary.Node | None] = [None] * 8
+
+    def pop(
+            self,
+            key: Hashable,
+            default: Optional[object] = None
```
Oleksl888, line 99: must be default: Optional[object] = object()

Oleksl888 (approved):
You deserve some rest after this :) 🍰

PR #759, opened 2024-08-19
first

egor250330:


app/main.py
```diff
@@ -1,2 +1,72 @@
...
+
+@dataclass
+class Node:
+    key: Hashable
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8) -> None:
```
Oleksl888, line 17: capacity shouldn't be passed as parameter - it cannot be set from outside of Dict

app/main.py
```diff
@@ -1,2 +1,72 @@
...
+            index = (index + 1) % self.capacity
+        return index
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self.calculate_index(key)
+        if self.buckets[index] is None:
+            self.size += 1
+        self.buckets[index] = Node(key, value)
+
+        if self.size >= self.max_size:
```
Oleksl888, line 37: resizing actually happens before you insert a new element

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+    CAPACITY = 8
+    INCREASE_MULTIPLIER = 2
+    LOAD_FACTOR = Fraction(2, 3)
+
+    @dataclass
+    class Node:
+        key: Hashable
+        value: Any
+
+        def __post_init__(self) -> None:
```
Oleksl888, line 16: well done with this

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+        return self.size
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self.calculate_index(key)
+
+        if self.buckets[index] is None:
+            raise KeyError(f"Can't find element with key {key}")
+
+        self.buckets[index] = None
+        self.size -= 1
```
Oleksl888, line 75: In case if there were collisions some indexes lead to incorrect Nodes after a Node is deleted. It can be fixed by recalculating indexes for remaining elements of hash table

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+        key: Hashable
+        value: Any
+
+        def __post_init__(self) -> None:
+            self.key_hash = hash(self.key)
+
+    def __init__(self) -> None:
+        self.capacity = self.CAPACITY
+        self.size = 0
+        self.buckets = [None] * self.capacity
```
Oleksl888: Would make sense to annotate hash table as a list of Optional Nodes

app/main.py
```diff
@@ -1,2 +1,86 @@
...
+
+        if self.buckets[index] is None:
+            raise KeyError(f"Can't find element with key {key}")
+
+        self.buckets[index] = None
+        self.size -= 1
+
+        self.rehash()
+
+    def rehash(self) -> None:
```
Oleksl888, line 80: This method is basically the same as your resize method. You can adjust existing resize method so it runs with or without multiplier and reuse it in setiitem and delitem  to make your code DRY

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+        if self.buckets[index] is None:
+            raise KeyError(f"Can't find element with key {key}")
+
+        self.buckets[index] = None
+        self.size -= 1
+
+        if self.size <= self.max_size // self.INCREASE_MULTIPLIER:
+            self.resize(use_multiplier=False)
+
+    def rehash(self) -> None:
```
Oleksl888, line 80: You don't use it anymore - so remove

Oleksl888 (approved):
Well done! Please clean up as per last comment

PR #760, opened 2024-08-19
Solution

olena4reunited:


app/main.py
```diff
@@ -1,2 +1,101 @@
+from dataclasses import dataclass
+from typing import Hashable, Any, Iterator
+
+INITIAL_CAPACITY = 8
```
Oleksl888, line 4: these should be class variables inside of Dictionary since it is part of it implementation

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+RESIZE_THRESHOLD = 2 / 3
```
Oleksl888, line 5: Fraction type would be perfect here

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+CAPACITY_MULTIPLIER = 2
+
+
+@dataclass
+class Node:
```
Oleksl888, line 10: Hash should be also included here + would make sense to declare this class inside of Dictionary since it is part of it implementation

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+    key: Hashable
+    value: Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = INITIAL_CAPACITY) -> None:
```
Oleksl888, line 16: capacity shouldnt be parametrized - it's part of dict internal structure

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        self.capacity = capacity
+        self.size = 0
+        self.hash_table: list[Node | None] = [None] * self.capacity
+
+    def _calculate_index(self, key: Hashable) -> int:
+        index = hash(key) % self.capacity
+
+        while (
+            self.hash_table[index] is not None
```
Oleksl888, line 25: You must check that key and hash value of a Node equals to one that is being looked, right now you only check if key is the same

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        return self.capacity * RESIZE_THRESHOLD
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+
+        self.__init__(self.capacity * CAPACITY_MULTIPLIER)
+
+        for node in old_hash_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
```
Oleksl888, line 43: you can use syntactic sugar since magic method has been defined

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            if self.size + 1 >= self.current_max_size:
+                self.resize()
+                return self.__setitem__(key, value)
```
Oleksl888, line 51: can you rewrite so you avoid the recursive call?

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot delete value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
+
+    def __iter__(self) -> Iterator[tuple[Hashable, Any]]:
+        for node in self.hash_table:
```
Oleksl888, line 74: nice!

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def __str__(self) -> str:
+        items = [
+            f"{node.key}: {node.value}"
+            for node in self.hash_table
+            if node is not None
+        ]
+        return "{" + ", ".join(items) + "}"
+
+    def get(self, key: Hashable, default: Any = None) -> Node:
```
Oleksl888, line 89: annotate that default is optional

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Node:
```
Oleksl888, line 95: annotate that default is optional + if key is not in dictionary and default is not set - error must be triggered,

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        value = self[key]
+        self.__delitem__(key)
+        return value
+
+    def update(self, data: Node) -> None:
```
Oleksl888, line 100: update method takes another dict or an interable with key value pairs

Oleksl888 (changes requested):

app/main.py
Oleksl888: not changed

app/main.py
Oleksl888: not fixed

app/main.py
```diff
@@ -1,2 +1,204 @@
...
+        return self.hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot delete value for key: {key}")
+
+        self.hash_table[index] = None
+        self.size -= 1
```
Oleksl888, line 74: you need to recalculate remaining positions after deleting an element because indexes may be wrong for elements that had collisions

app/main.py
```diff
@@ -1,2 +1,204 @@
...
+        """
+        Get the value for the given key.
+        If the key does not exist, return the default value.
+
+        Default value is optional.
+        If it was not provided, method will raise KeyError.
+        """
+        try:
+            return self[key]
+        except KeyError:
```
Oleksl888, line 102:
```
        try:
            return self[key]
        except KeyError:
            return default

File: app/main.py [hunk 14]
```diff
@@ -1,2 +1,204 @@
...
+            if default is not None:
+                return default
+            else:
+                raise KeyError(
+                    f"Cannot find value for key: {key}"
+                    "and no default value is provided"
+                )
+
+    def pop(self, key: Hashable) -> Any:
```
Oleksl888, line 111: pop should also have a default param but it behaves a bit different from `get` when default value is not set and key is not found:
Dictionary.pop(key, None) should return None instead of triggering KeyError

app/main.py
```diff
@@ -1,2 +1,204 @@
...
+        self.__setitem__(key, value)
+
+    def update(
+            self,
+            data: Dictionary | list[Node | dict | tuple] | Node | dict | tuple
+    ) -> None:
+        """
+        Updates Dictionary with various types of data.
+        """
+        if isinstance(data, Dictionary):
```
Oleksl888, line 158: this method is overcomplicated

app/main.py
```diff
@@ -1,2 +1,204 @@
...
+
+    def copy(self) -> Dictionary:
+        """
+        Creates a copy of the dictionary.
+        """
+        new_dict = Dictionary(self.capacity)
+
+        for node in self.hash_table:
+            if node is not None:
+                new_dict.__setitem__(node.key, node.value)
```
Oleksl888, line 184: use sugar interface wherever possible since magic methods are defined

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,209 @@
...
+        return "{" + ", ".join(items) + "}"
+
+    def __eq__(self, other: Dictionary) -> bool:
+        """
+        Compare key, value pairs between two Dictionaries.
+        """
+        if not isinstance(other, Dictionary):
+            return False
+
+        for key, value in self:
```
Oleksl888, line 106: is there implementation for unpacking in your dictionary? how does this work?

app/main.py
```diff
@@ -1,2 +1,209 @@
...
+
+        Default value is optional.
+        If it was not provided, method will raise KeyError.
+        """
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Hashable, default: Optional[Any] = None) -> Any:
```
Oleksl888, line 124: for correct implementation you need to pass object() as default value and check for that. None is a valid parameter for this method, it shouldnt trigger a KeyError

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,2 +1,224 @@
...
+            if node is not None:
+                new_dict.__setitem__(node.key, node.value)
+
+        return new_dict
+
+    def clear(self) -> None:
+        """
+        Resets the dictionary to its initial state.
+        """
+        self.__init__()
```
Oleksl888: __init__ should be called only once on object creation, it is not recommended to call it multiple times within the class

app/main.py
```diff
@@ -1,2 +1,224 @@
...
+        if isinstance(data, Dictionary):
+            self._update_from_dict(data)
+        elif isinstance(data, dict):
+            self._update_from_builtin_dict(data)
+        elif isinstance(data, list):
+            for item in data:
+                if isinstance(item, dict):
+                    self._update_from_builtin_dict(item)
+                else:
+                    raise TypeError(f"Unsupported type {type(item)}")
```
Oleksl888: No need to raise those errors - if wrong types are being passed, python will trigger exceptions by itself. and we don't need to catch them here - they can be caught when we work with the object already

app/main.py
```diff
@@ -1,2 +1,224 @@
...
+        index = self._calculate_index(key)
+
+        if self.hash_table[index] is None:
+            raise KeyError(f"Cannot find value for key: {key}")
+
+        return self.hash_table[index].value
+
+    def __delitem__(self, key: Hashable) -> None:
+        if self.size - 1 < self.current_max_size // self.CAPACITY_MULTIPLIER:
+            self._resize_down()
```
Oleksl888: Good job on implementing size down. But actually it would make sense to call it after an element has been removed

Oleksl888 (approved):
Approved but check last comments please

PR #764, opened 2024-08-29
Realized Dictionary

NinqueRaven13:


app/main.py
```diff
@@ -1,2 +1,49 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 8
+        self.load_factor = 2 / 3
+        self.dictionary = [-1 for _ in range(8)]
+        self.insert_elem = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
vasylhnatiuk:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

vasylhnatiuk (changes requested):
add resize method

app/main.py
```diff
@@ -1,2 +1,52 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length = 8
+        self.load_factor = 2 / 3
+        self.dictionary = [-1 for _ in range(8)]
```
vsmutok:
```suggestion
        self.dictionary = [None for _ in range(8)]
```
better to use `None` in this case

vsmutok (changes requested):
now your code is not working correctly, because I can write 2 identical key value pairs in the dictionary 

image

EdAlekseiev (approved):

PR #772, opened 2024-09-05
Full solution

arsen-arutiunov:


app/main.py
```diff
@@ -1,2 +1,117 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.dict_as_list = [(), (), (), (), (), (), (), ()]
+        self.cells = len(self.dict_as_list)
+        self.max_elements = 5
+
+    def __setitem__(self,
+                    key: Any,
```
vsmutok, line 11:
```suggestion
                    key: Hashable,
```
please fix everywhere

app/main.py
```diff
@@ -1,2 +1,117 @@
...
+            if self.dict_as_list[hash(key) % self.cells][0] == key:
+                return self.dict_as_list[hash(key) % self.cells][2]
+
+        if not (self.search_element(key) is None):
+            return self.dict_as_list[self.search_element(key)][2]
+        else:
+            raise KeyError
+
+    def __len__(self) -> int:
+        return len(self.dict_as_list) - self.dict_as_list.count(())
```
vsmutok, line 77: This is a bad approach. You should have a variable `self.length`and increment it by one each time a new value is added to your dictionary. Then, just return `self.length` here.

app/main.py
```diff
@@ -1,2 +1,117 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.dict_as_list = [(), (), (), (), (), (), (), ()]
+        self.cells = len(self.dict_as_list)
+        self.max_elements = 5
```
vsmutok, line 8: You should have not a fixed number, but 2/3 that you can use to fill the dictionary

vsmutok (changes requested):
Also you code style not good
please use `black` to fix it.

PR #783, opened 2024-10-08
dictionary

dmytrik:


app/main.py
```diff
@@ -1,2 +1,90 @@
...
+            value
+        )
+        self.size += 1
+        if self.size == round(self.capacity * self.load_factor):
+            self.update()
+
+    def __getitem__(self, key: object) -> object:
+        cell = self.get_cell(key)
+        if not self.hash_table[cell]:
+            raise KeyError
```
AnyoneClown: We can add error message for this raise

app/main.py
```diff
@@ -1,2 +1,90 @@
...
+    def __init__(self) -> None:
+        self.size = 0
+        self.load_factor = 2 / 3
+        self.capacity = 8
+        self.hash_table: list = [None] * self.capacity
+
+    def get_cell(self, key: object) -> int:
+        return hash(key) % self.capacity
+
+    def update(self) -> None:
```
AnyoneClown: Better to name it like `resize` or smth. Because we are chaning size of our dictionary

AnyoneClown (approved):
Good job! Only small fixes

PR #785, opened 2024-10-09
custom dict

Lev2098:


app/main.py
```diff
@@ -1,2 +1,82 @@
+from typing import Any, List, Tuple
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.my_dict = []
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
vasylhnatiuk:
```suggestion
    def __setitem__(self, key: Hashable, value: Any) -> None:
```

vasylhnatiuk: fix everywhere

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,2 +1,82 @@
...
+        return default
+
+    def pop(self, key: Hashable, default: Any = None) -> Any:
+        key_hash = self._hash(key)
+        for i, (h, (k, v)) in enumerate(self.my_dict):
+            if h == key_hash and k == key:
+                del self.my_dict[i]
+                return v
+        if default is None:
+            return default
```
vsmutok: what will happen if we pass another value for default?
this should work as well

vsmutok (changes requested):
You did not fulfill one of the key requirements, which is the correct maintenance of the hash table. The initial table must be created with a size of 8. Once it is filled by 2/3, resizing should occur.

vasylhnatiuk (approved):

PR #787, opened 2024-10-14
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -1,2 +1,107 @@
...
 class Dictionary:
-    pass
+    DELETED = "DELETED"
+
+    def __init__(self) -> None:
+        """Initialize the Dictionary with default capacity and load factor."""
+        self.capacity = 8
+        self.length = 0
+        self.load_factor = 2 / 3
+        self.hash_table = [None] * self.capacity
```
AnyoneClown, line 19: Read task properly. Your hash table should be a list of `Node`. Create `Node` class with 3 attributes(Key, hash, value). Refactor your code.

> You'll store each key-value pair as a node in a hash table. Each node should contain the following information:
> - `key`
> - `hash`
> - `value`

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+        old_table = self.hash_table
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+        self.length = 0
+
+        for node in old_table:
+            if node and node is not self.DELETED:
+                self.__setitem__(node.key, node.value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
Arsen-hrynevych, line 36: Consider adding a check to avoid infinite loops

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+            index = (index + 1) % self.capacity
+
+        node = Node(key, value, self._hash(key))
+        self.hash_table[index] = node
+        self.length += 1
+
+        if self.length / self.capacity > self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: Any) -> Any:
```
Arsen-hrynevych, line 56: Consider adding a check to avoid infinite loops

app/main.py
```diff
@@ -1,2 +1,121 @@
...
+        self.capacity = 8
+        self.length = 0
+        self.load_factor = 2 / 3
+        self.hash_table = [None] * self.capacity
+
+    def _hash(self, key: Any) -> int:
+        """Return the hash index for a given key."""
+        return hash(key) % self.capacity
+
+    def _resize(self) -> None:
```
Arsen-hrynevych, line 25: The `pop` method checks for a default value but doesn’t address the case where the default is `None`. It might be clearer to return the default only when the key doesn’t exist.

Arsen-hrynevych (approved):
Ok

PR #788, opened 2024-10-17
Solution

Atikiho:


app/main.py
```diff
@@ -1,2 +1,69 @@
...
+                self.nodes[hash(key) % self.capacity] = [key, value, hash(key)]
+        else:
+            self.nodes[hash(key) % self.capacity] = [key, value, hash(key)]
+
+    def __getitem__(self, key: any) -> any:
+        node = self.nodes[hash(key) % self.capacity]
+        if node:
+            if node[0] != key:
+                for i in range(0, self.capacity):
+                    if self.nodes[i] and self.nodes[i][0] == key:
```
Arsen-hrynevych: In the `__setitem__` method, during hash table resizing, you're transferring nodes with this line:

```python
if not self.nodes[node[2] % self.capacity]:
    self.nodes[node[2] % self.capacity] = node
```

The issue is that `node[2]` assumes the hash is stored in the node, but not all entries may follow this format. Also, you're directly assigning nodes instead of rehashing and reinserting them based on their new hash values. The nodes need to be properly rehashed and inserted using the correct mechanism.

app/main.py
```diff
@@ -1,2 +1,69 @@
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.nodes = [[]] * 8
```
Arsen-hrynevych: When initializing `self.nodes`, you're using:

```python
self.nodes = [[]] * 8
```

This creates 8 references to the same list, so changing one affects all others. Instead, you should use:

```python
self.nodes = [[] for _ in range(8)]
```

This ensures each bucket is a separate, distinct list.

app/main.py
```diff
@@ -1,2 +1,69 @@
...
+                    if not self.nodes[i] or self.nodes[i][0] == key:
+                        self.nodes[i] = [key, value, hash(key)]
+                        break
+            else:
+                self.nodes[hash(key) % self.capacity] = [key, value, hash(key)]
+        else:
+            self.nodes[hash(key) % self.capacity] = [key, value, hash(key)]
+
+    def __getitem__(self, key: any) -> any:
+        node = self.nodes[hash(key) % self.capacity]
```
Arsen-hrynevych: When searching for a key in the hash table, you're using:

```python
node = self.nodes[hash(key) % self.capacity]
```

This assumes the key is always at the hashed index, but it won't work if there are collisions (multiple keys hashed to the same index). You need to handle collisions by iterating through the list of items at that index to find the correct key.

app/main.py
```diff
@@ -1,2 +1,69 @@
...
+        if node:
+            if node[0] != key:
+                for i in range(0, self.capacity):
+                    if self.nodes[i] and self.nodes[i][0] == key:
+                        return self.nodes[i][1]
+            return node[1]
+        raise KeyError("There is no such key in the dictionary")
+
+    def __len__(self) -> int:
+        return len(self.nodes) - self.nodes.count([])
```
Arsen-hrynevych: In the `__len__` method, you're counting non-empty lists with:

```python
return len(self.nodes) - self.nodes.count([])
```

However, this doesn't give the correct number of key-value pairs. Instead, you should track the actual number of key-value pairs inserted into the dictionary and return that count.

app/main.py
```diff
@@ -1,2 +1,69 @@
...
+
+    def __delitem__(self, key: any) -> None:
```
Arsen-hrynevych, line 37: The deletion logic is incomplete. It attempts to delete a key based on its hash match, but doesn't properly handle collisions. If multiple keys exist in the same bucket, the method should iterate through the bucket (list), find the correct key, and then delete it.

Arsen-hrynevych (changes requested):
Mostly correct. Good job

Arsen-hrynevych (approved):
ok

PR #792, opened 2024-10-22
'Solution'

AndriiDementiev:


vladyslav-tmf (approved):
Looks good.

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+
+
 class Dictionary:
-    pass
+
+    class Node:
+        def __init__(self, key: Any, value: Any) -> None:
+            self.key = key
+            self.value = value
+            self.hash = hash(key)
```
Nattalli: btw you can separate that class and write outside of `Dictionary` class

Nattalli (changes requested):

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+        self.size = 0
+        old_table = self.hash_table
+        self.hash_table = [None] * self.capacity
+
+        for node in old_table:
+            if node is not None:
+                self.__setitem__(node.key, node.value)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        index = hash(key) % self.capacity
```
Arsen-hrynevych, line 34: Currently, resizing happens after inserting a new element when the load factor exceeds 2/3. But this means for every new insertion past the limit, an unnecessary resize might occur, especially for high-load situations.

Perform the resize before the insertion to avoid inserting and then resizing immediately.

```
 if self.size / object > object:  # Check before insertion
        self.resize()
```

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+
+        while True:
```
Arsen-hrynevych, line 36: In your current solution, if the table reaches full capacity without resizing, the while True loop could be in an infinite loop.

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+            node = self.hash_table[index]
+
```
Arsen-hrynevych, line 38: Implement a termination condition or an assertion to ensure the loop exits correctly.

app/main.py
```diff
@@ -1,2 +1,63 @@
...
+            index = (index + 1) % self.capacity
+
+    def __getitem__(self, key: Any) -> Any:
+        index = hash(key) % self.capacity
+
+        while True:
+            node = self.hash_table[index]
+
+            if node is None:
+                raise KeyError(f"Key {key} not found")
```
Arsen-hrynevych: include a bit more information in the KeyError

app/main.py
```diff
@@ -1,2 +1,63 @@
+from typing import Any
+
+
+class Node:
```
Arsen-hrynevych, line 4: Did you find that solution on web or this is suggestion of generative AI?

Arsen-hrynevych (changes requested):
Try to add more dunder methods and fix the comments below.

PR #799, opened 2024-10-23
solution

VolodymyrSemchysyn:


Bondzik-S (approved):

app/main.py
```diff
@@ -1,2 +1,38 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._hash(key)
+
+        if self.hash_table[index] is None:
+            self.hash_table[index] = []
+
+        for i, (k, v) in enumerate(self.hash_table[index]):
+            if k == key:
+                self.hash_table[index][i] = (key, value)
+                return
```
AnyoneClown: > You'll store each key-value pair as a node in a hash table. Each node should contain the following information:
> 
> key
> hash
> value

Read your task again and fix it

app/main.py
```diff
@@ -1,2 +1,38 @@
...
+        self.hash_table: list = [None] * 8
+
+    def _hash(self, key: Any) -> int:
+        return hash(key) % len(self.hash_table)
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        index = self._hash(key)
+
+        if self.hash_table[index] is None:
+            self.hash_table[index] = []
```
AnyoneClown: Your dicitionary will only have max 8 elements inside? What if we want to store 18 elements?

Create `resize` logic to increase size of your dict

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+
+        for node in self.hash_table[index]:
+            if node.key == key:
+                node.value = value
+                return
+
+        self.hash_table[index].append(Node(key, value))
+        self.length += 1
+
+    def __getitem__(self, key: Any) -> Any:
```
Arsen-hrynevych, line 44: key: Any -> key: Hashable

app/main.py
```diff
@@ -1,2 +1,52 @@
...
+        self.hash_table: List[List[Node]] = [[] for _ in range(self.capacity)]
+
+    def _hash(self, key: Any) -> int:
+        return hash(key) % self.capacity
+
+    def _resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.capacity *= 2
+        self.hash_table = [[] for _ in range(self.capacity)]
+        self.length = 0
```
Arsen-hrynevych, line 24: consider reducing length by 1 when an item is replaced. For example, if self[key] = value is called for an existing key, length should stay the same.

Arsen-hrynevych (approved):
ok

PR #800, opened 2024-10-23
Solution-py-dictionary

skyfoxwork:


FursykIvan (approved):
Sorry for the lack of good comments, too complicated for me to understand. If it works, let the mentor check it.

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+    def __init__(self) -> None:
+        self.length: int = 0
+        self.capacity: int = 8
+        self.hash_table: list[Optional[tuple]] = [None] * self.capacity
+
+    def _get_key_value(self, key: Any) -> int | None:
+        index = hash(key) % self.capacity
+        original_index = index
+
+        while self.hash_table[index] is not None:
```
Nattalli: you can remove `is not None`part

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+
+        return None
+
+    def _set_key_value(self, key: Any, value: Any) -> int | None:
+        self._check_extend_memory()
+
+        index = hash(key) % self.capacity
+        original_index = index
+
+        while self.hash_table[index] is not None:
```
Nattalli: you can remove `is not None` part

app/main.py
```diff
@@ -1,2 +1,66 @@
...
+
+            if index == original_index:
+                break
+
+        self.hash_table[index] = (key, hash(key), value)
+        self.length += 1
+
+    def __getitem__(self, key: Any) -> Any:
+        value = self._get_key_value(key)
+        if value is None:
```
Nattalli, line 44:
```suggestion
        if not value:
```

Nattalli (changes requested):

app/main.py
skyfoxwork: when I use:

    if not value:

instead of:

    if value is None:

tests crash, when value == 0

I think it is because bool(0) == False,  bool(None) == False and bool(other type) == True.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.length: int = 0
+        self.capacity: int = 8
+        self.hash_table: list[Optional[tuple]] = [None] * self.capacity
+
+    def _get_key_value(self, key: Hashable) -> Any:
```
Arsen-hrynevych, line 10: You're recalculating the hash value for the key multiple times, both in _get_key_value and _set_key_value as well as inside the tuple (key, hash(key), value). This can be optimized by calculating the hash once and passing it around where necessary.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+            if self.hash_table[index][0] == key:
+                return self.hash_table[index][2]
+            index = (index + 1) % self.capacity
+
+            if index == original_index:
+                break
+
+        return None
+
+    def _set_key_value(self, key: Hashable, value: Any) -> None:
```
Arsen-hrynevych, line 24: You're recalculating the hash value for the key multiple times, both in _get_key_value and _set_key_value as well as inside the tuple (key, hash(key), value). This can be optimized by calculating the hash once and passing it around where necessary.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        index = hash(key) % self.capacity
+        original_index = index
+
+        while self.hash_table[index]:
+            if self.hash_table[index][0] == key:
+                self.hash_table[index] = (key, hash(key), value)
+                return
+            index = (index + 1) % self.capacity
+
+            if index == original_index:
```
Arsen-hrynevych, line 36: This check is redundant. Resizing mechanism should prevent this situation from occurring.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+                break
+
+        self.hash_table[index] = (key, hash(key), value)
+        self.length += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
+        value = self._get_key_value(key)
+        if value is None:
+            raise KeyError
```
Arsen-hrynevych, line 45: Please add error message

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+            if entry is not None:
+                # it == self.__setitem__(entry[0], entry[2])
+                self[entry[0]] = entry[2]
+
+    def _check_extend_memory(self) -> None:
+        if self.length >= (2 / 3) * self.capacity:
+            old_table = self.hash_table
+            self.capacity *= 2
+            self.hash_table = [None] * self.capacity
+            self.length = 0
```
Arsen-hrynevych, line 65: Remove the reset of self.length, as it's unnecessary during rehashing. Instead, just keep self.length intact while resizing.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        return self.length
+
+    def _re_sort_memory(self, old_table: list) -> None:
+        for entry in old_table:
+            if entry is not None:
+                # it == self.__setitem__(entry[0], entry[2])
+                self[entry[0]] = entry[2]
+
+    def _check_extend_memory(self) -> None:
+        if self.length >= (2 / 3) * self.capacity:
```
Arsen-hrynevych, line 61: Include a check to ensure that the capacity never shrinks below a certain threshold.

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+            raise KeyError
+        return value
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        self._set_key_value(key, value)
+
+    def __len__(self) -> int:
+        return self.length
+
+    def _re_sort_memory(self, old_table: list) -> None:
```
Arsen-hrynevych, line 54: Consider adding docstrings to your method

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        for entry in old_table:
+            if entry is not None:
+                # it == self.__setitem__(entry[0], entry[2])
+                self[entry[0]] = entry[2]
+
+    def _check_extend_memory(self) -> None:
```
Arsen-hrynevych, line 60: Consider adding docstrings to your method

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        self.hash_table[index] = (key, hash(key), value)
+        self.length += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
+        value = self._get_key_value(key)
+        if value is None:
+            raise KeyError
+        return value
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
```
Arsen-hrynevych, line 48: Consider adding docstrings to your method

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+                return
+            index = (index + 1) % self.capacity
+
+            if index == original_index:
+                break
+
+        self.hash_table[index] = (key, hash(key), value)
+        self.length += 1
+
+    def __getitem__(self, key: Hashable) -> Any:
```
Arsen-hrynevych, line 42: Consider adding docstrings to your method

app/main.py
Arsen-hrynevych, line 24: Consider adding docstrings to your method

app/main.py
Arsen-hrynevych, line 10: Consider adding docstrings to your method

Arsen-hrynevych (changes requested):
keep going.

PR #802, opened 2024-10-24
Solution

FursykIvan:


app/main.py
```diff
@@ -1,2 +1,64 @@
...
+        new_capacity = self.capacity * 2
+        new_table = [None] * new_capacity
+
+        for node in self.table:
+            while node is not None:
+                current_index = hash(node.key) % new_capacity
+                new_node = Node(node.key, node.value, node.hash)
+                new_node.next = new_table[current_index]
+                new_table[current_index] = new_node
+                node = node.next
```
pohgen: time complexity in the worse situation is O(N^2) - it's unacceptable

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+            self._resize()
+
+    def __getitem__(self, key: Any) -> None:
+        current_index = self._hash_index(key)
+        current = self.table[current_index]
+        while current is not None:
+            if current.key == key:
+                return current.value
+            current = current.next
+        raise KeyError(f"Key '{key}' is not found")
```
pohgen: you can replace "while current is not none" -> "while current"

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+
+    def __len__(self) -> int:
+        return self.size
+
```
pohgen: len return wrong value, when we change key

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+        self.table = [None] * self.capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        current_index = self._hash_index(key)
+        current = self.table[current_index]
+        while current is not None:
+            if current.key == key:
+                current.value = value
+                return
+            current = current.next
```
pohgen: you can replace "while current is not none" -> "while current"

YaYaYaroslav (approved):
correct

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+        self.hash = hash_value
+        self.next = None
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.size = 0
+        self.load_factor = 0.7
```
Nattalli: I believe that `2/3` would be better

app/main.py
Nattalli: agree
```suggestion
        while current:
            if current.key == key:
                current.value = value
                return
            current = current.next
```

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+        current_index = self._hash_index(key)
+        current = self.table[current_index]
+        while current is not None:
+            if current.key == key:
+                current.value = value
+                return
+            current = current.next
+
+        new_node = Node(key, value, hash(key))
+        new_node.next = self.table[current_index]
```
Nattalli: shouldn't `next` attribute save the next item in the table? do you really need that attribute at all?

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+        self.table[current_index] = new_node
+        self.size += 1
+
+        if self.size / self.capacity > self.load_factor:
+            self._resize()
+
+    def __getitem__(self, key: Any) -> None:
```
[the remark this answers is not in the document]
Nattalli: agree

app/main.py
Nattalli: agree, if you add a value with already existed key - size shouldn't be changed

app/main.py
```diff
@@ -1,2 +1,64 @@
...
+    def _hash_index(self, key: Any) -> int:
+        return hash(key) % self.capacity
+
+    def _resize(self) -> None:
+        new_capacity = self.capacity * 2
+        new_table = [None] * new_capacity
+
+        for node in self.table:
+            while node is not None:
+                current_index = hash(node.key) % new_capacity
```
Nattalli: why don't you use existed method that calculates hash index?

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #804, opened 2024-10-24
solution

pohgen:


spa7id (approved):
LGTM

app/main.py
```diff
@@ -1,2 +1,101 @@
+from typing import Any, Generator
+
+
 class Dictionary:
-    pass
+
+    def __init__(self) -> None:
+        self.max_capacity = 8
+        self.threshold = self.max_capacity - (self.max_capacity * 2 // 3)
```
Arsen-hrynevych, line 8: The calculation for the threshold might not be ideal. You're using self.max_capacity - (self.max_capacity * 2 // 3), which is equivalent to self.max_capacity // 3, meaning the table resizes when it's about 1/3rd full instead of 2/3rds full as typically expected with hash tables.

`
self.threshold = int(self.max_capacity * 2 / 3)`

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        else:
+            raise KeyError("Key not found")
+
+    def __len__(self) -> int:
+        return len(self.hash_table) - self.hash_table.count(None)
+
+    def __delitem__(self, key: Any) -> None:
+        index = self.find_value_from_key(key)
+        if self.hash_table[index]:
+            del self.hash_table[index]
```
Arsen-hrynevych, line 30: Instead of deleting the index, set it to None to maintain the structure of the hash table

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def __init__(self) -> None:
+        self.max_capacity = 8
+        self.threshold = self.max_capacity - (self.max_capacity * 2 // 3)
+        self.hash_table = [None] * self.max_capacity
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.hash_table.count(None) <= self.threshold:
+            self.resize_table()
+        index = self.find_index(key)
```
Arsen-hrynevych, line 14: __setitem__ and find_index perform key lookups to find an empty slot. This means the same key might be hashed and compared multiple times.

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        return len(self.hash_table) - self.hash_table.count(None)
+
+    def __delitem__(self, key: Any) -> None:
+        index = self.find_value_from_key(key)
+        if self.hash_table[index]:
+            del self.hash_table[index]
+        else:
+            raise KeyError("Key not found")
+
+    def find_index(self, key: Any) -> int:
```
Arsen-hrynevych, line 34: Both methods have almost identical probing logic. This can be simplified and unified into a single helper method.

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+            while self.hash_table[index]:
+                if index == self.max_capacity - 1:
+                    index = 0
+                    continue
+                if self.hash_table[index][0] == key:
+                    return index
+                index += 1
+        return index
+
+    def find_value_from_key(self, key: Any) -> int:
```
Arsen-hrynevych, line 49: Both methods have almost identical probing logic. This can be simplified and unified into a single helper method.

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        hashed_ = hash(key)
+        index = hashed_ % self.max_capacity
+        if not self.hash_table[index]:
+            raise KeyError("Key doesn't exist")
```
Arsen-hrynevych, line 53: raise KeyError(f"Key {key} not found.")

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+
+    def __len__(self) -> int:
+        return len(self.hash_table) - self.hash_table.count(None)
+
+    def __delitem__(self, key: Any) -> None:
+        index = self.find_value_from_key(key)
+        if self.hash_table[index]:
+            del self.hash_table[index]
+        else:
+            raise KeyError("Key not found")
```
Arsen-hrynevych, line 32: raise KeyError(f"Key {key} not found.")

app/main.py
```diff
@@ -1,2 +1,101 @@
...
+        return self.hash_table[index][1]
+
+    def pop(self, key: Any) -> Any:
+        index = self.find_value_from_key(key)
+        pair = self.hash_table[index]
+        self.hash_table[index] = None
+        return pair[1]
+
+    def update(self, other: "Dictionary") -> None:
+        for index, key in enumerate(other.hash_table):
```
Arsen-hrynevych, line 91: this can be simplified

Arsen-hrynevych (changes requested):
Mostly correct. Keep going

PR #822, opened 2024-10-31
Solution

sofiiasavkova:


app/main.py
```diff
@@ -1,2 +1,46 @@
...
+        while node:
+            if node.key == key:
+                node.value = value
+                return
+            node = node.next
+
+        new_node = self.Node(key, value)
+        new_node.next = self.storage[index]
+        self.storage[index] = new_node
+        self.size += 1
```
AnyoneClown: So, your storage will only have 8 capacity? What about resizing of this dictionary?

>Load Factor: When should your dictionary resize to maintain efficiency?
>Resize Strategy: How should your dictionary expand its capacity when needed?

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+                new_index = hash(node.key) % new_capacity
+                new_node = self.Node(node.key, node.value)
+                new_node.next = new_storage[new_index]
+                new_storage[new_index] = new_node
+                node = node.next
+
+        self.capacity = new_capacity
+        self.storage = new_storage
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
Arsen-hrynevych, line 32: When setting an existing key to a new value, you correctly update node.value. However, you also continue searching the rest of the chain, which is unnecessary and can lead to performance inefficiencies.

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+                return node.value
+            node = node.next
+
+        raise KeyError(f"Key {key} not found.")
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __repr__(self) -> str:
+        return f"Dictionary with {self.size} items."
```
Arsen-hrynevych, line 65: mplement a __delitem__ method

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        self.size = 0
+        self.storage = [None] * self.capacity
+
+    class Node:
+        def __init__(self, key: Any, value: Any) -> None:
+            self.key = key
+            self.value = value
+            self.next = None
+
+    def _resize(self) -> None:
```
Arsen-hrynevych, line 16: When _resize() is called, it correctly rehashes and repositions nodes in the new new_storage array. However, if there are any collisions at the new index, the code will overwrite the existing chain by creating a new chain rather than properly appending it.

When adding nodes to new_storage, continue chaining nodes at the correct index in new_storage by appending to the end of the existing chain (rather than always assigning new_node.next = new_storage[new_index]).

Arsen-hrynevych (approved):
Ok

PR #848, opened 2024-12-23
Solution

maxakim0v:


dottfmar (approved):

taraskhom (approved):
It seems that everything is ok

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8, load_factor: float = 0.7) -> None:
+        self.capacity = capacity
+        self.load_factor = load_factor
+        self.size = 0
+        self.table = [None] * capacity
+
+    def _hash(self, key: Any) -> int:
```
L1nk3rrr, line 11: key can't be Any, it must be Hashable type ( you can import from typing lib)
In all places where you waiting for Any you can change it

app/main.py
```diff
@@ -1,2 +1,91 @@
...
+        return hash(key) % self.capacity
+
+    def _resize(self) -> None:
+        new_capacity = self.capacity * 2
+        new_table = [None] * new_capacity
+
+        for item in self.table:
+            if item:
+                key, value = item
+                new_index = hash(key) % new_capacity
```
L1nk3rrr, line 21: Use your method `_hash`, not sense from code duplication (you can modificate your method, so to have not required attribute, if it is None -> take from self, for example
```

    def _hash(self, key: Hashable, capacity: int = None) -> int:
        if capacity is None:
            capacity = self.capacity
        return hash(key) % capacity
```

L1nk3rrr (changes requested):
Well done! Left some minor comments for improvement of your code

PR #849, opened 2024-12-23
Solution

julia4406:


app/main.py
```diff
@@ -1,2 +1,75 @@
...
+        index = hash(key) % self.capacity
+        if self.hash_table[index] and self.hash_table[index].key == key:
+            return self.hash_table[index].value
+        for index, el in enumerate(self.hash_table):
+            if el and el.key == key:
+                return self.hash_table[index].value
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        self.length = len([el for el in self.hash_table if el])
```
valerii-kashpur: You also have self.length atr, so that can be used instead of

valerii-kashpur (approved):
interesting solution

PR #854, opened 2024-12-24
Solution_Dictionary

Bleit1:


VovaGld (approved):

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+                self.__setitem__(node.key, node.value)
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        key_hash = hash(key)
+        index = self.calculate_index(key, key_hash)
+
+        if self.hash_table[index] is None:
+            if self.size + 1 > self.capacity * LOAD_FACTOR:
+                self.resize()
+                return self.__setitem__(key, value)
```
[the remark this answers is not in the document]
L1nk3rrr: at __setitem__ you don't need to return smth. 
If you resize you simply go out (return)

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+    value: Any
+    _hash: int
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = INITIAL_CAPACITY) -> None:
+        self.capacity = capacity
+        self.size = 0
+        self.hash_table: List[Optional[Node]] = [None] * self.capacity
```
L1nk3rrr, line 20: Maybe it would be better that you hash_table contains by default Nodes, not a None with Nodes?

app/main.py
```diff
@@ -1,2 +1,114 @@
...
+                self.hash_table[index] is not None
+                and self.hash_table[index].key != key
+        ):
+            index = (index + 1) % self.capacity
+        return index
+
+    def resize(self) -> None:
+        old_hash_table = self.hash_table
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
```
L1nk3rrr, line 34: same about None

app/main.py
L1nk3rrr, line 20: maybe better to use private variable access type?

L1nk3rrr (changes requested):
Well done!
Minor points for improvement

PR #868, opened 2024-12-24
Solution_dictionary

SummerKatrin:


app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        if self.hashtable[index] is None:
+            if self.size + 1 > self.capacity * RESIZE_THRESHOLD:
+                old_hash_table = self.hashtable
+                self.__init__(self.capacity * 2)
+
+                for node in old_hash_table:
+                    if node is not None:
+                        self.__setitem__(node.key, node.value)
+
+                return self.__setitem__(key, value)
```
L1nk3rrr: do you really need to return smth from __setitem__?

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+        self.key = key
+        self.hash = hash(key)
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = INITIAL_CAPACITY) -> None:
+        self.capacity = capacity
+        self.size = 0
+        self.hashtable: list[None | Node] = [None] * self.capacity
```
L1nk3rrr: Maybe it would be better for your variables to have not public access type.

app/main.py
```diff
@@ -1,2 +1,56 @@
...
+            index = (index + 1) % self.capacity
+        return index
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        index = self.calculate_index(key)
+
+        if self.hashtable[index] is None:
+            if self.size + 1 > self.capacity * RESIZE_THRESHOLD:
+                old_hash_table = self.hashtable
+                self.__init__(self.capacity * 2)
```
L1nk3rrr: Thats very bad! We don't use __init__ after initializing of our object!
Write a resize method or smth like that
`Resize Strategy: How should your dictionary expand its capacity when needed?`

L1nk3rrr (changes requested):
Nice!
Need to improve some minor points

L1nk3rrr (approved):
Well done!

PR #869, opened 2024-12-25
Solution

bodiakof:


yelizaveta-zh (approved):
Good work =)

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.hash_table: list[Optional[Node]] = [None] * self.capacity
+        self.length: int = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        index = hash(key) % self.capacity
```
L1nk3rrr: code duplication, make a method to know a hash value

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.hash_table: list[Optional[Node]] = [None] * self.capacity
+        self.length: int = 0
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
L1nk3rrr: also key is not Any type, it's a Hashable type from typing lib, not Any can be hashable

app/main.py
```diff
@@ -1,2 +1,68 @@
...
+        node = self.hash_table[index]
+        while node:
+            if node.key == key:
+                return True
+            node = node._next
+        return False
+
+    def __getitem__(self, key: Any) -> Any:
+        if key not in self:
+            raise KeyError("Key not found in dictionary")
```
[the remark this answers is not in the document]
L1nk3rrr: AI is right, rework this to not use `if key not in self`

L1nk3rrr (changes requested):
Great work, but also here are some points for improvement.
Key in all places must be hashable, not Any type.
Also a question - why your node know smth about next node? Is it efficient?

bodiakof:
> why your node know smth about next node? Is it efficient?

as i understand, this is how linked lists handle collisions and if collisions aren't frequent, it should be pretty efficient, but otherwise, i guess other data structures (like arrays) might be more efficient

L1nk3rrr (approved):
Well done!

PR #872, opened 2024-12-29
Solution

VolodymyrPovroznyk:


app/main.py
```diff
@@ -1,2 +1,144 @@
+import math
+from collections.abc import Hashable
+from typing import Any, Dict, Generator, Iterable, Tuple
```
vsmutok: If `Hashable` is used only for annotation, the import should be done from `from typing import Hashable `.

app/main.py
```diff
@@ -1,2 +1,144 @@
...
+
+        while True:
+            target = self.hash_table[index]
+            if target is None:
+                return default
+            if target != "del" and target[1] == key_hash and target[0] == key:
+                return target[2]
+            index = (index + 1) % self.capacity
+            if start_index == index:
+                return default
```
vsmutok: Try using `__getitem__ ` here to simplify the code.

app/main.py
```diff
@@ -1,2 +1,144 @@
...
+            if target != "del" and target[1] == key_hash and target[0] == key:
+                value = target[2]
+                self.hash_table[index] = "del"
+                self.length -= 1
+                return value
+            index = (index + 1) % self.capacity
+            if start_index == index:
+                if default == "raise":
+                    raise KeyError(f"Key {key} does not exist")
+                return default
```
vsmutok: Try using `__getitem__ ` here to simplify the code.

app/main.py
```diff
@@ -1,2 +1,144 @@
...
+
+        while True:
+            target = self.hash_table[index]
+            if target is None:
+                if default == "raise":
+                    raise KeyError(f"Key {key} does not exist")
+                return default
+            if target != "del" and target[1] == key_hash and target[0] == key:
+                value = target[2]
+                self.hash_table[index] = "del"
```
vsmutok: The `pop` method should immediately remove the element. Your current solution is incorrect.

vsmutok (changes requested):

app/main.py
VolodymyrPovroznyk: Ok

app/main.py
VolodymyrPovroznyk: How to implement returning a default value instead of raising KeyError?`__getitem__` method raises KeyError when the key doesn't exist, and the `get `method should return a default value instead of an error.

VolodymyrPovroznyk: Maybe by using try-except blocks?

app/main.py
VolodymyrPovroznyk: Used `__getitem__` and` __delitem__`

app/main.py
VolodymyrPovroznyk: Used` __delitem__` method

vasylhnatiuk (approved):

PR #893, opened 2025-02-13
Solution

Baranotik15:


app/main.py
```diff
@@ -1,2 +1,83 @@
...
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self.hash_formula(key)
+
+        for k, v in self.container[index]:
+            if k == key:
+                return v
+        raise KeyError(f"Key {key} not found")
+
+    def __len__(self) -> int:
+        return self.quantity_key
```
Boyok1337, line 46: Group all magic methods and move to the beginning. Magic methods should always be at the beginning, and then the rest of the methods

Boyok1337 (approved):

PR #896, opened 2025-02-18
Solution. Dict v1.0

kram3ko:


app/main.py
```diff
@@ -1,2 +1,136 @@
...
+        self.capacity *= 2
+        self.hash_table = [None] * self.capacity
+        self.threshold = self.capacity * self.load_factor
+        self.current_load = 0
+        for node in old_hash_table:
+            index = node.hash_value % self.capacity
+            while self.hash_table[index] is not None:
+                index = (index + 1) % self.capacity
+            self.hash_table[index] = node
+            self.current_load += 1
```
TopalRuslan: you just transfer all the elements from the old to the new hash table, there is no point in calculating the numbers, it will be the same

Baranotik15 (approved):
Good job, but you can remove some actions

PR #913, opened 2025-03-10
Solution

Nikita-pyth:


BronkstoneBro (approved):
Well done!)

app/custom_dict.py
```diff
@@ -0,0 +1,96 @@
...
+        self.value = value
+
+
+class CustomDict:
+    _no_value = object()
+    _deleted_marker = object()
+
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.dictionary: list[Node | object | None] = [None] * self.capacity
```
Boyok1337, line 17: Here are better to use `from typing import Optional`

app/custom_dict.py
```diff
@@ -0,0 +1,96 @@
...
+            raise KeyError(f"Key {key} is not present in the dictionary."
+                           f" Unable to delete.")
+        self.dictionary[index] = self._deleted_marker
+        self._used_cells_counter -= 1
+
+    def pop(self, key: Hashable, default: Any = _no_value) -> Any:
+        index = self._probe(key)
+        if index is None:
+            if default is self._no_value:
+                raise KeyError(f"Key {key} is not present in the dictionary."
```
Boyok1337, line 66: You can check directly `if default is None:`

app/node.py
```diff
@@ -0,0 +1,12 @@
+from typing import Any, Hashable
+
+
+class Node:
+    def __init__(self, key: Hashable, value: Any) -> None:
+        try:
+            self.hash = hash(key)
+        except TypeError:
+            raise TypeError(f"Unhashable key: {key} ({type(key).__name__})")
```
Boyok1337, line 9: For it we have good construction like this `if not hasattr(key, '__hash__'):`

app/node.py
```diff
@@ -0,0 +1,12 @@
+from typing import Any, Hashable
+
+
+class Node:
+    def __init__(self, key: Hashable, value: Any) -> None:
```
Boyok1337, line 5: Why you need 2 `class Node`?

app/custom_dict.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.dictionary: list[Node | object | None] = [None] * self.capacity
+        self.load_factor: float = 2 / 3
+        self._used_cells_counter: int = 0
+
+    @property
+    def threshold(self) -> int:
+        return ceil(self.capacity * self.load_factor)
```
Boyok1337, line 23: Magic methods - firtly, then regular one, follow best practices here

Boyok1337 (changes requested):

PR #930, opened 2025-03-12
Develop

yarms1:


Nikita-pyth (changes requested):
Magic methods should be before regular ones, and i believe they asked us to implement it with Node class, not a list of tuples? I mean, your implementation still works, but i'm not sure this is the way they wanted us to do this exercise

PR #943, opened 2025-04-05
Solution

olek007:


olek007:
@mateacademy-ai-mentor 
Potential collisions are handle well. There is key comparison and key hash comparison.

olek007:
@mateacademy-ai-mentor 
Error messages was added.

PR #982, opened 2025-07-05
Develop

zqpiq:


app/main.py
```diff
@@ -1,2 +1,29 @@
...
+    def __init__(self, key: Any, value: Any) -> None:
+        self.key = key
+        self.hash = hash(key)
+        self.value = value
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.data = []
```
AnyoneClown: >Key Concepts to Consider

As you implement your custom dictionary, you'll need to keep the following concepts in mind:

    Initial Capacity: How much space should your dictionary allocate at the start?
    Load Factor: When should your dictionary resize to maintain efficiency?
    Resize Strategy: How should your dictionary expand its capacity when needed?
>

You didn't implement this logic, right? Add it to your code. Read about `capacity` `lenght` and `resizing` of a dictionary

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        self.value = value
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self._initial_capacity = 8
+        self._data = [None] * self._initial_capacity
+        self._load_factor = 2 / 3
+        self._threshold = int(self._initial_capacity * self._load_factor)
```
L1nk3rrr, line 16: same as self._threshold = self. update_threshold()

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        position = hash(key) % self._initial_capacity
+        while True:
+            if self._data[position] is None:
+                self._data[position] = Node(key, value)
+                break
+            elif self._data[position].key == key:
+                self._data[position].value = value
+                break
+            else:
+                position = (position + 1) % self._initial_capacity
```
L1nk3rrr, line 55: redundant else statement, it can be simple:
```
            if self._data[position] is None:
                self._data[position] = Node(key, value)
                break
            elif self._data[position].key == key:
                self._data[position].value = value
                break
            position = (position + 1) % self._initial_capacity
```

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+
+    def __getitem__(self, key: Any) -> Any:
+        for node in self._data:
+            if node is not None:
+                if node.key == key:
```
L1nk3rrr, line 60: can be used one if
```
if node is not None and node.key == key:
```

app/main.py
```diff
@@ -1,2 +1,65 @@
+from typing import Any
+
+
+class Node:
+    def __init__(self, key: Any, value: Any) -> None:
```
L1nk3rrr, line 5: key can be only hashable values

L1nk3rrr (changes requested):
review comments to improve your code!

PR #985, opened 2025-07-06
Solution

Arch0998:


zqpiq (approved):

app/main.py
```diff
@@ -1,2 +1,35 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.lenght = 0
+        self.hash_table = [[] for _ in range(8)]
```
AnyoneClown: > Storage Structure
> 
> You'll store each key-value pair as a node in a hash table. Each node should contain the following information:
> 
>     key
>     hash
>     value
> 

Refactor using nodes

AnyoneClown (changes requested):
You didn't implement this features:
>
    Initial Capacity: How much space should your dictionary allocate at the start?
    Load Factor: When should your dictionary resize to maintain efficiency?
    Resize Strategy: How should your dictionary expand its capacity when needed?
>

AnyoneClown (changes requested):
Check comments and refactor your code

AnyoneClown (approved):
LGTM

PR #990, opened 2025-07-15
Solution

farzamgt:


viannik (approved):
The implementation is clean, efficient, and matches the core requirements

PR #992, opened 2025-07-16
my terrible solution...

Orixara:


viannik (approved):
Implements all core features correctly
Meets time complexity requirements (exceeds in resizing)
Clean, readable, and well-structured

Optimize _resize to reuse stored hashes.
The current implementation (hash(self.x + self.y)) may cause collisions for points like (1, 2) and (2, 1). Update Point’s __hash__ to hash((self.x, self.y)) for better distribution.

PR #997, opened 2025-07-17
Solution

yagginov:


Artcrafterrra (approved):
Awesome job! You fully implemented all additional logic. It works the same way as a dictionary under the hood.

yagginov:
@mateacademy-ai-mentor There is actually a mistake in your comment, because I move the index forward until I find the desired element, or until I find an empty cell.

PR #1007, opened 2025-07-17
Initial solution with basic methods

Kostenikov:


Vskesha (approved):
Short and simple but clear and functional.

PR #1011, opened 2025-07-18
Solution

VadymKyrylko:


app/main.py
```diff
@@ -1,2 +1,57 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8,
+                 load_factor_threshold: float = 0.7) -> None:
```
L1nk3rrr: one attribute per line

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        self.capacity = capacity
+        self.load_factor_threshold = load_factor_threshold
+        self.table = [[] for _ in range(self.capacity)]
+        self.size = 0
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __setitem__(self, key: Any, value: Any) -> None:
```
L1nk3rrr: key can be only hashable type

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        for i, element in enumerate(self.table[index]):
+            if element[0] == key:
+                self.table[index][i] = (key, value, hash_key)
+                break
+        else:
+            self.table[index].append((key, value, hash_key))
+            self.size += 1
+
+    def __getitem__(self, key: Any) -> Any:
+        index = hash(key) % len(self.table)
```
L1nk3rrr: repeated part of code, maybe better to make some method

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        for element in self.table[index]:
+            if element[0] == key:
```
L1nk3rrr: element[0], element[1], element[2] ... better to make some dataclass or smth like that for every cell/node that you will use. It will improve readability of your code

app/main.py
```diff
@@ -1,2 +1,57 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    def __init__(self, capacity: int = 8,
+                 load_factor_threshold: float = 0.7) -> None:
+        self.capacity = capacity
+        self.load_factor_threshold = load_factor_threshold
+        self.table = [[] for _ in range(self.capacity)]
```
L1nk3rrr: looks like it must be protected or private attr

app/main.py
```diff
@@ -1,2 +1,57 @@
...
+        self.size = 0
```
L1nk3rrr: for size you can realize __len__ magic method

L1nk3rrr (changes requested):
Left comments that will improve your code!

L1nk3rrr (approved):
Now better, well done!

PR #1012, opened 2025-07-18
Solution

Irahromska:


app/main.py
```diff
@@ -1,2 +1,74 @@
+from typing import Any, List
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity: int = 8
+        self.size: int = 0
+        self.load_factor: float = 0.75
```
MorsImmortalis05, line 8: it's better to define default capacity and load factor as class attributes, as they won't be changed later

app/main.py
```diff
@@ -1,2 +1,74 @@
...
+        self.capacity *= 2
+        self.buckets = [[] for _ in range(self.capacity)]
+        self.size = 0
+
+        for bucket in old_buckets:
+            for node in bucket:
+                self[node.key] = node.value
+
+
+class Node:
```
MorsImmortalis05, line 70: it's better to define this classs in the begginning since you're using this class before it's defined, it will be more readable. and also you can use @dataclass here

MorsImmortalis05 (approved):

MNDonut (approved):

PR #1014, opened 2025-07-18
Solution.

Otst0y:


AnyoneClown (changes requested):
Read your task properly, refactor using `Nodes`

PR #1015, opened 2025-07-18
Develop

trellle:


app/main.py
```diff
@@ -1,2 +1,108 @@
...
+
+    def __iter__(self) -> iter:
+        return DictIterator(self)
+
+
+class Node:
+    def __init__(self) -> None:
+        self.key = None
+        self.hash = None
+        self.value = None
```
dimashevchukk: class Node should be defined before Dictionary

app/main.py
```diff
@@ -1,2 +1,108 @@
+from __future__ import annotations
+from typing import Any
+
+
+max_cap = 8
+cap = round(max_cap * 2 / 3)
```
dimashevchukk: define constants in class Dictionary

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+
+
 class Dictionary:
-    pass
+    def __init(self) -> None:
```
dimashevchukk: def __init__

app/main.py
```diff
@@ -1,2 +1,108 @@
...
+        self.__hash_table = [Node() for _ in range(0, max_cap)]
```
dimashevchukk: use None instead of Node while initing

dimashevchukk (changes requested):

dimashevchukk:
rewrite code so it works and rerequest me later

dimashevchukk (approved):

PR #1017, opened 2025-07-19
solution

MorsImmortalis05:


app/main.py
```diff
@@ -1,2 +1,87 @@
...
+        result = 0
+        for i in self.__table:
+            if isinstance(i, Node):
+                result += 1
+        return result
+
+    def _find_bucket_index(self, key: Hashable) -> int:
+        key_hash = hash(key)
+        bucket_index = key_hash % len(self.__table)
+        return bucket_index
```
L1nk3rrr: it can be one liner

`return hash(key) % len(self.__table)`

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+
+    def _find_empty_bucket(self, key: Hashable) -> Union[int, None]:
+        bucket_index = self._find_bucket_index(key)
+        for i in range(len(self.__table)):
```
L1nk3rrr: if you don't use `i` you can use `for _ in smht`

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._find_existing_bucket(key=key)
+        if index is None:
+            raise KeyError(key)
+        return self.__table[index].value
+
+    def __len__(self) -> int:
+        result = 0
+        for i in self.__table:
+            if isinstance(i, Node):
```
L1nk3rrr: use more readable names, use node or smth like this

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+                result += 1
+        return result
```
L1nk3rrr: it can be one liner, you need to count a lines in a table that is  nodes, try to figure out how to do it

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+    def __init__(self, buckets: int = INITIAL_CAPACITY) -> None:
+        self.__buckets = buckets
+        self.__table = [None] * buckets
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        existing_bucket = self._find_existing_bucket(key)
+        bucket = self._find_empty_bucket(key)
+        if existing_bucket is not None:
+            self.__table[existing_bucket].value = value
+        elif self.__len__() >= self.__buckets * Dictionary.LOAD_FACTOR:
```
L1nk3rrr: len(self)

app/main.py
```diff
@@ -1,2 +1,87 @@
...
+    INITIAL_CAPACITY = 8
+    LOAD_FACTOR = 2 / 3
+
+    def __init__(self, buckets: int = INITIAL_CAPACITY) -> None:
+        self.__buckets = buckets
+        self.__table = [None] * buckets
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        existing_bucket = self._find_existing_bucket(key)
+        bucket = self._find_empty_bucket(key)
```
L1nk3rrr, line 25: no sense to use it as you call the same lower

L1nk3rrr (changes requested):
Review comments to improve your code!

app/main.py
MorsImmortalis05: but i use this existing bucket in the else clause, and call find_empty_bucket again only if the capacity increased

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+            )
+
+    def __getitem__(self, key: Hashable) -> Any:
+        index = self._find_existing_bucket(key=key)
+        if index is None:
+            raise KeyError(key)
+        return self.__table[index].value
+
+    def __len__(self) -> int:
+        return sum(1 for item in self.__table if isinstance(item, Node))
```
AnyoneClown, line 51: len() implementation is correct but O(n); consider tracking size with a counter for O(1) performance.

app/main.py
```diff
@@ -1,2 +1,81 @@
...
+
+    def _find_existing_bucket(self, key: Hashable) -> Union[int, None]:
+        bucket_index = self._find_bucket_index(key)
+        while self.__table[bucket_index] is not None:
+            if self.__table[bucket_index].key == key:
+                return bucket_index
+            bucket_index += 1
+            if bucket_index >= len(self.__table):
+                bucket_index = 0
+        return None
```
AnyoneClown: Use item_hash in _find_existing_bucket to speed up comparisons.

AnyoneClown (changes requested):

PR #1020, opened 2025-07-21
dict castom

Salmon-eat:


app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        self.capacity: int = initial_capacity
+        self.load_factor: float = load_factor
+        self.size: int = 0
+        self.table: List[Optional[Node]] = [None] * self.capacity
+
+    def __len__(self) -> int:
+        return self.size
+
+    def _find_slot(self, key_hash: int, key: Any) -> int:
+        index: int = key_hash % self.capacity
```
L1nk3rrr, line 25: `key_hash % self.capacity` - repeated, can be in external method

app/main.py
```diff
@@ -1,2 +1,67 @@
+from typing import Any, Optional, List
+
+
+class Node:
+    def __init__(self, key: Any, value: Any) -> None:
```
L1nk3rrr, line 5: key can be only hashable

L1nk3rrr: not only here, look for all places

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        self.hash: int = hash(key)
+        self.value: Any = value
+
+
 class Dictionary:
-    pass
+    def __init__(self,
+                 initial_capacity: int = 8,
+                 load_factor: float = 2 / 3
+                 ) -> None:
```
L1nk3rrr, line 15: can be better formated

app/main.py
```diff
@@ -1,2 +1,67 @@
...
+        self.capacity: int = initial_capacity
+        self.load_factor: float = load_factor
+        self.size: int = 0
+        self.table: List[Optional[Node]] = [None] * self.capacity
+
+    def __len__(self) -> int:
+        return self.size
```
L1nk3rrr, line 22: better to make size protected or private and use len

L1nk3rrr (changes requested):
You can improve your code!

PR #1039, opened 2025-08-28
Solution

Konstantin-Svt:


Konstantin-Svt:
@mateacademy-ai-mentor 
1. resize trigger in "len(self.__hash_table) * 2 // 3 == self.__length" should be <= instead of >=, because self.__length is amount of occupied cells, not total length of hash table. Testing shows that with  >= it resizes table prematurely even with 1 occupied cell.
2. Testing shows that "__delitem__" does raise KeyError when key is not found, because KeyError is triggered inside value_finder_by_key.
3. Testing shows that "__delitem__" does not disrupt searching chains, because  value_finder_by_key keeps searching even if it encounters None untill it completes full circle in hash table. Only then it raises KeyError. But I implemented __TOMBSTONE feature anyway to reduce amount of iterations in value_finder_by_key function.

PR #1059, opened 2025-11-03
'Solution'

Vorongor:


app/main.py
```diff
@@ -1,2 +1,172 @@
...
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        """
+        Initialize the dictionary with a default
+        capacity and prepare internal storage.
+        """
+        self.size = 8
```
DonDmytron: bad naming, capacity will be better

app/main.py
```diff
@@ -1,2 +1,172 @@
...
+        self.current_size = 0
```
DonDmytron: just size

app/main.py
```diff
@@ -1,2 +1,172 @@
...
+        self.resize_point = round(self.size * (2 / 3))
```
DonDmytron: 2 / 3 would be better have as an attribute

app/main.py
```diff
@@ -1,2 +1,172 @@
...
+        Insert or update a key-value pair in the dictionary.
+        Automatically resizes if load factor exceeds threshold.
+        """
+        idx = self.__hash_index(key)
+
+        while (
+            self.bucket[idx] is not None
+            and self.bucket[idx] is not _TOMBSTONE
+            and self.bucket[idx][0] != key
+        ):
```
DonDmytron: please, use this logic in separate method and create separate class Node with key, hash, value for better reading

DonDmytron (changes requested):

app/main.py
```diff
@@ -1,2 +1,209 @@
...
+        default: The value to return if the key is not found. Defaults to None.
+    Returns:
+        The value associated with the key, or the provided default.
+    """
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def pop(self, key: Any) -> Any:
```
DonDmytron, line 170: from typing import Hashable would be better

DonDmytron (approved):
good job )

PR #1076, opened 2025-11-17
completed task

ostapshpeha:


app/main.py
```diff
@@ -1,2 +1,71 @@
...
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.size = 0
+        self.capacity = 8
+        self.hash_table: list = [None] * self.capacity
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        hsh = hash(key)
+        idx = hsh % len(self.hash_table)
```
leemonch1k-space, line 12: You can change this:
idx = hsh % len(self.hash_table)
To:
idx = hsh % self.capacity
if you want to simplify it

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+                next_node = node.next_iter
+                new_idx = node.hsh % new_capacity
+                node.next_iter = new_table[new_idx]
+                new_table[new_idx] = node
+                node = next_node
+        self.capacity = new_capacity
+        self.hash_table = new_table
+
+
+class Node:
```
leemonch1k-space, line 60: Why this class after the Dictionary class?

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+                node = node.next_iter
+            new_node = Node(key, hsh, value)
+            new_node.next_iter = self.hash_table[idx]
+            self.hash_table[idx] = new_node
+            self.size += 1
+
+        if self.size > self.capacity * 0.75:
+            self.__resize__()
+
+    def __getitem__(self, key: Hashable) -> None:
```
leemonch1k-space: Wrong annotation type. You are trying to return a value by key, but your annotation says: I will not return anything.
will be more suitable.

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+                return head.value
+            head = head.next_iter
+        raise KeyError(f"Key {key} not found")
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __resize__(self) -> None:
+        old_table = self.hash_table
+        new_capacity = self.capacity * 2
```
leemonch1k-space, line 46: Here you can change self.capacity immediately, there is no need to do it later

app/main.py
```diff
@@ -1,2 +1,71 @@
...
+        self.capacity = new_capacity
+        self.hash_table = new_table
+
+
+class Node:
+    def __init__(self,
+                 key: Hashable,
+                 hsh: int,
+                 value: Any,
+                 next_iter: Optional = None
```
leemonch1k-space, line 65: It's okey, but will be better if you use:
next_iter: Optional["Node"] = None

leemonch1k-space (approved):
good job!

PR #1089, opened 2025-11-18
Solution

Guillelmina8:


app/main.py
```diff
@@ -1,2 +1,65 @@
+from typing import Any, Hashable
+
+
 class Dictionary:
-    pass
+    def __init__(self) -> None:
+        self.capacity = 8
+        self.length = 0
+        self.hash_table = [None] * self.capacity
+        self.load_factor = 0.75
```
DonDmytron: incorrect value

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        try:
+            key_hash = hash(key)
+        except TypeError:
+            raise TypeError("unhashable type")
+        if self.length >= self.capacity * 0.75:
+            self._resize()
+        index = key_hash % self.capacity
+        if self.hash_table[index] is None:
+            self.hash_table[index] = [[key_hash, key, value]]
```
DonDmytron: Add a Node class to store the hash, key, value and use instance of this class in a your hash table.

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+            self.length += 1
+            return
+        for obj in self.hash_table[index]:
```
DonDmytron: incorrect collision handling

app/main.py
```diff
@@ -1,2 +1,65 @@
...
+        self.hash_table = new_hash_table
+
+    def __getitem__(self, key: Hashable) -> Any:
+        key_hash = hash(key)
+        index = key_hash % self.capacity
+
+        obj = self.hash_table[index]
+        if obj is None:
+            raise KeyError
+        for item in obj:
```
DonDmytron: maybe 'elif' would be better instead of 'for'

DonDmytron (changes requested):

oliasymak24 (approved):

PR #1156, opened 2026-04-08
create dictionary simulate class

pawunder1992:


app/main.py
```diff
@@ -1,2 +1,80 @@
...
+    lenght = 0
+
+    def __init__(self) -> None:
+        self.hash_table: list = [None] * 8
+
+    def resize(self) -> None:
+        self.CAPACITY *= 2
+        self.resize_table = self.hash_table
+        self.hash_table = [None] * self.CAPACITY
+        self.lenght = 0
```
VadimPusiak788: Typo in variable name, should be length

app/main.py
```diff
@@ -1,2 +1,80 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    CAPACITY = 8
```
VadimPusiak788: Shared across all instances  mutating them affects every instance. Each dictionary object should have its own capacity and length.

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+        if self.lenght >= self.CAPACITY * 2 // 3:
+            self.resize()
+
+        element = Node(key, value)
+        index = element._hash % self.CAPACITY
+        while True:
+            if index == len(self.hash_table):
+                index = 0
+                continue
+            if self.hash_table[index] is None:
```
VadimPusiak788, line 31: If your table is full, and the key is absent, this may become an infinite loop because there may be no None. 
You can use 
`for _ in range(self.capacity):`

app/main.py
```diff
@@ -1,2 +1,80 @@
...
-    pass
+    CAPACITY = 8
+    lenght = 0
+
+    def __init__(self) -> None:
+        self.hash_table: list = [None] * 8
+
+    def resize(self) -> None:
+        self.CAPACITY *= 2
+        self.resize_table = self.hash_table
```
VadimPusiak788: resize_table becomes an unnecessary instance attribute. old table should be a local variable
`old_table = self.hash_table`

app/main.py
```diff
@@ -1,2 +1,80 @@
+from typing import Any
+
+
 class Dictionary:
-    pass
+    CAPACITY = 8
+    lenght = 0
+
+    def __init__(self) -> None:
+        self.hash_table: list = [None] * 8
```
VadimPusiak788: Hardcoded 8, you should use CAPACITY variable

app/main.py
```diff
@@ -1,2 +1,80 @@
...
+            else:
+                index += 1
+
+    def __getitem__(self, key: Any) -> Any:
+        index = hash(key) % self.CAPACITY
+        while True:
+            if index == len(self.hash_table):
+                index = 0
+                continue
+            if self.hash_table[index] is None:
```
VadimPusiak788, line 49: If your table is full, and the key is absent, this may become an infinite loop because there may be no None.
You can use
for _ in range(self.capacity):

VadimPusiak788 (changes requested):

VadimPusiak788 (approved):

PR #1157, opened 2026-04-09
Develop

Neyrony:
In order to prevent your dissatisfaction on me, the False in list it's a stub to prevent disapiring of some variables, that mean if you find something False mean that your element could be in the next position


Neyrony:
@mateacademy-ai-mentor i removed unnecesary func

PR #1159, opened 2026-04-14
'Solution'

yakinder46:


app/main.py
```diff
@@ -1,2 +1,75 @@
...
+
 class Dictionary:
-    pass
+    """
+    Custom Dictionary implementation using hash table with open addressing.
+    Each node in the table is a tuple: (key, stored_hash, value).
+    """
+
+    def __init__(self, capacity: int = 8) -> None:
+        self.capacity: int = capacity
```
Boyok1337, line 10: Validate capacity in __init__. Right now, passing 0 or a negative value can break the implementation because self.size / self.capacity will fail.

app/main.py
```diff
@@ -1,2 +1,75 @@
...
+    def __setitem__(self, key: Any, value: Any) -> None:
+        if self.size / self.capacity >= self.load_factor:
+            self._resize()
+
+        full_hash: int = self._hash(key)
+        index: int = full_hash % self.capacity
+
+        while self.table[index] is not None:
+            node = self.table[index]
+            # Безопасная проверка: node уже не None благодаря циклу
```
Boyok1337: Replace the non-English comment with an English one, or remove it if it does not add much value.

Boyok1337 (changes requested):
Reduce duplicated probing logic. The slot-searching logic is repeated in __setitem__, __getitem__, and _insert_during_resize(), so it would be cleaner to extract it into a helper method.

app/main.py
```diff
@@ -1,2 +1,63 @@
+from typing import Any, Optional, List, Tuple
+
+
 class Dictionary:
-    pass
+
+    def __init__(self, capacity: int = 8) -> None:
+        if capacity <= 0:
+            raise ValueError("Capacity must be a positive integer")
```
vsmutok:
```suggestion
        if not isinstance(capacity, int) or capacity <= 0:
            raise ValueError("Capacity must be a positive integer")
```
Validate capacity in init. Right now, passing 0 or a negative value can break the implementation because self.size / self.capacity will fail.

vsmutok (changes requested):

Boyok1337 (approved):
LGTM!

PR #1227, opened 2026-06-28
Solution

swirtam:


swirtam:
@mateacademy-ai-mentor aside from fixing a mistake in deletion, which you noticed, I’ve added recalculation of table desired index for a first new element after resizing of the table, which behavior was previously missing.
I’ve also changed methods and attributes that are used internally by dictionary to protected attributes and methods and the string representation of the Dictionary is now closer to the original dict

